PR #4, opened 2015-02-10
Add an autostyle lua script, which autoapplies a style on a tag found by...

marco44:
Hi, I just added this autostyle lua script. I hope it may be useful to others.

This is useful when you want to automate a few operations and the tag you want to use isn't known by darktable.

### DIFF
--- a/contrib/autostyle.lua
+++ b/contrib/autostyle.lua
@@ -80,7 +80,11 @@ function autostyle_apply_one_image (image)
   end
 
   -- Apply the style to image, if it is tagged
-  local auto_dr_attr= exiftool_attribute(image.path .. '/' .. image.filename,tag);
+  local ok,auto_dr_attr= pcall(exiftool_attribute,image.path .. '/' .. image.filename,tag)
+  -- If the lookup fails, stop here
+  if (not ok) then
+    return
+  end
   if auto_dr_attr==value then
 --	  darktable.print("Image " .. image.filename .. ": autostyle automatically applied " .. darktable.preferences.read("autostyle","exif_tag","string") )
 	  darktable.styles.apply(style,image)
@@ -101,7 +105,12 @@ end
 function exiftool_attribute(path,attr)
   local cmd="exiftool -" .. attr .. " '" ..path.. "'";
   local exifresult=get_stdout(cmd)
-  local attribute= assert(string.match(exifresult,": (.*)"), "Could not find the attribute " .. attr .. " using the command: <" .. cmd .. ">")
+  local attribute=string.match(exifresult,": (.*)")
+  if (attribute == nil) then
+    darktable.print( "Could not find the attribute " .. attr .. " using the command: <" .. cmd .. ">")
+    -- Raise an error to the caller
+    error( "Could not find the attribute " .. attr .. " using the command: <" .. cmd .. ">");
+  end
   return attribute
 end
 
--- a/contrib/autostyle.lua
+++ b/contrib/autostyle.lua
@@ -49,9 +49,18 @@ function autostyle_apply_one_image (image)
   local tag,value,style_name=string.match(darktable.preferences.read("autostyle","exif_tag","string"),"(%g+)%s*=%s*(%g+)%s*=>%s*(%g+)")
 
   -- check they all exist (correct syntax)
-  assert (tag,"EXIF TAG not found in " .. darktable.preferences.read("autostyle","exif_tag","string"));
-  assert (value,"value to match not found in " .. darktable.preferences.read("autostyle","exif_tag","string"));
-  assert (style_name,"style name not found in " .. darktable.preferences.read("autostyle","exif_tag","string"));
+  if (not tag) then
+	  darktable.print("EXIF TAG not found in " .. darktable.preferences.read("autostyle","exif_tag","string"))
+	  return
+  end
+  if (not value) then
+	  darktable.print("value to match not found in " .. darktable.preferences.read("autostyle","exif_tag","string"))
+	  return
+  end
+  if (not style_name) then
+	  darktable.print("style name not found in " .. darktable.preferences.read("autostyle","exif_tag","string"))
+	  return
+  end
 
   -- First find the style (we have its name)
   local styles= darktable.styles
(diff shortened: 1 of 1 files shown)


contrib/autostyle.lua
```diff
@@ -0,0 +1,116 @@
...
+-- Apply the style to an image, if it matches the tag condition
+function autostyle_apply_one_image (image)
+  -- We need the tag, the value and the style_name provided from the configuration string
+  local tag,value,style_name=string.match(darktable.preferences.read("autostyle","exif_tag","string"),"(%g+)%s*=%s*(%g+)%s*=>%s*(%g+)")
+
+  -- check they all exist (correct syntax)
+  assert (tag,"EXIF TAG not found in " .. darktable.preferences.read("autostyle","exif_tag","string"));
+  assert (value,"value to match not found in " .. darktable.preferences.read("autostyle","exif_tag","string"));
+  assert (style_name,"style name not found in " .. darktable.preferences.read("autostyle","exif_tag","string"));
+
```
boucman: you probably want to replace these with the dt specific ones... probably a popup error message is OK in this case

contrib/autostyle.lua
```diff
@@ -0,0 +1,116 @@
...
+  -- First find the style (we have its name)
+  local styles= darktable.styles
+  local style
+  for _,s in ipairs(styles) do
+	  if s.name == style_name then
+		  style=s
+	  end
+  end
+
```
boucman: you probably want an error message here if the style is not found

contrib/autostyle.lua
marco44: You mean the asserts, right ?

contrib/autostyle.lua
```diff
@@ -0,0 +1,116 @@
...
+  local cmd="exiftool -" .. attr .. " '" ..path.. "'";
+  local exifresult=get_stdout(cmd)
+  local attribute= assert(string.match(exifresult,": (.*)"), "Could not find the attribute " .. attr .. " using the command: <" .. cmd .. ">")
+  return attribute
+end
+
+-- run command and retrieve stdout
+function get_stdout(cmd)
+  -- Open the command, for reading
+  local fd = assert(io.popen(cmd, 'r'))
```
boucman: please replace io.popen with the DT specific equivalent (yield("READ_FILE") IIRC

the DT version does not block other lua threads while it is running

boucman:
That's a great feature to have, awesome...

please see my comments, I don't have time to test here but I put a couple of comments in the code

contrib/autostyle.lua
marco44: Is it the popen or the following read on which I must do a yield ? sorry but I'm still a bit of a beginner with LUA.

I thought I should wait on file_readable…

Something like 

local fd = assert(io.popen(cmd, 'r'))
coroutine.yield("file_readable",fd)
local data = assert(fd:read('*a'))

Where am I mistaken ?

contrib/autostyle.lua
boucman: yes, I mean the asserts... normal users might not look at stdout and this is an error case a normal user can reach

contrib/autostyle.lua
boucman: you're right it's the read that blocks, so it's the read that needs a yield

marco44:
I just committed the asked changes

contrib/autostyle.lua
```diff
@@ -0,0 +1,130 @@
...
+  return data
+end
+
+-- Registering events
+darktable.register_event("shortcut",autostyle_apply,
+       "Apply your chosen style from exiftool tags")
+
+darktable.preferences.register("autostyle","exif_tag","string","Autostyle: EXIF_tag=value=>style","apply a style automatically if an EXIF_tag matches value. Find the tag with exiftool","")
+
+darktable.register_event("post-import-image",autostyle_apply_one_image)
```
boucman: this won't work. when called on the event, autostyle_apply_one_image will get the event name as its first parameter and the image as the second

so it will have "post-import-image" (the string) as its "image" parameter, leading to errors

please fix

marco44:
I corrected the  "this won't work. when called on the event, autostyle_apply_one_image will get the event name as its first parameter and the image as the second

so it will have "post-import-image" (the string) as its "image" parameter, leading to errors" problem.

Still I get a warning:
LUA WARNING locking from the gui thread should be avoided

I don't understand this message. If you know where it comes from, please tell me. I'll try to find this on my own later if not.

boucman:
That warning is something I need to correct, it's an error that's due to a change in the way threading works in gtk3 vs gtk2.x you don't need to fix that. I'll quickly review and merge if everything's fine

contrib/autostyle.lua
```diff
@@ -0,0 +1,135 @@
...
+  for _,image in pairs(images) do
+    autostyle_apply_one_image(image)
+  end
+end
+
+-- Retrieve the attribute through exiftool
+function exiftool_attribute(path,attr)
+  local cmd="exiftool -" .. attr .. " '" ..path.. "'";
+  local exifresult=get_stdout(cmd)
+  local attribute= assert(string.match(exifresult,": (.*)"), "Could not find the attribute " .. attr .. " using the command: <" .. cmd .. ">")
```
boucman: if the attribute is not found, this results in a lua error which ends up on stderr. A missing attribute is not a problem per-se (all attributes are not guarenteed to be on all images) so you probably don't want to assert on that. A simple message on the lua log would make more sense

boucman:
ok, one last tweak about an assert when an image doesn't have a given tag and it's good to go

marco44:
Ok. One thing I couldn't find: how can I abort the lua script, as I will be removing this assert ?

boucman:
abort as in "finish the current processing event" ? 

the cleanest way to do it would be to raise an error (which is more or less what assert does) but catch it at the top level (your _on_event function) using pcall() check that the error is the one you expected and if it's not re-raise it. if it's the expected error, just finish the function call

marco44:
I think what I just committed should be ok. Thanks for your mentoring anyway :)

PR #5, opened 2015-02-18
cr2hdr.lua: Magic Lantern Dual ISO processing.

ttill:
This script automates the steps involved to process an image created with the Magic Lantern Dual ISO module.  Upon invoking the script with a shortcut "cr2hdr" provided by Magic Lantern is run on the selected images. The processed files are imported. They are also made group leaders to hide the original files.

Will make working with Dual ISO a lot easier until #10170 is fixed.

### DIFF
--- a/contrib/cr2hdr.lua
+++ b/contrib/cr2hdr.lua
@@ -1,0 +1,79 @@
+--[[
+    Copyright (C) 2015 Till Theato <theato@ttill.de>.
+
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 3 of the License, or
+    (at your option) any later version.
+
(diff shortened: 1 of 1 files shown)


contrib/cr2hdr.lua
```diff
@@ -0,0 +1,113 @@
...
+local darktable = require "darktable"
+
+-- Tested with darktable 1.6.2
+darktable.configuration.check_version(...,{2,0,2})
+
+local queue = {}
+local processed_files = {}
+local job
+
+function file_imported(event, image)
```
boucman: functions can be made local too, for the sake of cleanness

boucman:
Ok, I can't really test myself but this looks good...

I will merge the PR but we intend to have native (C side) support for dual-iso at some point... When this happens I might remove this script to avoid confusion

PR #6, opened 2015-07-10
Fix description and license. Selection uses action_images now. Added …

chrik5:
…GUI messages for copy, attach and detach.


boucman:
As a side note, you can now do libs/UI elements in trunk... so if you want to have an actual button to complement the shortcut, you can do that...

PR #7, opened 2015-07-31
Add storage hugin panorama export

chrik5:
I started off with the script from darktable lua wiki http://www.darktable.org/redmine/attachments/download/835/hugin.lua and hope the copyright part is okay.

### DIFF
--- a/contrib/hugin.lua
+++ b/contrib/hugin.lua
@@ -1,0 +1,66 @@
+--[[
+  Hugin storage for darktable 
+
+  copyright (c) 2014  Wolfgang Goetz
+  copyright (c) 2015  Christian Kanzian
+  
+  darktable is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+  
+  darktable is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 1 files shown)


boucman:
agreed on @supertobi's points... can you respin ?

contrib/hugin.lua
```diff
@@ -0,0 +1,66 @@
...
+]]
+
+--[[
+HUGIN
+Add a new storage option to send images to hugin. 
+Images are exported to darktable tmp dir first. 
+
+USAGE
+* require this file from your main luarc config file.
+
```
boucman: you might want to add a comment about needing hugin installed (and which version)

contrib/hugin.lua
```diff
@@ -0,0 +1,66 @@
...
+This plugin will add a new storage option and calls hugin after export.
+]]
+
+dt = require "darktable"
+
+-- should work with lua version 2.0.0
```
boucman: it's not the lua version that we check, it's the DT API version...

please correct/remove the comment i'm afraid it might confuse other readers

contrib/hugin.lua
```diff
@@ -0,0 +1,66 @@
...
+local img_list
+ 
+dt.register_storage("module_hugin","Hugin panorama",
+   function(storage, image, format, filename,
+        number,total,high_quality,extra_data)
+        dt.print("Export to hugin " .. tostring(number).."/"..tostring(total))
+   end,
+   function(storage,image_table,extra_data) --finalize
+        -- reset and create image list
+        img_list = ""
```
boucman: you could declare "local img_list" here, and avoid having a global variable

PR #13, opened 2015-09-04
Small script that fixes lens information for some third party Pentax lenses

GHswitt:


boucman:
The script seeems correct but.... such a fix probably belongs in DT core, not as a lua script... maybe even directly in exiv2.

Replacing lens data in lua makes sense for non-electronic lens that don't provide any exif data at all...

Either I don't understand exactly your use-case, or you should provide a fix in one of those two places. Please ask on the DT mailing list how to do that, there are people in the know there...

GHswitt:
I'll try asking on the mailing list. The reason I'm using the scripts is that the some lens ID codes are used by multiple lenses... I don't think all can theses fixes can by done in exiv2 because one has to know which lens was used. I'm using the script on images I know the lens they were taken with. And they were also meant as example.

PR #16, opened 2015-09-22
Renamed to video_mencoder

supertobi:
Renamed, so we can make video export scripts with other encoders

### DIFF
--- a/contrib/video_mencoder.lua
+++ b/contrib/video_mencoder.lua
@@ -42,15 +42,18 @@ local function checkIfBinExists(bin)
     local result = handle:read()
     local ret
     handle:close()
-    if (not result) then
+    if (result) then
+        --dt.print_error("true checkIfBinExists: "..bin)
+        ret = true
+    else
         dt.print_error(bin.." not found")
         ret = false
     end
-    ret = true
+  
     return ret
 end
 
-local function create_video(storage, image_table, extra_data)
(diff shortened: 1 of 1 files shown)


boucman:
I'm ok, with the patch, but a couple of notes
* once 2.0 is out I will branch the lua script repository, so please avoid renaming options and/or files in that branch, because it will break user's scripts
* We now can add UI elements in the export section of DT's GUI. there is documentation in the git usermanual and i'm availalble on IRC if you want help. It would probably be nicer than preferences.

I

houz:
Why not just add a list of supported encoders to the preferences so that people can just pick which one they want?

supertobi:
I renamed it because MEncoder is no longer in the Ubuntu repository. Thats because mencoder was part of mplayer but they switched to mplayer2 and mplayer2 doesn't include mencoder.
That means I can't test it any longer and because of that I'll not maintain this script any longer.

I looked at some other encoders  (ffmpeg/avconv, g'mic, gst-launch-1.0, transcode) and as far as I have seen non of them supports a list of single images. They all use some kind of filename pattern like "image%d.jpg" or "*.jpg" for the input images.
Here I could need some help how to solve this best.

So now to your question houz:
 * I don't like to have code in my script, that I can't test.
 * The different encoders don't support the same codecs and I have a codec selector in the preferences.

PR #18, opened 2015-12-23
update check_version to {3,0,0}

itarozzi:
fix check_version() against 3.0.0 as discussed in dev mailing list


chrik5:
Hi,

Will the scripts still work with older dt versions?

If not:
My copy_attach_detach_tags works since API 2.

So setting check_version to something like this
dt.configuration.check_version(...,{2,0,0},{3,0,0})
will be more appropriate?

Christian

boucman:
yes, if you don't change anything you should append the newly tested api version at the end of the list of tested versions, not replace the list.... 

did you check the scripts entirely ? the main non-compatibility is that calls to yield now MUST have the type of yield parameter written in upper case

boucman:
I will close this pull requests since we are checking scripts one by one... thx for the work anyway...

PR #19, opened 2015-12-28
make check_api_version work with API 3

chrik5:
Hi,

For me this script work with API 3 without any issues.
I do not use any yield command.

Christian


boucman:
this needs to be backported to the 2.0.x branch,I need to formalize a process for that

PR #24, opened 2016-01-04
Update from my fork

supertobi:
New script to open an image geo location in gnome-maps: geo_uri.lua
Update of slideshowMusic.lua for darktable 2.0

### DIFF
--- a/contrib/geo_uri.lua
+++ b/contrib/geo_uri.lua
@@ -1,0 +1,96 @@
+--[[
+    This file is part of darktable,
+    Copyright 2014 by Tobias Jakobs.
+
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 3 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+]]
+--[[
+darktable script to open a geo uri in gnome-maps
+
+ADDITIANAL SOFTWARE NEEDED FOR THIS SCRIPT
+* gnome-maps >= 3.20
+
+WARNING
+This script is not testet.
+
+USAGE
+* require this script from your main lua file
+* register a shortcut
+]]
+
+local dt = require "darktable"
+dt.configuration.check_version(...,{2,0,1})
+
+local function checkIfBinExists(bin)
+    local handle = io.popen("which "..bin)
+    local result = handle:read()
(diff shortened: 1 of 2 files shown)


contrib/geo_uri.lua
```diff
@@ -0,0 +1,92 @@
...
+WARNING
+This script is not testet.
+
+USAGE
+* require this script from your main lua file
+* register a shortcut
+]]
+
+local dt = require "darktable"
+dt.configuration.check_version(...,{2,0,1})
```
boucman: this won't work with api 3.0.0 (dt 2.0)

contrib/geo_uri.lua
```diff
@@ -0,0 +1,92 @@
...
+    return
+  end	
+    
+  local sel_images = dt.gui.selection()
+  
+  local lat1 = 0;
+  local lon1 = 0;
+  local i = 0;
+  
+  local sel_images = dt.gui.selection()
```
boucman: this variable is defined twice (see line 54)

contrib/geo_uri.lua
```diff
@@ -0,0 +1,92 @@
...
+  
+  -- Use the first image with geo information
+  for _,image in ipairs(sel_images) do
+    if ((image.longitude and image.latitude) and 
+        (image.longitude ~= 0 and image.latitude ~= 90) -- Sometimes the north-pole but most likely just wrong data
+       ) then
+        
+        i = i + 1;
+        
+        if (i == 1) then
```
boucman: please use the "break" statement instead of the variable "i" it will be cleaner

contrib/geo_uri.lua
```diff
@@ -0,0 +1,92 @@
...
+          lat1 = image.latitude;
+          lon1 = image.longitude;
+        end
+  
+      end
+    end
+
+    if (lat1 and lon1) then
```
boucman: these variables are initialized to 0 (and not nil)

in lua, 0 is true, (only nil and false are false. 0 and "" are true) so this expression is always true

contrib/geo_uri.lua
```diff
@@ -0,0 +1,92 @@
...
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+]]
+--[[
+darktable script to open a geo uri in gnome-maps
+
+ADDITIANAL SOFTWARE NEEDED FOR THIS SCRIPT
+* gnome-maps >= 3.20
+
+WARNING
+This script is not testet.
```
boucman: then please test it before I merge it :)

contrib/geo_uri.lua
```diff
@@ -0,0 +1,92 @@
+--[[
+    This file is part of darktable,
+    Copyright 2014 by Tobias Jakobs.
```
boucman: 2014 ?

boucman:
interesting script overall, and a nice one to have... i'd rather have the update to slideshow.lua be a different pull request, but it doesn't really matter, i'll cherry pick those commits to the stable branch separately...

supertobi:
Lets close this. I'll clean this up.

PR #30, opened 2016-01-05
exiftool on export UI box to manipulated metadata of currently export…

chrik5:
…ed images - first version

Basicaly I needed to copy keywords from xmp to IPTC. I have added some more options, which might be useful. I hope the script makes sense in some way.

AFAIK there is no way to store presets in the darktable lua API?

Christian


supertobi:
123 commits and 38 files changed. Are you sure this pull request is right?

chrik5:
No! I updated my branch on my fork. What happend here?
I will close it.

PR #38, opened 2016-01-22
Move example scripts in a new folder to seperate them from the 'usefull' scripts

supertobi:


boucman:
awesome, thx for doing that...

PR #39, opened 2016-01-22
added gettext to kml_export.lua 

supertobi:


boucman:
since this is the first script using translations we neeed to be sure to use a coding style that will be ok and reuse it in all other scripts... (ok, technically the second with the example script)

* you should create a local alias 
```lua
local gettext = darktable.gettext
```

* you should create an _ alias or use one of the functions in 

https://www.gnu.org/software/gettext/manual/html_node/Lua.html

this way xgettext can correctly extract strings from the source

your .po file seems perfectly generated though, so you might use another process to extract the strings, in which case I would be very interested in knowing how you do that...

supertobi:
> * you should create an _ alias or use one of the functions in 

Could you change the example script, so I can see how it should look like? That would be grate. 

> your .po file seems perfectly generated though, so you might use another process to extract the strings, in which case I would be very interested in knowing how you do that...

The file is handwritten.

boucman:
I have no DT time right now, but if you do what I said earliery (i.e declare gettext as a local variable than use the functions detailes in the webpage like gettext.dgettext) then you can run xgettext on your script and that will generate the file from you

you need to have the functions named exaclty like they are named in the gettext manual because xgettext does pattern-matching. it doesn't look at the lua syntax itself....

in the same way, if you define a local function called _ and have it call dgettext with the proper domain, you can use that function and make your code more readable. xgettext will recognize that too...

supertobi:
please review

PR #42, opened 2016-01-30
Update kml_export.lua - use os.remove(image)

supertobi:
use os.remove(image)


boucman:
that sound like a regression to me... yield is the proper way to do this...

supertobi:
Why us yield better then os.remove()?

boucman:
hmm, no you're right os.remove should be ok. yield has the benifit of not blocking the lua thread, but unless you remove a network file, that should be fine

supertobi:
OK, fine, then please push

PR #55, opened 2016-04-07
External editing with gimp

wpferguson:
Created contrib/gimp.lua which exports an image(s) then opens them with gimp for editing.  Edit, then save the changes back to the exported file. When gimp exits the external files will be merged back into darktable and grouped with the original files.


contrib/gimp.lua
```diff
@@ -0,0 +1,179 @@
...
+  else
+    dt.print_error(bin.." not found")
+    ret = false
+  end
+  return ret
+end
+
+local function show_status(storage, image, format, filename,
+  number, total, high_quality, extra_data)
+  dt.print("Exporting to gimp "..tostring(number).."/"..tostring(total))
```
supertobi: I had a fist look and one thing that could be improved is to mark this string for translation.
Have a look at "kml_export.lua" line 72:
https://github.com/darktable-org/lua-scripts/blob/master/contrib/kml_export.lua#L72 
> dt.print(string.format(_("Export Image %i/%i"), number, total))

contrib/gimp.lua
```diff
@@ -0,0 +1,179 @@
...
+  for _,v in pairs(image_table) do
+    local fname = get_filename(v)
+    for _,w in pairs(dt.gui.action_images) do
+      if basename(fname) == basename(w.filename) then
+        os.execute("mv "..v.." "..w.path)
+        local myimage = dt.database.import(w.path.."/"..fname)
+        for _,x in pairs(dt.tags.get_tags(w)) do 
+          if not (string.sub(x.name,1,9) == "darktable") then
+            dt.tags.attach(x,myimage)
+          end
```
supertobi: Longer variable name then v, w and x would improve the readability of the code.

boucman:
I don't know if reimporting from gimp really makes sense... you loose the non-destructive aspect of DT and DT is not that usefull to deal with jpg or other formats exported by gimp... but please explain your workflow, I might change my mind on the subject....

I think it would be more interesting to have a gimp exporter that could
* open all the images exported
* or export all images as a single multi-layerd xcf
* or use the group info from DT to regroup images

you can add widgets to exporters in master if you are intereted in looking into that

wpferguson:
I import my raws into Darktable, then process them.  Once I've taken them as far as Darktable can go, then I take them to gimp for further processing, such as cloning out distracting objects or doing a wavelet decompose for skin smoothing.  By using the export, I get all the processing I've done to that point brought into gimp.  In the past, the gimp result ended up in an export directory that I couldn't manage from Darktable, so I had to access the file manually.  Importing the file back into Darktable and grouping it with the raw lets me find it easily.  I also apply the tags from the raw to the gimp export so that I can search for it with collect.  Once it's back in Darktable, I can also apply further edits by duplicating and converting to monochrome or split toning or...  I look at exporting to gimp as a final, or near final, step in the workflow.

This program will open multiple images at one time.  During development I was doing 5 at a time, but the only limit should be available memory.  You can also do a tag search and simultaneously open images from different folders in gimp.  When it closes, the images will be grouped with the originals in the original's folder.

I wrote the multi layer version last night.  It adds a new export, 'Edit as layers in gimp'.  I stacked three images and took the sky from one, the land from another and the water from another.  It brings the result back as stack.tif.  

I don't understand 'use the group info from DT to regroup images'.  Perhaps you could give me an example so that I could wrap my head around it.

I'm still a lua noob, so I don't want to mess with the exporter just yet.  Adding new targets is one thing, but adding widgets is another.

wpferguson:
When I originally tried the string.format I kept get errors about my arguments.  Looking at it again, I miscoded it.  Should have done a copy and paste instead of trusting my reading and typing skills.  It's back in.  Thanks

contrib/gimp.lua
```diff
@@ -0,0 +1,179 @@
...
+  -- find the matching original image
+  -- then move the exported image into the directory with the original  
+  -- then import the image into the database which will group it with the original
+  -- and then copy over any tags other than darktable tags
+
+  for _,exp_img in pairs(image_table) do
+    local fname = get_filename(exp_img)
+    for _,orig_img in pairs(dt.gui.action_images) do
+      if basename(fname) == basename(orig_img.filename) then
+        os.execute("mv "..exp_img.." "..orig_img.path)
```
supertobi: I've just tested the script and it failed because my orig_img.path has a space in it. I think you should use os.rename here.

http://www.lua.org/manual/5.1/manual.html#pdf-os.rename

contrib/gimp.lua
```diff
@@ -0,0 +1,179 @@
...
+          end
+        end
+      end
+    end
+  end
+
+end
+
+-- Register
+dt.register_storage("module_gimp", _("Edit with gimp"), show_status, gimp_edit)
```
supertobi: The GIMP team like to have the name written in all upper case: "GIMP" I think it would be nice to respect this and write GIMP in all user visible strings.

contrib/gimp.lua
wpferguson: Thanks, good catch.  os.rename() also handles special characters in the path and file name.

contrib/gimp.lua
wpferguson: Done

supertobi:
I'm not a git expert, but now it looks like we have contrib/gimp.lua and gimp.lua in this pull request. 
wpferguson could you please try to fix that.

gimp.lua
```diff
@@ -0,0 +1,186 @@
...
+   return split(str,'[\\/]+')
+end
+
+local function get_filename(str)
+  parts = split_path(str)
+  return parts[#parts]
+end
+
+local function basename(str)
+  return string.sub(str,1,-4)
```
supertobi: I think this only works if the file extension have the same length in the original file and the in gimp edited one. e.g. If the original file is a .raw and the one edited a .tiff or .jpeg. 

Have a look at this code that should work better:
https://github.com/darktable-org/lua-scripts/blob/master/contrib/kml_export.lua#L127

gimp.lua
```diff
@@ -0,0 +1,186 @@
...
+  -- for each of the exported images
+  -- find the matching original image
+  -- then move the exported image into the directory with the original  
+  -- then import the image into the database which will group it with the original
+  -- and then copy over any tags other than darktable tags
+
+  for _,exp_img in pairs(image_table) do
+    local fname = get_filename(exp_img)
+    dt.print_error(fname)
+    for _,orig_img in pairs(dt.gui.action_images) do
```
supertobi: And here I had the problem, that dt.gui.action_images chanced if I move the mouse above other images. I was simply not able to get the edited image grouped to the old one. For me it looks like a bug in dt. :( At least it is different then in the user manual described:
https://www.darktable.org/usermanual/ch09.html.php
https://www.darktable.org/lua-api/index.html.php#darktable_gui

But I think it would be even better to avoid the double for loop with this:
> for image,exported_image in pairs(image_table) do

gimp.lua
wpferguson: That is an awesome regular expression.  I had to fire up lua in a shell and play with it for awhile to figure it out.  Well done, and thanks.

gimp.lua
wpferguson: I didn't realize what the other half of the image_table was, so I used dt.gui.action_images to get what was already in the image_table.  I should have read kml_export more carefully.  That cleaned up the code nicely.  Thanks

chrik5: "dt.gui.action_images" changes all the time. That is not a bug. 

If you look at the samples - current action_images will always be assigned to variables before something is done with the list of images "local images = darktable.gui.action_images".

supertobi: Then there is a bug in the manuel, have a look at this lines from https://www.darktable.org/usermanual/ch09.html.php :

darktable.gui.action_images is a table containing all the images of interest. darktable **will act on selected images if any image is selected**, and on the image under the mouse if no image is selected. This function allows to easily follow darktable's UI logic in lua. 

Perhaps boucman can check this in the darktable code.

contrib/gimp.lua
```diff
@@ -0,0 +1,154 @@
...
+  --   move the exported image into the directory with the original  
+  --   then import the image into the database which will group it with the original
+  --   and then copy over any tags other than darktable tags
+
+  for image,exported_image in pairs(image_table) do
+
+    local myimage_name = image.path .. "/" .. get_filename(exported_image)
+
+    dt.print_error("moving " .. exported_image .. " to " .. myimage_name)
+    result = os.rename(exported_image, myimage_name)
```
supertobi: One thing that I thought about and haven't tested is: What happens if you import e.g. a jpeg in dt and then edit it with gimp as jpeg. Will you then overwrite the original file? If yes, you could do something like renaming the in gimp changed from FILENAME.jpg to FILENAME_gimp.jpg or FILENAME_1.jpg.

wpferguson: The way it stands now, if you exported a raw to gimp as a .jpg, then exported the resulting .jpg to gimp, the .jpg would get overwritten when gimp exits.  I could do the xxxx.jpg, then xxxx_01.jpg.  The only question is if I export _01.jpg does it become _02.jpg or _01_01.jpg?

I also developed a shortcut for the occasion when you did want to overwrite the file, i.e. I exported and edited, but forgot to clone out one more thing so I'll just reopen it and fix it.  While running that I discovered an unfortunate darktable "feature" (bug).  If I've not opened the image in darkroom, then dropping the cache causes it to regenerate the thumbnail and it shows correctly.  If I open the image in darkroom, it does not reflect the changes, so it seems that darkroom also maintains a cache.  When you return to lighttable, the thumbnail gets regenerated again and shows the image without any changes, so it appears to regenerate from the darkroom cache.  If you open it with gimp again and make more changes then save it, the thumbnail cache gets dropped, but it regenerates the image from the darkroom cache and not the file.  The only way to get things synced up is to restart darktable.

contrib/gimp.lua
```diff
@@ -0,0 +1,247 @@
...
+    dt.print_error("true checkIfBinExists: "..bin)
+    ret = true
+  else
+    dt.print_error(bin.." not found")
+    ret = false
+  end
+  return ret
+end
+
+local function checkIfFileExists(filepath)
```
supertobi: Eek, this doesn’t work with file names with spaces the "ls" fails. Use this function, I've just tested it and it works with the spaces:

```
function checkIfFileExists(filepath)
  local file = io.open(filepath,"r")
  local ret
  if file ~= nil then 
    io.close(file) 
    dt.print_error("true checkIfFileExists: "..filepath)
    ret = true
  else 
    dt.print_error(filepath.." not found")
    ret = false
  end
  return ret
end
```

contrib/de_DE/LC_MESSAGES/gimp.po
```diff
@@ -0,0 +1,39 @@
...
+msgid "GIMP not found"
+msgstr "GIMP nicht gefunden"
+
+#: gimp.lua:198
+msgid "Launching GIMP..."
+msgstr "Starten von GIMP"
+
+#: gimp.lua:243
+msgid "Edit with GIMP"
+msgstr "Bearbeiten mit Gimp"
```
supertobi: You missed one Gimp:

msgstr "Bearbeiten mit GIMP"

contrib/gimp.lua
wpferguson: Just created ~/Pictures/Darktable/test data/path with a space in it/ and populated it with every kind and variation of messed up filename I can think of.  Thanks for the fix.

supertobi:
Done. This is now really a nice script.

wpferguson:
Thank you for the tremendous amount of time and help!

PR #61, opened 2016-04-27
Import filters fix

spaceChRiS:
Fix accidentally deleted first image during import. Lua is very picky about sparse table handling. It is even a difference for different densities of the sparse table. Especially, `table.remove` does only remove the last entry if the table is a “dense” sparse table, for others it will remove the last element of the first contiguous sequence of integer indices (I guess it is even more complicated).

### DIFF
--- a/official/import_filters.lua
+++ b/official/import_filters.lua
@@ -1,6 +1,6 @@
 --[[
     This file is part of darktable,
-    copyright (c) 2015 Tobias Ellinghaus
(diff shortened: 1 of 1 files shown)


supertobi:
Please add a copyright line with your name to the script and perhaps add some comments to the code.

spaceChRiS:
You mean something like that? I find it somehow presumptuous to add my name to Tobias script :-/

supertobi:
Yes, exactly like this.

Issue #59, opened 2016-04-20
kml_export - read global options and use it for ImageMagick


elman22:
Current implementation uses fixed size for ImageMagick conversion

> local concertCommand = "convert -size 438x438 "..exported_image.." -resize 438x438 +profile \"*\" "..exportDirectory.."/"..imageFoldername..filename..".jpg"

Could you read settings under export selected -> global options -> max size and use it as variable for ImageMagick?

elman22:
I was just thinking that actually there might not be a need to call ImageMagick twice. First conversion is fine - you create thumbnail. But instead of second conversion you could simply use image that was exported by darktable itself. And this image will have format and resolution you specify in export selected -> global options -> max size.

supertobi:
Patches are welcome. Lua is not that hard. ;)

elman22:
OK, here is my patched version. I did these changes:
- Fixed issue with exporting ARW (or any other than JPG format). Simply I strip extension instead of changing it into suffix separated by underscore.
- Removed second call to ImageMagick and I use image generated by darktable. This means that  user can choose image format, size and quality. Tested with jpg and png.Tiff is not working because GoogleEarth will not display it.
- Fixed bug with invalid XML when & was used in image.title or description. & is now encoded.
- Changed baloon style to be more user friendly and now it displays title and description.

[kml_export-elman.zip](https://github.com/darktable-org/lua-scripts/files/232436/kml_export-elman.zip)

supertobi:
Coll thanks for the changes. I'm not at home at the moment and don't have a Linux PC here for testing. But if you could make a git pull request instead of adding a ZIP file here it would make the review process much easier. 

> Removed second call to ImageMagick and I use image generated by darktable. This means that user can choose image format, size and quality. Tested with jpg and png.Tiff is not working because GoogleEarth will not display it.

I think we should then add a warning, if something else then PNG or JPEG is selected. 

> Fixed bug with invalid XML when & was used in image.title or description. & is now encoded.

You should move this into function like the stripAccents( str ) function. I think we have to escape other characters, like < and >, too.

https://stackoverflow.com/questions/1091945/what-characters-do-i-need-to-escape-in-xml-documents

elman22:
OK, I have forked the repo and will make pull request when I implement your comments.

Issue #58, opened 2016-04-20
kml_export - wrong handling of image formats other than jpg


elman22:
How to reproduce:
1. Open ARW file in darktable, e.g. DSC10661.ARW
2. Select KML/KMZ export and use format options JPEG
3. Press export

Exported image is saved as JPEG and ImageMagick converts it to JPEG, e.g. DSC10661_JPG.jpg is created. kml_export expects to zip file DSC10661_ARW.jpg and fails (also name DSC10661_ARW.jpg is used in kml and references non existing image)

If I change format options to TIFF. Exported image is saved as TIFF and ImageMagick converts it to JPEG, e.g. DSC10661_TIF.jpg is created. kml_export expects to zip file DSC10661_ARW.jpg and fails.

So kml_export should check format options and expect correct file name.

supertobi:
I think I must use exported_image and not image here:
https://github.com/darktable-org/lua-scripts/blob/master/contrib/kml_export.lua#L332

I'll have a closer look to it later.
@elman22 If you can would give me your email address I could send you test versions.

elman22:
I fixed it in issue #59 . See attachment in my post.

Issue #54, opened 2016-04-01
kml_export - unable to open kmz in Google Earth (Linux)


elman22:
Problem is that I use diacritics in folders and when I import film roll into darktable, it also has diacritics in the name. Consequently when kml is generated, name with diacritics is used. If this is the case, Google Earth will not open kmz. 

Maybe it's Google's bug, but you could strip diacritics from kml name.

Error - Etiópia 2016.kml
Working - Etiopia 2016.kml

Thanks

supertobi:
I don't know when I'll find the time to fix this. But if you are interested in fixing this yourself this links may help:

http://lua-users.org/lists/lua-l/2007-04/msg00346.html
https://forums.coronalabs.com/topic/43048-remove-special-characters-from-string/
http://lua.2524044.n2.nabble.com/removing-diacritical-marks-from-strings-td7626507.html

supertobi:
I've just tested it with Google Earth 7.1.2.2041 and it works with the filename "Etiópia 2016.kml". What is your Google Earth version?

elman22:
Hi, mine is Google Earth 7.1.4.1529, so it looks like it's newer - I use the one from AUR.
BTW sorry about not answering to your previous comment. I couldn't find time to study it as well :(

supertobi:
I've tested it again. It works if you open the kml file with the file open dialogue but not if you use the command line. 

I think I've fix the problem, please test.

elman22:
Unfortunately no kmz is generated. I see that thumbnails are generated, but they remain in ~/.local/tmp.

supertobi:
You need the version before the last changes from boucman:
https://github.com/darktable-org/lua-scripts/blob/204868d73369673a5c7e992aa27926e463d349b3/contrib/kml_export.lua

elman22:
Well this version does create kmz (Google Earth opens it when I check "Open file after export"), but it doesn't place it in folder I specified in settings.

supertobi:
That's strange. I've tested it now several times and it works here fine. What is the name of the output folder in the settings, and where does the script place the KMZ files?

elman22:
I use folder 
lua/kml_export/ExportDirectory=/home/elman/Pictures/Travelling

supertobi:
Are there any error messages, if you call darktable from the command line with:
`darktable -d lua`

elman22:
When exporting 7 ARW images, I got KMZ with just KML file and no images

[...]

PR #62, opened 2016-04-27
Fix for issues #58 and #59 - Issue with exporting RAW images

elman22:
- Fixed issue with exporting ARW (or any other than JPG format). Simply I strip extension instead of changing it into suffix separated by underscore.
- Removed second call to ImageMagick and I use image generated by darktable. This means that user can choose image format, size and quality. Tested with jpg and png.Tiff is not working because GoogleEarth will not display it.
- Fixed bug with invalid XML when & was used in image.title or description. & is now encoded.
- Changed baloon style to be more user friendly and now it displays title and description.

### DIFF
--- a/contrib/kml_export.lua
+++ b/contrib/kml_export.lua
@@ -30,6 +30,7 @@ WARNING
 This script is only tested with Linux
 
 USAGE
+* require script "official/yield" from your main Lua file in the first line
 * require this script from your main Lua file
 * when choosing file format, pick JPEG or PNG as Google Earth doesn't support other formats
 
@@ -202,7 +203,7 @@ local function create_kml_file(storage, image_table, extra_data)
         -- Creates dir if not exsists
         imageFoldername = "files/"
         local mkdirCommand = "mkdir -p "..exportDirectory.."/"..imageFoldername
-        coroutine.yield("RUN_COMMAND", mkdirCommand) 
+        dt.control.execute(mkdirCommand) 
     end
 
 
@@ -222,8 +223,7 @@ local function create_kml_file(storage, image_table, extra_data)
             --	profiles that might be present in the input and aren't needed in the thumbnail.
 
             local convertToThumbCommand = "convert -size 96x96 "..exported_image.." -resize 92x92 -mattecolor \"#FFFFFF\" -frame 2x2 +profile \"*\" "..exportDirectory.."/"..imageFoldername.."thumb_"..filename..".jpg"
-            -- USE coroutine.yield. It does not block the UI
-            coroutine.yield("RUN_COMMAND", convertToThumbCommand)
+            dt.control.execute(convertToThumbCommand)
         else
             -- Remove exported image if it has no GPS data
             os.remove(exported_image)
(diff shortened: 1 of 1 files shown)


supertobi:
Please add a copyright line with your name to the script.

contrib/kml_export.lua
```diff
@@ -348,7 +365,7 @@ local function create_kml_file(storage, image_table, extra_data)
         else
             kmlFileOpenCommand = "xdg-open "..exportDirectory.."/\""..exportKMLFilename.."\""
 	end
-        dt.control.execute( kmlFileOpenCommand) 
+        coroutine.yield("RUN_COMMAND", kmlFileOpenCommand)
```
supertobi: @boucman 
What do you think about reverting your changes here?

contrib/kml_export.lua
```diff
@@ -236,28 +248,32 @@ local function create_kml_file(storage, image_table, extra_data)
     kml_file = kml_file.."    <description>Exported from darktable</description>\n"
     
     for image,exported_image in pairs(image_table) do
-	filename = string.upper(string.gsub(image.filename,"%.", "_"))
+    filename = string.upper(string.gsub(image.filename,"%.%w*", ""))
+    extension = string.match(exported_image,"%.%w*$")
```
supertobi: I think a comment would not hurt here.

contrib/kml_export.lua
```diff
@@ -155,6 +156,19 @@ function string.stripAccents( str )
...
+
+    str = string.gsub(str,"&", "&amp;")
+    str = string.gsub(str,"\"", "&quot;")
+    str = string.gsub(str,"'", "&apos;")
+    str = string.gsub(str,"<", "&lt;")
+    str = string.gsub(str,">", "&gt;")
+
+    return str
+end
+
```
supertobi: This function is really nice, I like it.

contrib/kml_export.lua
chrik5: What about an if clause for compatibility?

supertobi: Last time I talked with @boucman the plan was to add something like the code below to darktable, to avoid that we have an if cause in every script. But I don't know the actual status.

local dt = require "darktable"
local yield_old = coroutine.yield
coroutine.yield = function(yield_type, command)
  if (dt.configuration.api_version_major < 4) then
    yield_old(yield_type,command)
  else
    if (yield_type == "RUN_COMMAND") then
      dt.control.execute(command)
    elseif (yield_type == "FILE_READABLE") then
      dt.control.read(command)
    elseif (yield_type == "WAIT_MS") then
      dt.control.sleep(command)
    end
  end
end

elman22: As a base I was actually using version you suggested in issue #54 because current version with dt.control didn't work on my computer (I think I had darktable 2.0.1 at the time).

supertobi: @elman22 Yes, we know. The version from @boucman only work with the latest git version. We are trying to find a solution that works with both versions without the need to change all scripts

elman22:
Added comments where necessary.

contrib/kml_export.lua
supertobi: @elman22 I've added the script official/yield.lua. If you now download it and then require it in the first line of your luarc  the "dt.control.execute(command)" should work with your version of dt. 
If it works, please redo the changes from @boucman

elman22: Works fine, thanks.
So I reverted changes back to dt.control.execute().

PR #63, opened 2016-05-08
Overview of available scripts

spaceChRiS:
I started to gather information on all available scripts in the repository, because I think such an overview is useful. However, this PR is not meant to be merged directly but is a place where discussion about this proposal can happen. There are several possibilities to continue with an overview documentation besides the trivial ones (merge as is or drop the idea entirely):
1. We keep the manual approach and I fix the issues with the actual state, e.g. by requesting additional information from the authors.
2. We do 1. and feed back improvements on the documentation that arise while I dig through the scripts.
3. I try to write a parser to extract information automatically and try to convince authors to tailor their documentation to be parsed. The automatically generated overview would, however, need manual tweaking.
4. We introduce a standard format for documentation/metadata within scripts to be parsed automatically. The version compatibility is already available as lua variable, similar structures could be introduced for the short description, maintainer info and tags. How that would look like (e.g. as lua variables or as comments in a parsable format) is open for discussion. This would eventually not require manual intervention at all. Furthermore, it would flatten the way to a scripts manager inside Darktable, if something like this will be planned someday.

Any thoughts on this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,9 @@ The available scripts are briefly desctibed subsequently. For the individual scr
 
 ## Official
 
+These scripts are maintained by the darktable project, therefore no individual maintainers are provided. Suitable places for questions and bug reports are the darktable mailing list and bug tracker and the github issues tracker.
+
+
 ### check_for_updates.lua
 
 *Compatibility: 1.x, 2.x*
@@ -21,11 +24,13 @@ The available scripts are briefly desctibed subsequently. For the individual scr
 Automatically look for newer releases on github and inform
 when there is something. It will only check on startup and only once a week.
 
+
 ### copy_paste_metadata.lua
 
 *Compatibility: 2.x*
 
 Copy metadata (title, description, …), rating, colour labels and tags between images by keyboard shortcut or by buttons in the “selected images” lighttable module.
+
 
 ### debug-helpers.lua
 
@@ -33,11 +38,14 @@ Copy metadata (title, description, …), rating, colour labels and tags between 
 
 A collection of helper functions to help debugging lua scripts.
 
+
 ### delete_long_tags.lua
 
 *Compatibility: 1.x, 2.x*
+*Tags: tagging*
 
 Automatically delete all tags longer than a given length.
+
 
 ### enfuse.lua
 
@@ -45,11 +53,13 @@ Automatically delete all tags longer than a given length.
 
 Run enfuse on selected images to merge them into a HDR and make the HDR available in Darktable. Does not work with raw files, only formats understood by enfuse are possible (e.g., jpeg).
 
+
 ### generate_image_txt.lua
 
 *Compatibility: >1.2.x, 2.x*
 
 Run a custom command on images to generate text metadata. This data is stored in a text sidecar file and can be overlaid over the image by darktable.
+
 
 ### image_path_in_ui.lua
 
@@ -57,11 +67,13 @@ Run a custom command on images to generate text metadata. This data is stored in
 
 Add a widget with the paths of all selected images for easy copy/paste.
 
+
 ### import_filter_manager.lua
 
 *Compatibility: unknown*
 
 Adds a dropdown list with import filters to the import dialog. Several import filters can plug into the manager. This allows customized import schemes such as “prefer raw over jpeg” where a jpeg is only added if no corresponding raw file exists. Requires suitable filters, an example set is implemented in “import_filters.lua”.
+
 
 ### import_filters.lua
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,3 +8,267 @@ To enable one of the scripts you have to add a line like `require "official/hell
...
+
+
+### kml_export.lua
+
+*Compatibility: 2.x*  
+*Maintainer: Erik Augustin*  
+*Dependencies: mkdir, zip, convert (ImageMagick), xdg-open, xdg-user-dir*  
+*Tags: geo, storage*
+
+Adds a new export option to export KML files. **I have no clue what it really does.**
```
supertobi: I wrote a blog post about the KML and the geoJSON export script here:
http://dablogter.blogspot.co.uk/2016/03/tutorial-creating-maps-from-geo-tagged.html

README.md
```diff
@@ -8,3 +8,267 @@ To enable one of the scripts you have to add a line like `require "official/hell
...
+Play music during a slide show.
+
+
+### video_mencoder.lua
+
+*Compatibility: 1.x*  
+*Maintainer: Tobias Jakobs*  
+*Dependencies: mencoder, xdg-open, xdg-user-dir*
+
+Video export from darktable. **Some more information would be great.**
```
supertobi: It creates a video from the selected images like this one (Sorry, the text is in German.):
http://dablogter.blogspot.co.uk/2014/04/timelapse-tiltshift-video.html 

I can't use this script any more, because mencoder is no longer part from Ubuntu. :( It should be ported to ffmpeg.

README.md
```diff
@@ -8,3 +8,267 @@ To enable one of the scripts you have to add a line like `require "official/hell
...
+
+
+### geo_uri.lua
+
+*Compatibility: 2.x*  
+*Maintainer: Tobias Jakobs*  
+*Dependencies: gnome-maps ≥ 3.20*  
+*Tags: geo*
+
+Open a geo uri in gnome-maps. **Where is this URI found? What is a typical use case?**
```
supertobi: This will open the location of the image with the standard program for geo uri. E. g. GnomeMaps supports this:
http://jonasdn.blogspot.co.uk/2015/12/whats-cooking-in-maps-in-gnome-320.html
https://en.wikipedia.org/wiki/Geo_URI_scheme

README.md
```diff
@@ -8,3 +8,267 @@ To enable one of the scripts you have to add a line like `require "official/hell
...
+
+
+### geoJSON_export.lua
+
+*Compatibility: 2.x*  
+*Maintainer: Tobias Jakobs*  
+*Dependencies: mkdir, convert (ImageMagick), xdg-open, xdg-user-dir*  
+*Tags: geo*
+
+Generate GeoJSON file from image metadata. **What exactly is exported? What is a typical use case?**
```
supertobi: I think the main usecase is creating maps for the web. But GnomeMaps supports e.g. geoJSON too.

spaceChRiS:
Thanks @supertobi for the explanations. German text is not a problem. I will improve the descriptions as soon as I find some time. Do you have any remarks concerning the fundamental idea? Good/bad? Suggestions?

supertobi:
1. I like the idea of better documentation. 
2. I don't know what's the best way to do it. We are still at a very early state with the scripts and a lot is in change.
3. Go an with your work, but don't get frustrated if we find a better way later.
4. I'd love to see some kind of script manager inside darktable.
5. We should have a closer look at this:
https://blogs.gnome.org/hughsie/2014/06/11/application-addons-in-gnome-software/

boucman:
I think it would be better to code the documentation in lua. There are various way to do that, but simply setting a variable 
metadata = {
name = "script name"
shordesc= "a short description"
}

would be a good way to do it. The "include_all" script could collect/process that variable and reset it before each script, and if the script is included without require_all, a simple global variable would cause no harm. 

Generating the README.md from that data would be pretty trivial too.

If we want an external manifest, I wouldn't go with the gnome format, though... https://github.com/keplerproject/luarocks/wiki/Rockspec-format would be more appropriate for us. 


No, to continue with your PR, I'd rather have the info in the format I described above, a manually generated README.md is too cumbersome...

we can continue discussing that future format here (if you could post a more complete idea than my tiny example) and then move from there

sounds good to you ?

spaceChRiS:
It sounds good to me, however, I do not understand some details of the implementation. Maybe you can enlighten me? But first the homework:
```
metadata = {
    name = "script name" -- not the file name
    version = "1.0" -- optional?
    shordesc= "a short description" -- unicode string without markdown syntax
    desc= "a long description, not a manual replacement" -- with markdown; do we need this? Or better only shortdesc?
    usage = "the manual" -- markdown allowed. Or is this part of desc?
    oscompat = "linux" -- restrictions in OS compatibility, just in case
    tags = "tags for categorization" -- e.g. for filtering all scripts that do geo stuff or that work on tags etc. Better than a single category (e.g. tree order), since the script could e.g. translate between geo metadata and tags
    maintainer = "name <email>" -- for contributed scripts
    license = "license" -- do we need it? Or is it given for the whole repo?
    dependencies = "imagemagick, ..." -- do we need it or is it part of the description, since it will not be possible to resolve dependencies automatically. We are still talking about scripts ... Anyway, I would prefer it like here.
}
```
What do we do with the API compatibility information?
This is more or less a simplified version of the rockspec format, but with the ability to use markdown and an extra field.
Now my questions:

- Do we have the possibility to extract the variable's contents without running the script? Otherwise we may run into troubles if not all dependencies are fulfilled. Especially if we have one linux-only and another osx-only script, we cannot generate the overview.
- Can we make the variable local to the script in a way that the script knows about itself? Maybe difficult since it is still scripts.
- What kind of consensus do we need among the Darktable devs and especially those involved in lua?

I would definitely prefer a solution that does not require an extra file. I would even prefer this for the translation information, since most of the scripts do not have many strings to translate. However, that would change if this metadata format is used.

boucman:
ok, first the homework
* no filename : indeed, the file name can probably be obtained/filed automatically
* version : just to be clear, this is the version of the script, not the DT version needed
* shortdesc : one line description, no markdown (I intend to use it as a tooltip in DT)
* desc+usage : probably one or the other, you decide you know the content better than I do
* oscompat : should be optional. 99% of script will work on all OS anyway
* tags : I'd rather have a single category, since I will use it to sort them in DT menu
* dependencies : this one needs a little thinking. we might want to check that, but you need to load the script to find the dependency. we need a way to solve that loop

now the other questions
* API compatibility : I don't know. We are rethinking api compatibility with @supertobi right now (because I broke yield) so no clear answer yet
* I don't see a way to extract a variable frmo a script without running it. I'll think some more about that
* there are many way to get that to work, we need to choose one. the simplest is that the global require all sets the variable to nil, this way
  in the script
  * metadata = ... ; local metadata_copy =  metadata
in require_all
* metadata = nil
in a secons script
* metatada = <new content>

at this point metadata_copy still contain the script's original metadata


So far I have been the only dev really active on the lua side, and the only heavily involved on the C side of lua, so the question has never really been asked. 

@supertobi helps me maintaining the script repo and I hope to leave more and more of that in his hands, but overall we would be glad to have more contributors looking into the script repo and not just a single script

There are just the three of us.I am sure we can reach a consensus. We will formalize the decision process when we reeach a point where ad-hoc doesn't work for use anymore...

supertobi:
@spaceChRiS are you still working on this? What is the status? Do you need help?

spaceChRiS:
I'd love to, but real life chimed in. Next week I will move to a new house (+ am very busy at work), but after that I hopefully have more time.

There's no new status, just a plan. My next step is to tag some (I thought 3 would be a good number) scripts according to the scheme discussed and add it to the PR. On this branch, a scripting interface and an API from/for within Darktable could be built and tested then (there I would need a lot of help).

Nevertheless, the IMO most important part is not solved: The script has to be run to extract metadata. Therefore, I am still not sure if this is the right way to go, since not all scripts will run on all systems. And it would be a mess to build a dummy API with no functionality just to be able to run the scripts from outside Darktable (e.g. on  a server to generate the web overview page).

Another possibility would be to have the metadata in comments. To parse this from within a script, the script must know about its file system position. Sounds as well as a mess, especially when other OS are considered. Furthermore, formal/syntax errors in the metadata would not automatically be detected, so that a script could break the metadata engine, and additional testing steps would be necessary.

The same holds more or less for extra files containing the metadata, plus it is likely that they will tend to be outdated/forgot to be updated with script releases.

Maybe we could tie a function call to enter the metadata instead of a variable assignment. That way, the metadata could be assigned to the corresponding table and another function could be called at the end that would either jump out of the script after metadata processing if it is run from outside Darktable or that would do nothing if run from Darktable.

What do you think?

supertobi:
I like the idea with the special function call. On the other hand I don't really care about other OS then Linux and I believe most of the scripts are only tested there.
But @boucman is the one with the most knowledge here.

supertobi:
I've just seen, that Ardour has Lua support, and they use a similar approve then boucman suggested:
http://manual.ardour.org/lua-scripting/brain_dump/
https://github.com/Ardour/ardour/tree/master/scripts

supertobi:
And hexchat uses a register method like this:

```
local name = "alias.lua"
local desc = "Create aliases"
local version = "0.1"

hexchat.register(name, version, desc)
```
https://github.com/hexchat/hexchat-addons/tree/master/lua

spaceChRiS:
This is the method that I like best atm, but maybe more like

    dt.register_lua(metadata)

with `metadata` being the table from above. The method has IMHO 2 benefits: throwing a table makes it easily extensible, more metadata fields could be treated if needed and old ones could be abandoned without breaking the functionality, and using a function call makes it easy to define a dummy environment just for read-out of metadata listing. Of course, it would require the register function to be called at the beginning of the script, what makes it a little less appealing, but I have no solution for that.

boucman:
with lua's awesom syntax you van even write

dt.register{<metadata's conter} 

directly

wpferguson:
@boucman you have a little more help.  @supertobi  asked me in the spring if I'd like to help out and I said yes. Real life reared up and I was sidetracked for awhile, but I'm able to do it now.  I've taken care of the enfuse todo list. I asked @supertobi about pulling common functions into a library and he agreed that it should happen, but that I should talk to you.  When I saw this PR it grabbed my interest since documenting the library would be a necessity.  I've read through the thread and looked at the references and thought that maybe the library would be a good test bed.  Back in the spring I coded up a proof of concept for the library and got it working.  I can pull it out, dust it off, and try adding documentation to it to see if I can figure out a way.

I had a thought about pulling the documentation from a script without running it.  Most of the scripts end with either dt.register_lib or dt.register_storage.  If we made a slight coding change to the scripts and we wrap the dt.register_xxx in an activate function so that you could load the script but not actually run it.  If we had a dt.unregister_xxx function we could wrap that with a deactivate function, then we could have a script manager that could turn individual scripts/functions on and off by calling the activate/deactivate functions.

Speaking of a script manager, we currently use the luarc file as a crude and ineffective script manager.  Today I took the luarc file and coded up a slightly less crude and more effective script management solution.  Now when you start darktable, the luarc script looks for the lua directory.  If it doesn't exist, a module (Script Install/Update) is created that gives you a one click solution to install the scripts, by doing a git clone into the users home directory, creating a link and then finding the scripts.  After the scripts are found the install button changes to an update button.  If the scripts are already installed, the module finds the scripts and displays an update button.  Once the scripts are found, a button widget is created for each to Activate/Deactivate the script.  All the button widgets are put into a Script Controller module that gets created.  Once the Script Controller is active, you can load a script by selecting Activate for the desired script.  It loads and creates the storage or module.  The preferences get written, so that the same configuration is available the next time darktable starts.  At this time there isn't a way to unload a lib or storage, so deactivating a module doesn't take effect until the next time darktable is started.

If we add the luarc file to the darktable distribution then we make access to the scripts trivial for the users, since they can click a button to install or update them, then pick and choose what they want.

I still have one function to clean up, which I should have done tomorrow.  Once that happens I'll post a link on here so that you play with it and see what you think.

Now, back to the documentation.  Do we maybe want a Help module that allows displaying script documentation on demand?  I'm not sure how we would display it.  Format it as a pdf and open in a pdf viewer?

boucman:
wow, lots of stuff in there... this is awesome. 

so, first thing first, your system of update is interesting, but deserves its own pull request. I like the idea but i'd rather discuss it when I see the code and in its own thread.

I also like the idea of having a lib for commonly used functions, again proposal wecomed in a new issue or pull request

However I dont think that lib is the right place to experiement with documentation. By its very nature, a lib is a special case and i'd rather use a "normal" case to experiment. Documentation would need to be integrated (somehow) in your script. Something that is usable right now would be to include the documentation text (which are currently a couple of paragraph long) as tooltips in your activation menu. A more complete solution (including some new widget to display the text) would require more work. It's worth considering, but using tooltips allows you to move forward without waiting for me for the time being

wpferguson:
@boucman, the tooltip idea is great.  For now the code is sitting at https://github.com/wpferguson/script-manager.git  I included a screenshot so you can see it in action with a tooltip up.  I'll get started on the pull request.

wpferguson:
Added pull request **lua script management** to darktable.  @spaceChRiS there is a function in script_manager.lua, get_script_doc(script), which pulls the documentation from the existing scripts (assuming that it's the second block comment in the file).

supertobi:

spaceChRiS:
I think I can close this as everything is now well documented inside the scripts, and in the documentation, and we even got the wonderful script manager. Thank you all, I don't think this PR is of any help here anymore.

wpferguson:
Thanks

PR #65, opened 2016-08-23
Added fileCopy and fileMove functions to replace os.rename which does…

wpferguson:
…n't work across filesystems.  Added an error message if the move isn't accomplished so that the user knows what happened.  Updated translations to reflect changes.

### DIFF
--- a/contrib/de_DE/LC_MESSAGES/gimp.po
+++ b/contrib/de_DE/LC_MESSAGES/gimp.po
@@ -26,14 +26,18 @@ msgstr ""
 msgid "Export Image %i/%i"
 msgstr "Exportiere Bild %i/%i"
 
-#: gimp.lua:194
+#: gimp.lua:221
+msgid "Unable to move edited file into collection.  Leaving it as %s"
+msgstr "Kann nicht bearbeitete Datei in Sammlung zu bewegen. Wenn sie unverändert %s"
+
+#: gimp.lua:228
 msgid "GIMP not found"
 msgstr "GIMP nicht gefunden"
 
-#: gimp.lua:209
+#: gimp.lua:243
 msgid "Launching GIMP..."
 msgstr "Starten von GIMP"
 
-#: gimp.lua:254
+#: gimp.lua:290
 msgid "Edit with GIMP"
 msgstr "Bearbeiten mit GIMP"
--- a/contrib/de_DE/LC_MESSAGES/gimp.po
+++ b/contrib/de_DE/LC_MESSAGES/gimp.po
@@ -26,18 +26,18 @@ msgstr ""
 msgid "Export Image %i/%i"
 msgstr "Exportiere Bild %i/%i"
 
-#: gimp.lua:221
+#: gimp.lua:202
 msgid "Unable to move edited file into collection. Leaving it as %s"
 msgstr "Die bearbeitete Datei kann nicht in die Sammlung aufgenommen werden. Bearbeitet Datei ist %s"
 
-#: gimp.lua:228
+#: gimp.lua:209
 msgid "GIMP not found"
 msgstr "GIMP nicht gefunden"
 
-#: gimp.lua:243
+#: gimp.lua:224
 msgid "Launching GIMP..."
 msgstr "Starten von GIMP"
 
-#: gimp.lua:290
+#: gimp.lua:271
 msgid "Edit with GIMP"
 msgstr "Bearbeiten mit GIMP"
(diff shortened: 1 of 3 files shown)


contrib/de_DE/LC_MESSAGES/gimp.po
```diff
@@ -26,14 +26,18 @@ msgstr ""
 msgid "Export Image %i/%i"
 msgstr "Exportiere Bild %i/%i"
 
-#: gimp.lua:194
+#: gimp.lua:221
+msgid "Unable to move edited file into collection.  Leaving it as %s"
+msgstr "Kann nicht bearbeitete Datei in Sammlung zu bewegen. Wenn sie unverändert %s"
```
supertobi: Better:
"Die bearbeitete Datei kann nicht in die Sammlung aufgenommen werden. %s bleibt unverändert in der Sammlung"

contrib/de_DE/LC_MESSAGES/gimp.po
```diff
@@ -26,14 +26,18 @@ msgstr ""
 msgid "Export Image %i/%i"
 msgstr "Exportiere Bild %i/%i"
 
-#: gimp.lua:194
+#: gimp.lua:221
+msgid "Unable to move edited file into collection.  Leaving it as %s"
```
supertobi: "..on.  Le.." 
Please remove the double space.

boucman:
wouldn't it be safer to use the os function "cp" via a command execution ?

wpferguson:
That was the way I originally wrote it, but that method has problems with spaces in the path and file names.  Using os.rename() got us around that, but os.rename() won't cross file system boundaries.  fileCopy uses io.open() which handles spaces in the path and file names without problems.  fileMove uses fileCopy, then removes the original if the copy succeeds.

boucman:
I am a bit worried about performance (copying huge files/lots of files) within the same filesystem will be way slower, and accidentally copying weird files (devices) might cause problems.

What exactly was the problem with cp ? in theory you can quote names the shell way to deal with spaces...

wpferguson:
The theory is correct.  I ran a test and the quoted names worked.  I did a benchmark to see what the throughput was.  The lua fileCopy had a throughput of ~970 MB/s.  The os.execute("cp..... was ~300x faster.  Knowing this, I'll recode fileMove to use os.execute("mv.. and delete fileCopy.

wpferguson:
I thought some more about my last comment, along the lines of a pure lua solution should work across different operating systems.  I also agree with boucman's concerns about performance.  I thought about coding fileMove using checkIfBinExists to find the mv function (which is at /bin/mv on my system also) and using it if it's installed, otherwise using the lua solution...  i.e.

function fileMove(from, to)
  if checkIfBinExists(mv) then
    os.execute("mv .....")
  else
    fileCopy.....
  end
end

We could also do a check on where the binary was found, i.e. checkIfSafeBinExists, and act accordingly.

boucman:
I don't know if looking for mv at a fixed place is the right answer either. That's not really our job, PATH is something we rely on, not something we check. IOW if the user wants to override mv, he has to be able to do it.

houz:
Does `os.move` indicate some error when not being successful due to file system boundaries? If yes then I would propose to try that first as it's probably the most portable way and only fall back to other means when it failed.

wpferguson:
os.rename returns a nil and an error message ("Invalid cross-device link" in the case of crossing file system boundaries).  

I propose that we:

1. use os.rename, and if it fails
2. use os.execute("cp...., and if it fails
3. use a pure lua solution, and if it fails
4. return an error

Does this work for everyone?

supertobi:
Sounds good.

houz:
Sounds good to me, too. And maybe put the whole logic into a function that can be used in other scripts, too. Same should be done with `checkIfBinExists`.

wpferguson:
I'll recode it this evening, test it, then push it.

A couple of weeks ago I mentioned to supertobi that we should look at pulling common code into a library.  He agreed.  checkIfBinExists is used in 8 scripts currently.  A couple of months ago I figured out how to do it and coded up a test that worked.  I'm currently working on enfuse's TODO list, which I'll push when we're done with this, then I'll start on the library.

contrib/de_DE/LC_MESSAGES/gimp.po
```diff
@@ -21,19 +21,23 @@ msgstr ""
...
-#: gimp.lua:189
+#: gimp.lua:194
 #, lua-format
 msgid "Export Image %i/%i"
 msgstr "Exportiere Bild %i/%i"
 
-#: gimp.lua:194
+#: gimp.lua:242
+msgid "Unable to move edited file into collection. Leaving it as %s"
+msgstr "Die bearbeitete Datei kann nicht in die Sammlung aufgenommen werden. Bearbeitet Datei ist %s"
```
supertobi: Better: "Die bearbeitete Datei kann nicht in die Sammlung aufgenommen werden. Bearbeitete Datei ist %s"

supertobi:
One more thing. Could you please make a separate pull request for the "geo_uri.lua" to keep the commit history clean?

PR #67, opened 2016-09-07
official/enfuse.lua TODO list

wpferguson:
Fixed the TODO list in official/enfuse.lua

Summary of  changes:

- Since enfuse can't process raw images, converted it to a storage so that raws can be converted and processed.  Named it Enfuse HDR
-  Added another storage to take advantage of enfuse's ability to process focus stack images and named it  Enfuse Focus Stack
- Fixed TODO - find a less stupid way to make sure the float value of exposure_mu gets turned into a string with a decimal point instead of a comma in some languages
- Fixed TODO - make the output filename unique so you can use it more than once per filmroll
- Fixed TODO - export images that are not ldr and remove them afterwards
- Removed TODO - save exposure-mu value when changed.  The value gets saved when the storage is executed.  It does not get saved every time the slider is moved.  The slider api does not support a changed_callback.  I'm not sure there  is any reason to save the slider value whenever it is changed and Enfuse HDR is not executed.
- Added German message translations
- Check for the multiprocessor version of enfuse and use it if installed.  On my system the mp version is 5 - 6  times faster (i7-6820HK CPU @ 2.70GHz)
- Added an option to align images using align_image_stack, part of the hugin project, if installed.

### DIFF
--- a/official/de_DE/LC_MESSAGES/enfuse.po
+++ b/official/de_DE/LC_MESSAGES/enfuse.po
@@ -29,7 +29,7 @@ msgstr "Exportiere Bild %i/%i"
 
 #: enfuse.lua:228
 msgid "enfuse-mp executable found, using it"
-msgstr "enfuse-mp ausführbare Datei gefunden, wird benutzen es"
+msgstr "enfuse-mp ausführbare Datei gefunden, wird verwenden es"
 
 #: enfuse.lua:254
 msgid "exposure mu"
@@ -37,11 +37,11 @@ msgstr "Belichtung mu"
 
 #: enfuse.lua:255
 msgid "center also known as MEAN of Gaussian weighting function (0 <= MEAN <= 1); default: 0.5"
-msgstr "Zentrum auch als Mittel der Gaußschen Gewichtungsfunktion ( 0 <= MEAN < = 1) bekannt ist; Standard: 0.5"
+msgstr "Zentrum auch als Mittel der Gaußschen Gewichtungsfunktion ( 0 <= MEAN < = 1) bekannt; Standard: 0.5"
 
 #: enfuse.lua:265 enfuse.lua:298
 msgid "depth"
-msgstr "Ausmaß"
+msgstr "Farbtiefe"
 
 #: enfuse.lua:266 enfuse.lua:299
 msgid "output image bits per channel"
@@ -49,7 +49,7 @@ msgstr "Ausgangsabbildbits pro Kanal"
 
 #: enfuse.lua:274 enfuse.lua:309
 msgid "Align Images"
-msgstr "Richten Sie Bilder"
+msgstr "Ausrichten der Bilder"
 
 #: enfuse.lua:319
 msgid "gray projector"
@@ -57,7 +57,7 @@ msgstr "grau Projektor"
 
 #: enfuse.lua:320
 msgid "type of grayscale conversion, default = average"
-msgstr "Art der Graustufen -Konvertierung, default = mittel"
+msgstr "Art der Graustufen-Konvertierung, Standard: mittel"
 
 #: enfuse.lua:329
 msgid "contrast window size"
@@ -65,7 +65,7 @@ msgstr "Kontrastfenstergröße"
 
 #: enfuse.lua:330
 msgid "size of box used for contrast detection, >= 3, default 5"
-msgstr "Größe des Kastens für Kontrasterkennung verwendet,> = 3, default 5"
+msgstr "Größe des Kastens, der für die Kontrasterkennung verwendet wird, > = 3, Standard: 5"
 
 #: enfuse.lua:338
 msgid "contrast edge scale"
@@ -73,11 +73,11 @@ msgstr "Kontrastkante Skala"
 
 #: enfuse.lua:339
 msgid "laplacian edge detection, 0 disables, >0 enables, .3 is a good starting value"
-msgstr "Laplace-Kantenerkennung , 0 deaktiviert die Funktion,> 0 ermöglicht, 0,3 ein guter Start"
+msgstr "Laplace-Kantenerkennung, 0 deaktiviert die Funktion, > 0 aktiviert die Funktion, 0,3 ist ein guter Start"
 
 #: enfuse.lua:377
 msgid "Aligning images..."
-msgstr "Ausrichten von Bildern..."
+msgstr "Ausrichten der Bildern..."
 
 #: enfuse.lua:379
 msgid "Image alignment failed"
@@ -97,11 +97,11 @@ msgstr "Starten von enfuse..."
 
 #: enfuse.lua:490 enfuse.lua:556
 msgid "enfuse failed, see terminal output for details"
-msgstr ""
+msgstr "enfuse fehlgeschlagen, siehe Terminal-Ausgabe für Details"
 
 #: enfuse.lua:501 enfuse.lua:567
 msgid "enfuse was successful, resulting image was imported"
-msgstr "enfuse fehlgeschlagen, siehe Terminal-Ausgabe für Details"
+msgstr "enfuse war erfolgreich, resultierendes Bild wurde importiert"
 
 #: enfuse.lua:576
 msgid "Enfuse HDR"
--- a/official/de_DE/LC_MESSAGES/enfuse.po
+++ b/official/de_DE/LC_MESSAGES/enfuse.po
@@ -8,106 +8,105 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2016-09-06 00:29-0400\n"
+"POT-Creation-Date: 2016-09-06 23:30-0400\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: Bill Ferguson <wpferguson@gmail.com>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
-"Language: de_DE\n"
 "MIME-Version: 1.0\n"
-"Content-Type: text/plain; charset=CHARSET\n"
+"Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
-X-Generator: Poedit 1.5.4\n"
+"X-Generator: Poedit 1.5.4\n"
 "Language: de_DE\n"
 "X-Poedit-SourceCharset: UTF-8\n"
 "X-Poedit-KeywordsList: gettext;dgettext:2;dcgettext:2ngettext:1,2;"
 "dngettext:2,3\n"
 "X-Poedit-Basepath: .\n"
 
-#: enfuse.lua:134
+#: enfuse.lua:217
 #, lua-format
 msgid "Export Image %i/%i"
 msgstr "Exportiere Bild %i/%i"
 
-#: enfuse.lua:142
+#: enfuse.lua:228
 msgid "enfuse-mp executable found, using it"
-msgstr "enfuse -mp ausführbare Datei gefunden, wird benutzen es"
+msgstr "enfuse-mp ausführbare Datei gefunden, wird benutzen es"
 
-#: enfuse.lua:162
+#: enfuse.lua:254
 msgid "exposure mu"
 msgstr "Belichtung mu"
 
-#: enfuse.lua:163
+#: enfuse.lua:255
 msgid "center also known as MEAN of Gaussian weighting function (0 <= MEAN <= 1); default: 0.5"
 msgstr "Zentrum auch als Mittel der Gaußschen Gewichtungsfunktion ( 0 <= MEAN < = 1) bekannt ist; Standard: 0.5"
 
-#: enfuse.lua:171 enfuse.lua:206
+#: enfuse.lua:265 enfuse.lua:298
 msgid "depth"
 msgstr "Ausmaß"
 
-#: enfuse.lua:172 enfuse.lua:207
+#: enfuse.lua:266 enfuse.lua:299
 msgid "output image bits per channel"
 msgstr "Ausgangsabbildbits pro Kanal"
 
-#: enfuse.lua:182 enfuse.lua:240
+#: enfuse.lua:274 enfuse.lua:309
 msgid "Align Images"
 msgstr "Richten Sie Bilder"
 
-#: enfuse.lua:213
+#: enfuse.lua:319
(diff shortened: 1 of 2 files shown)


official/enfuse.lua
```diff
@@ -16,26 +17,93 @@
...
+    enfuse - export the selected images and run enfuse in one of two modes
+    * Enfuse HDR - combine the exported images into a single HDR image
+    * Enfuse Focus Stack - Combine a set of focus stack images into a single image
+
+    USAGE
+    * require this script from your main lua file
+    * two new storage options are created, Enfuse HDR and Enfuse Focus Stack
+    * Enfuse HDR
+      * select the bracketed images to use for the hdr
+      * adjust exposure mu to change the brightness of the output image
```
boucman: mu ?

wpferguson: I guess I should have written it exposure-mu.  It is an option to enfuse that determines the center of the Gaussian exposure weight curve.

boucman:
boy, we really need a "common" lib, or a way to deal with dependencies (luarock, maybe ?)

i'll do a complete proofread ASAP, but that looks good.

@houz , since you're the original author, could you have a look too ?

wpferguson:
The library is the next project... :)

official/de_DE/LC_MESSAGES/enfuse.po
```diff
@@ -0,0 +1,112 @@
...
+"X-Poedit-Basepath: .\n"
+
+#: enfuse.lua:217
+#, lua-format
+msgid "Export Image %i/%i"
+msgstr "Exportiere Bild %i/%i"
+
+#: enfuse.lua:228
+msgid "enfuse-mp executable found, using it"
+msgstr "enfuse-mp ausführbare Datei gefunden, wird benutzen es"
```
supertobi: msgstr "enfuse -mp ausführbare Datei gefunden, wird verwenden sie"

official/de_DE/LC_MESSAGES/enfuse.po
```diff
@@ -0,0 +1,112 @@
...
+
+#: enfuse.lua:254
+msgid "exposure mu"
+msgstr "Belichtung mu"
+
+#: enfuse.lua:255
+msgid "center also known as MEAN of Gaussian weighting function (0 <= MEAN <= 1); default: 0.5"
+msgstr "Zentrum auch als Mittel der Gaußschen Gewichtungsfunktion ( 0 <= MEAN < = 1) bekannt ist; Standard: 0.5"
```
supertobi: msgstr "Zentrum auch als Mittel der Gaußschen Gewichtungsfunktion ( 0 <= MEAN < = 1) bekannt; Standard: 0.5"

official/de_DE/LC_MESSAGES/enfuse.po
```diff
@@ -0,0 +1,112 @@
...
+
+#: enfuse.lua:265 enfuse.lua:298
+msgid "depth"
+msgstr "Ausmaß"
```
supertobi: msgstr "Farbtiefe"

official/de_DE/LC_MESSAGES/enfuse.po
```diff
@@ -0,0 +1,112 @@
...
+
+#: enfuse.lua:266 enfuse.lua:299
+msgid "output image bits per channel"
+msgstr "Ausgangsabbildbits pro Kanal"
+
+#: enfuse.lua:274 enfuse.lua:309
+msgid "Align Images"
+msgstr "Richten Sie Bilder"
```
supertobi: msgstr "Ausrichten der Bilder"

official/de_DE/LC_MESSAGES/enfuse.po
```diff
@@ -0,0 +1,112 @@
...
+
+#: enfuse.lua:319
+msgid "gray projector"
+msgstr "grau Projektor"
+
+#: enfuse.lua:320
+msgid "type of grayscale conversion, default = average"
+msgstr "Art der Graustufen -Konvertierung, default = mittel"
```
supertobi: msgstr "Art der Graustufen-Konvertierung, Standard: mittel"

official/de_DE/LC_MESSAGES/enfuse.po
```diff
@@ -0,0 +1,112 @@
...
+
+#: enfuse.lua:329
+msgid "contrast window size"
+msgstr "Kontrastfenstergröße"
+
+#: enfuse.lua:330
+msgid "size of box used for contrast detection, >= 3, default 5"
+msgstr "Größe des Kastens für Kontrasterkennung verwendet,> = 3, default 5"
```
supertobi: msgstr "Größe des Kastens, der für die Kontrasterkennung verwendet wird, > = 3, Standard: 5"

official/de_DE/LC_MESSAGES/enfuse.po
```diff
@@ -0,0 +1,112 @@
...
+
+#: enfuse.lua:338
+msgid "contrast edge scale"
+msgstr "Kontrastkante Skala"
+
+#: enfuse.lua:339
+msgid "laplacian edge detection, 0 disables, >0 enables, .3 is a good starting value"
+msgstr "Laplace-Kantenerkennung , 0 deaktiviert die Funktion,> 0 ermöglicht, 0,3 ein guter Start"
```
supertobi: msgstr "Laplace-Kantenerkennung, 0 deaktiviert die Funktion, > 0 aktiviert die Funktion, 0,3 ist ein guter Start"

official/de_DE/LC_MESSAGES/enfuse.po
```diff
@@ -0,0 +1,112 @@
...
+
+#: enfuse.lua:377
+msgid "Aligning images..."
+msgstr "Ausrichten von Bildern..."
```
supertobi: msgstr "Ausrichten der Bildern..."

official/de_DE/LC_MESSAGES/enfuse.po
```diff
@@ -0,0 +1,112 @@
...
+msgid "not enough suitable images selected, nothing to do for enfuse"
+msgstr "nicht genug geeignete Bilder ausgewählt, nichts für enfuse zu tun"
+
+#: enfuse.lua:488 enfuse.lua:554
+msgid "Launching enfuse..."
+msgstr "Starten von enfuse..."
+
+#: enfuse.lua:490 enfuse.lua:556
+msgid "enfuse failed, see terminal output for details"
+msgstr ""
```
supertobi: msgstr "enfuse fehlgeschlagen, siehe Terminal-Ausgabe für Details"

official/de_DE/LC_MESSAGES/enfuse.po
```diff
@@ -0,0 +1,112 @@
...
+
+#: enfuse.lua:501 enfuse.lua:567
+msgid "enfuse was successful, resulting image was imported"
+msgstr "enfuse fehlgeschlagen, siehe Terminal-Ausgabe für Details"
```
supertobi: msgstr "enfuse war erfolgreich, resultierendes Bild wurde importiert"

wpferguson: the last 2 were when copy an paste goes wrong....

houz:
I am not too happy about moving this into export. First, it's harder to find that way. Second, it will interfere with you export settings as you have to change them to enable this. Third, conceptually it doesn't belong there. I understand that it's easy to write the code like that, but from a user's point of view it's not an export but a processing of images.
So as an alternative I would suggest keeping it in its own lib and calling export manually for all images. That way every file type is supported while keeping it stand alone.

official/enfuse.lua
boucman: ok. I thought it was a typo, yes the dash helps...

wpferguson:
I looked at this way,  enfuse can't do anything with a raw, so the images have to be exported somewhere.  If I tie it into the exporter, then I gain the advantages of the exporter.  I can export lower resolution 8-bit jpgs and process them while playing with the enfuse settings, gaining a large increase in speed.  Once I have the settings I want, then I can export 16 bit tifs at full resolution and get a cup of coffee while I wait for it to process (okay, it's not quite that slow but it can take awhile). 

I'm trying to understand the conceptual statement.  If I understand it correctly, then any process that takes an image or images, processes them, then produces an image intended for further processing in darktable  should be a module on it's own?  And the exporter should only be used to generate end products, that are for distribution and not editing?  Conceptually, that makes sense.  However, I have a couple of concerns.  First, the exporter provides a lot of control over the "intermediate" image that we give to the process, allowing us to adjust for quality or speed plus applying the processing history.  Second, we're going to run out of screen real estate.  Some of the items in the exporter could be recoded  as shortcuts (GIMP and Hugin), but anything with settings is going to take space in the UI.  The exporter allows a lot of different UI's in a single section of the UI.

Do we need something that's not an exporter, but provides that functionality as an intermediate interface to external functions?

wpferguson:
I've been thinking about this over the last few hours and have one more thought about the exporter.  I'm guessing it started life as a way to get developed images out of darktable.  Over time we've learned how to send data out, process it and return it, so now, conceptually, exporter is really an interface.  It interfaces with the file system, flickr, google+, latex, facebook, hugin, GIMP, google earth, and more.  I think I can probably interface the Google Nik Collection without too much effort.  The exporter has evolved over time into something it was never imagined to be.  Maybe we should stop calling it the exporter and pick a name that better reflects the way it's being used.

boucman:
the idea of an image coming out and back in DT is totally new. I'll call them side-processors.

we don't handle SP well in DT (or at all) and those break a lot of stuff. including the fact that SP loose some of the non-destructive aspects of DT. You are not able to reproduce the processed image from its RAW file.

SP don't fit well as exporters, but they don't fit well anywhere at this point. Are they tools that produce new RAW files ? should we remove their source image ? keep a duplicate of the source image in the SPed image to be able to reproduce them ? should we provide some sort of generic IOP to do side processors rather than deal with them in the light table ?

iop have the problem that most side processors use multiple images as source. gimp add a SP that can't be automated (the user acts on GIMP)

This needs some thought, so we provide a generic answer. maybe a dedicated lib, but we must first conceptualize what a SP is and how it is supposed to work

wpferguson:
I think we've created plug-ins.  Oops.... ;-)

If we look at Enfuse HDR, Enfuse Focus Stack, and Hugin, they all take a series of images and produce a new image.  In essence they are "cameras" tethered to DT producing new images.  If we have our imaginary plug-in architecture,  then when we run the plug-in it can make an entry in the sidecar file about the plug-in and settings used to create the file.  And maybe we put an "information" entry into the source sidecar files.  I don't really know enough about sidecar files, yet, to figure out the mechanics, but it's an idea.  Once the image is imported, it's really like any other image from a camera, with the exception that we can recreate it.

GIMP is another issue.  In my workflow I import the raws into DT, adjust white balance, exposure, etc. until I get the image I want.  If I'm doing portraits then sometimes I need to remove blemishes, so I export to GIMP, do a frequency separation or wavelet decompose and clean up the skin.  I overwrite the file to get it back to DT, then cringe when I discard the xcf data.  For me, GIMP is an intermediate export where I make a final change or two.  After that I sharpen it, export it etc.  It's quick, it's easy, it works, and it's ugly.  If we have our imaginary plug-in architecture and a plug-in aware lighttable, then we could host an xcf file.  The plug-in could generate the thumbnail.  The exif data could be copied from the original image.  Maybe the plug-in could put the layer stack in the sidecar file so that the starting image data could be changed by selecting a layer and having the plug-in regenerate the image data.  If you opened the image in darkroom then the plug-in could generate the base image data and the pipeline could run on top of that.  In addition you could spawn GIMP directly from the lighttable and edit the xcf file, then the plug-in could update the thumbnail, sidecar layer stack, and image data on save.  Or, we could create a plug-in-data directory next to the darktable exported directory and store the xcf there and keep it out of the lighttable but accessible.  These are just ideas (wishes) of where we could go with this.

boucman:
FYI I just added the possibility to add any widget to the preference window via lua.

hope you find this usefull...

wpferguson:
Thanks @boucman.  I've pulled it, built it, and am looking at it...

I figured since I "broke" things with the plugin paradigm, I should try and "fix" it.  I've been adding a plugin architecture to the lua-scripts directory.  I took clear_GPS and "converted" it to a plugin, since it was trivial.  I'm working on hugin now, then I'll do enfuse and finally GIMP.  Once I have all of that working I'll do a pull request and we can start discussing the good, the bad, and the ugly.

While working on plugins, I'm also working on the library.  I've built it and am using it while working with the plugins, which is helping drive out what should be in there.  I'm also tweaking script_manager, making it self aware as well as plugin aware and working on the interface.

boucman:
you should have a look at luarocks too. it's the lua equivalent of pypy or cpan.

people will want to have some support for easily adding features, and scripts will want to use that for dependencies. some sort of support (standard way for scripts to ask for a particular module to be installed, standard way of installing, i'm not sure) would be usefull at some point. maybe not at the first go, though...

boucman:
in particular something like http://keplerproject.github.io/luafilesystem/index.html is a classic...

wpferguson:
I looked at luarocks when I was looking at the other open pull request.  Once I read through it I thought that it would be a good plugin installer.  If we used something like that, then we could create a repository where plugins could be registered, pull an updated list into a plugin manager, install the selected plugins with luarocks, then activate them with the plugin manager.

boucman:
I don't know if adding dt specific plugins to luarocks make sense... luaroks is about generic lua modules not about application specific stuff...

However luarocks extends greatly the capacities of lua, so our lua modules will probably want to pull stuff from luarocks, which is more in the line of what I htink would be really usefull

wpferguson:
I was thinking we should fork it, call it dtrocks, and set up a rock server with plugins.  :D

I have plugins working.  I ported clear_GPS, enfuse, and gimp to the plugin format.  I also added Color Efex Pro 4 from the Nik Collection.  I still need to do a lot of code cleanup before I do a PR, but I'm making progress.  I'll attach a screenshot showing what I have now.  You'll notice a really garish tif file which was edited with Color Efex Pro 4, which might be an argument for why we shouldn't include it. :)
dtplugin screenshot

wpferguson:
And another screenshot showing Enfuse Focus Stack selected with the widgets that go with it
dtplugin screenshot2

supertobi:
Show us the code. ;)

wpferguson:
@boucman, @supertobi should I close this?  It's been turned into a plugin so this is OBE (overtaken by events) and should probably go away.

supertobi:
It's your PR. Feel free to close it.

PR #72, opened 2016-10-05
first version of a quicktag modul

chrik5:
Hi,

Sorry I know that this is not ready for the public, but i need some advice on creating UI elements within a modul dynamically. The idea is to have a button and an entry field side by side using new_widget("box"). 

I hope it is okay to ask via pull request. 

Christian


wpferguson:
You could do

local widget_table = {}

for i=1,qnr do
  widget_table[#widget_table + 1] = button[i]
  widget_table[#widget_table  + 1] = tagentries[i]
end

then in your box widget do

  orientation = "vertical",
  table.unpack(widget_table),

Another suggestion is do away with the entry and make the button label  

"Apply " .. tostring(tagentries[j].text

Hope this helps

chrik5:
Yes that helps!

I have updated the script and I think it is usable now.

Christian

spaceChRiS:
It works very well on 2.0.6, so API version 3,0,0 could be enabled.

Should we discuss possible enhancements which might require changes in the lua implementation here or better leave it as is for a while (it is a great module as is, I will use it a lot :-))? I think on features such as enabling automatic tag completion for text entry fields and access to the preset system (or a parallel implementation with an in dependent storage) to enable storage of tag lists.

chrik5:
@spaceChRiS: Thanks for testing the script!

Well, generell darktable LUA discussions should be continued on the mailing list.

1. I do not want to create a parallel taglist. I've already messup my tags a bit. That's why new tags are not added by default. Of course I could add an option to enable tag creation. I'm not sure if it makes sense or improves the tagging speed. With CTRL+t you can add tags quickly enough.

2. A search while write functionality will be nice, but for now I do not know howto implement it.

spaceChRiS:
I used the module during the first days of my holiday and it's a big time saver to tag my family members on the photos. Still some days to go for the holiday stress test :-).

To your comments:

1. I did not refer to a parallel tag list, but a storage backend for entire module settings. E.g., at the moment I am tagging my family members, if I want to tag e.g. the bridal couple of a wedding of another film roll, I have to change the number of tags in the module, the tags themselves and maybe the keys used for tagging. If these lists would be stored as presets, I could easily switch between different configurations.
1. To implement it in lua, something like an “input changed” signal or callback would be necessary, I don't think it is already there. Maybe better would be a direct access to the tag autocompletion feature of darktable.

Some additional findings (some of them were already discussed, I just repeat them to have everything in one place):

- Setting shortcuts doesn't warn before overwriting some already there, even default ones. I guess this is a darktable “bug”, not directly reloaded to the module.
- It would be great if the apply button would show the shortcut.
- A better layout would be to have buttons to the right of tags, but this may require some changes in lua gui implementation.
- Setting shortcuts from the module would be great.
- Temporary activation/overwriting of global dt shortcuts would be great as well, to not mess with the shortcuts defined in preferences.

chrik5:
> Temporary activation/overwriting of global dt shortcuts would be great as well, to not mess with the shortcuts defined in preferences.

Just seen at [pixls.us](https://discuss.pixls.us):
Using a [gamepad](https://discuss.pixls.us/t/midi-controller-for-darktable/2582/12?u=pk5dark)  and map the gamepad keys to different shortcuts with [AntiMicro](https://github.com/AntiMicro/antimicro) opens new possibilities. 
Currently I use F1 to F5 for my tags and together with the copy attach tag script this allows really fast tagging. Time to get such a gamepad ;-)!

chrik5:
> A better layout would be to have buttons to the right of tags, but this may require some changes in lua gui implementation.

Done, but still not good? At least now I know how-to update button labels.

spaceChRiS:
First, thanks for continuing the development of the module, it is used a lot by me and really solves workflow problems. A proper UI for this module sounds so simple but seems very complicated. The new version doubles the information by having a button and a text entry with the same info. This makes the sidebar too wide. I would go with a simple apply button instead, maybe with an arrow or the typical “done” icon instead of the word apply to save space. Another idea would be to have the text entry separated from the buttons, e.g. by having only one text entry at the bottom and a button in each row which transfers the new tag to the row's button. An option to abbreviate the tag's path like “p..e/f..s/John” or “p/f/John” instead of “people/friends/John” would be handy as well.

Please tell me if I am getting on your nerves by posting ideas, unfortunately I always have plenty of them and so little time to work on these by myself :(.

spaceChRiS:
I tested a bit more and am now confident that the button should show the shortcut, or, if not assigned, the word “apply”. Unfortunately, I was only able to have a 50/50 space distribution between button and text entry in each row, any idea how to solve it? And I did not figure out how to access the shortcut. Any idea on that?

chrik5:
Here is a new attempt. Entry fields are gone and it looks much cleaner now.

 For the shortcuts:
```
$ egrep quick ~/.config/darktable/keyboardrc
(gtk_accel_path "<Darktable>/lua/quickt tag 1" "<Alt>y")
(gtk_accel_path "<Darktable>/lua/quick tag 3" "<Alt>t")
(gtk_accel_path "<Darktable>/lua/quick tag 4" "F5")
(gtk_accel_path "<Darktable>/lua/quick tag 2" "<Alt>r")
(gtk_accel_path "<Darktable>/lua/quick tag 1" "<Alt>y")
```

So, to have them as button labels one needs to "parse" this file into the label table ;-). Or even better as tool tip.

spaceChRiS:
While using an external command line tool to access this information could work, a saner way would be to make it available from the lua api. I guess we have 360 days minus some days of feature freeze period to solve the issue ;). (And to solve button text alignment, too ;).)

chrik5:
I think it works so far. For me it is a nice enhancement.

contrib/quicktag.lua
```diff
@@ -0,0 +1,221 @@
...
+
+USAGE
+   * set the number of quicktags between 2 and 10 in the preferences dialog and restart darktable
+   * if wanted set the shortcuts in the preferences dialog
+   * to add or change a quicktag, first select the old tag with the combobox "old quicktag",
+     enter a new tag in the "new quicktag" filed and press "set quicktag"
+   * use a shortcut or button to attach the tag to selected images
+
+TODO
+    * abbrevate button labels
```
supertobi: What about this ToDo, do you plan to do something here?

contrib/quicktag.lua
```diff
@@ -0,0 +1,221 @@
...
+   dt.preferences.register("quickTag",
+                        "quicktag"..j,
+                        "string",                    -- type
+                        "quick tag "..j,           -- label
+                        "please create the tags before in the tagging module",   -- tooltip
+                        ""                            -- default
+                        )
+  end
+]]
+
```
supertobi: If this isn't used you could clean it.

contrib/quicktag.lua
```diff
@@ -0,0 +1,221 @@
...
+      dt.print("quicktag \""..old_tag.."\" replaced by \""..new_quicktag.text.."\"")
+      update_quicktag_list()
+      new_quicktag.text = ""
+    end
+  end,
+}
+
+local new_qt_widget = dt.new_widget ("box") {
+    orientation = "horizontal",
+    dt.new_widget("label") { label = "new quicktag" },
```
supertobi: It would be nice, if you could make your script translatable.

supertobi:
And please remove the copy_attach_detach_tags.mo, we don't add binaries.

contrib/quicktag.lua
chrik5: This point was raised by spaceChRiS. For me the side panel is wide enough all tags I have. Tags can be really, really long. I've seen that there are string functions in lib/dtutils. 

I want to wait for complains and add this maybe later.

chrik5: So now the last point from the TODO list is done.

chrik5:
@spaceChRiS Can you test the script please?

contrib/quicktag.lua
```diff
@@ -0,0 +1,220 @@
...
+    quicktag_table[j] = dt.preferences.read("quickTag", "quicktag"..j, "string")
+    end
+  end
+
+
+
+-- quicktag function to attach tags
+local function tagattach(tag,qtagnr)
+  if tag == "" then
+    dt.print(_("quicktag").." "..qtagnr.." ".. _("empty, please set a tag"))
```
supertobi: For strings, that needs translations it is better to use string.format(), because in some languages the word order is different.
You can find here an example:
https://github.com/darktable-org/lua-scripts/blob/master/contrib/kml_export.lua#L77

contrib/quicktag.lua
```diff
@@ -0,0 +1,220 @@
...
+    return true
+  end
+
+  local counter = 0
+
+  for _,image in ipairs(sel_images) do
+     dt.tags.attach(tagnr,image)
+     counter = counter+1
+  end
+  dt.print(_("tag").."  \""..tag.."\" ".._("attached to").." "..counter.." ".._("image(s)"))
```
supertobi: the same then above

contrib/quicktag.lua
```diff
@@ -0,0 +1,220 @@
...
+local set_quicktag_button = dt.new_widget("button") {
+  label = _("set quicktag"),
+  clicked_callback = function()
+    local old_tag = quicktag_table[old_quicktag.selected]
+    if new_quicktag.text == "" then
+      dt.print(_("new quicktag is empty!"))
+    else
+      quicktag_table[old_quicktag.selected] = new_quicktag.text
+      dt.preferences.write("quickTag", "quicktag"..old_quicktag.selected, "string",  new_quicktag.text)
+      dt.print(_("quicktag") .." \""..old_tag.."\" ".._("replaced by").." \""..new_quicktag.text.."\"")
```
supertobi: the same then above

contrib/quicktag.lua
```diff
@@ -0,0 +1,220 @@
...
+  },
+  nil,-- view_enter
+  nil -- view_leave
+)
+
+-- create shortcuts
+for i=1,qnr do
+  dt.register_event("shortcut", 
+		   function(event, shortcut) tagattach(tostring(quicktag_table[i])) end,
+		  _("quicktag").." "..i)
```
supertobi: the same then above

contrib/quicktag.lua
supertobi: I'm OK with that

supertobi:
@chrik5 And now you need to fix the .po file. ;)

chrik5:
@supertobi Just pushed a few things in between, because I switched computers ;-) Thank you very much for your patience and support! I think it is time to GO.

contrib/de_DE/LC_MESSAGES/quicktag.po
```diff
@@ -6,82 +6,79 @@ msgid ""
...
 "Project-Id-Version: \n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2017-02-20 12:05+0100\n"
-"PO-Revision-Date: 2017-02-20 12:09+0100\n"
+"POT-Creation-Date: 2017-02-20 17:38+0100\n"
+"PO-Revision-Date: 2017-02-20 17:50+0100\n"
 "Last-Translator: Christian Kanzian <christian.kanzian@gmx.at>\n"
-"Language-Team: German <>\n"
-"Language: de_AT\n"
+"Language-Team: German <kde-i18n-de@kde.org>\n"
```
supertobi: kde-i18n-de@kde.org

???

spaceChRiS:
Hi @chrik5, that's a great script getting better with every iteration. For the abbreviation code, I would rather do something such as
```
local function abbrevate_tags(t)
  for i=1,qnr do
    local taglength = string.len(t[i])
    if taglength > max_label_length then

      s = {}; k = 0
      for str in string.gmatch(t[i], "([^|]+)") do
        s[k] = str
        k = k + 1
      end

      quicktag_label[i] = ""
      k = 0
      while s[k+1] do
        quicktag_label[i] = quicktag_label[i] .. string.sub(s[k],0,1) .. ".." .. string.sub(s[k],-1) .. "|"
        k = k + 1
      end
      quicktag_label[i] = quicktag_label[i] .. s[k]

    else
      quicktag_label[i] = t[i]
    end
  end
end
```

chrik5:
Hi @spaceChRiS, sounds nice -please create a PR so that @supertobi can review it. Any improvements are warmly welcome.

PR #73, opened 2016-10-06
Open nautilus on selected images' paths.

spaceChRiS:
Darktable script to open nautilus on hovered or selected image's paths. Hover has precedence if hovered image not part of selection. If only one image of every path is marked, the image will be selected in nautilus.

Maybe this is useful for somebody else.

There are still some open questions, especially how an external gui program can be handed over the window focus from darktable lua. Any ideas?

### DIFF
--- a/contrib/openInNautilus.lua
+++ b/contrib/openInNautilus.lua
@@ -1,0 +1,57 @@
+--[[
+    This file is part of darktable,
+    Copyright 2016 by Christian Mandel
+
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 3 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+]]
+--[[
+Darktable script to open nautilus on hovered or selected image's paths. Hover
+has precedence if hovered image not part of selection. If only one image of
+every path is marked, the image will be selected in nautilus.
+
+USAGE
+  * require this script from your main lua file
+]]
+   
+local dt = require "darktable"
+dt.configuration.check_version(...,{3,0,0})
+
+local function filemanager_shortcut(event, shortcut)
+    local images = dt.gui.action_images
+    local image_paths = {}
+    local multiple_images = 0
+    for _,v in pairs(images) do
+        if image_paths[v.path] then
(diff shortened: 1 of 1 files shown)


chrik5:
What about xdg-open instead of nautilus?
This will open the default filemanager and work for most users I guess.

spaceChRiS:
Unfortunately xdg-open does not allow to mark files by adding the file name to the path (at least I was not able to make it work). Maybe it would be better to get the default file manager from the desktop directly and make special cases for the different file managers, making maximal use of their capabilities (see also my TODO list at the end of the file).

chrik5:
>  (see also my TODO list at the end of the file).

Ups, sorry I have not seen it. 

>  * check for nautilus' existance

have a look at https://github.com/darktable-org/lua-scripts/blob/master/contrib/hugin.lua#L49

spaceChRiS:
Thanks for the hint, I wanted to steal it from https://github.com/darktable-org/lua-scripts/blob/master/contrib/slideshowMusic.lua#L38 later (now I will compare and select the better one :-)), but I knew that I will not have time to finish it today, therefore I started the PR without to ask for the window focus problem. Any hint for that?

wpferguson:
Window focus is controlled by the window manager (X11).  Depending on your focus settings (focus follows mouse, click to focus) the new window may or may not get focus.  However, if you are on an Ubuntu or Debian system there is a program /usr/bin/wmctrl which allows you to manipulate windows and focus.

Hope this helps,

Bill

spaceChRiS:
@wpferguson, it helps as it shows an opportunity to solve the issue if there is no possibility from the GTK side. However, I wonder how other software deals with this issue, e.g., the DownThemAll plugin for firefox can open nautilus in the foreground.

supertobi:
Perhaps have a look into the file manager DBus interface:
https://www.freedesktop.org/wiki/Specifications/file-manager-interface/

You can send DBus messages with "dbus-send" in the command line:
https://dbus.freedesktop.org/doc/dbus-send.1.html

boucman:
dbus-send has been superseeded by "busctl" also, i'm pretty sure there is a lua lib that provides a dbus API somewhere in luarocks... maybe you could depend on that (just an idea, not something i'm certain about)

supertobi:
@boucman 
Yes, there is a lua lib that provides a dbus API:
https://luarocks.org/modules/daurnimator/ldbus

@spaceChRiS 
What is your status here? Do you need more help?
I would suggest, that you add the check for nautilus existence fist. Then we can get this in. After that you can work on the other ToDos.

spaceChRiS:
I'm on vacation, alone with my 2.8 years old son, at the moment (this week), so not much time to work on this. But I'll try to finish it asap.

supertobi:
Any news here?

spaceChRiS:
Unfortunately not. New job, less time.

supertobi:

supertobi:
As we now have OpenInExplorer.lua with similar functionality I'll close this PR.

PR #74, opened 2016-10-15
Libraries

wpferguson:
Moved commonly used functions into lib/dtutils.lua.  Split out the file functions into dtutils.file sub library.  Moved official/debug-helpers.lua to dtutils.debug sub library.  Added dtutils.extensions sub library for extensions to the lua libraries and dtutils.processor for external processor specific functions.  Documented all functions and libraries.  Created embedded documentation in each library file.  Created tools to pull the embedded documentation in text, man, and pdf formats.


wpferguson:
Here's the generated man page and pdf documentation
[doc.zip](https://github.com/darktable-org/lua-scripts/files/530789/doc.zip)

lib/dtutils/file.lua
```diff
@@ -0,0 +1,523 @@
...
+local dt = require "darktable"
+
+local log = require "lib/libLog"
+
+local gettext = dt.gettext
+
+dt.configuration.check_version(...,{3,0,0})
+
+-- Tell gettext where to find the .mo file translating messages for a particular domain
+gettext.bindtextdomain("dtutils.file",dt.configuration.config_dir.."/lua/locale/")
```
boucman: again, I don't really like that way of doing things... that place won't be exactly where the lua-scripts was cloned, so chances are high that the user has to move files out of the clone to the proper location, and do it every time he updates the scripts

use the "dump" functionnality of dt.debug to see exactly what's in the ... argument provided to your script and find the proper directory that way... that will be way more future-proof and also a great example for futur users of this library

lib/dtutils/processor.lua
```diff
@@ -0,0 +1,266 @@
...
+      format - dt_imageio_module_format_t - the format of the export image
+      filename - string - the filename being exported to
+      number - integer - the number of this image in the sequence
+      total - integer - the total number of images being exported
+      high_quality - boolean
+      extra_data - extra data from the storage
+
+  DESCRIPTION
+    exporter_status runs prior to each image being exported and prints out a status that lets the user
+    know how many images, out of the total, have been exported
```
boucman: "should be run" (unless it's an internal function to dtutils.processor)

also, the important part is that it's printing the result in a common way. you might want to show what it prints in your example so people understand better what it does

boucman: adding an example would also help show how to use it correctly (the high number of unused parameters make me think it's meant to be used with the "..." parameters of a processing callback. You need to say this explicitely

lib/dtutils/processor.lua
```diff
@@ -0,0 +1,266 @@
...
+
+]]
+
+dtutils_processor.libdoc.functions[#dtutils_processor.libdoc.functions + 1] = {
+  Sections = {"Name", "Synopsis", "Description", "Return_Value"},
+  Name = [[make_output_filename- make an output filename from an image list]],
+  Synopsis = [[local dp = require "lib/dtutils.processor"
+    
+    local result = dtutils.make_output_filename(img_list)
+      img_list - string - a space separated list of filenames]],
```
boucman: wouldn't an image table be more consistant with the other functions in that block, rather that a space-separated list of filenames ?

lib/dtutils.lua
```diff
@@ -0,0 +1,457 @@
...
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+]]
+
+local dtutils = {}
+
+dtutils.libdoc = {
+  Sections = {"Name", "Synopsis", "Description", "See_Also", "License"},
```
wpferguson: I modeled the documentation after man pages.  At the same time I wanted to have the flexibility to add or remove sections, so it's not a "strict" man page format.  The only way I could figure out to get the sections to come out in order was using the array.  If I just used keys then the order isn't guaranteed.  It can be error prone.  I got caught once with a section and no key, which is why there's a check in the tools to make sure section doesn't reference a non-existent key.

Adding the copyright section is simply a matter of adding it to the array, then adding the key and value.  I'll do that.

lib/dtutils.lua
```diff
@@ -0,0 +1,457 @@
...
+  Name = [[dtutils - A Darktable lua utilities library]],
```
wpferguson: I've read the next couple of comments also, so I'll try and address them all here.  I modeled this format pretty closely after man pages, including keeping their terminology.  How about we make Name just the name, and change Synopsis to Usage.  Do we want a Synopsis, or do we just depend on the Description?

lib/dtutils.lua
```diff
@@ -0,0 +1,457 @@
...
+    occurrence of the supplied pattern.
+
+  RETURN VALUE
+    result - a table of strings on success, or an empty table on error
+
+  EXAMPLE
+    split("/a/long/path/name/to/a/file.txt", "/") would return a table like
+      {"a", "long", "path", "name", "to", "a", "file.txt"}
+
+]]
```
wpferguson: I kept them both in case you didn't like the libdoc. ;)

lib/dtutils.lua
```diff
@@ -0,0 +1,457 @@
...
+dtutils.libdoc.functions[#dtutils.libdoc.functions + 1] = {
+  Sections = {"Name", "Synopsis", "Description", "Return_Value", "Example"},
+  Name = [[split - split a string on a specified separator]],
+  Synopsis = [[local du = require "lib/dtutils"
+
+    local result = du.split(str, pat)
+      str - string - the string to split
+      pat - string - the pattern to split on]],
+  Description = [[split separates a string into a table of strings.  The strings are separated at each
+    occurrence of the supplied pattern.]],
```
wpferguson: good idea

lib/dtutils.lua
```diff
@@ -0,0 +1,457 @@
...
+
+    local result = du.prequire(req_name)
+      req_name - the filename of the lua code to load without the ".lua" filetype
+
+  DESCRIPTION
+    prequire is a protected require that can survive an error in the code being loaded without
+    bringing down the calling routine.
+
+  RETURN VALUE
+    result - the code or true on success, otherwise an error message
```
wpferguson: That's fine.  I use this in script_manager and plugin_manager so that a "bad" script doesn't bring everything crashing down.

lib/dtutils.lua
```diff
@@ -0,0 +1,457 @@
...
+  Synopsis = [[local du = require "lib/dtutils"
+
+    du.push(stack, value)
+      stack - table - a table being used as the stack
+      value - any type - the value to be put on the stack]],
+  Description = [[Push a value on a stack]],
+  Return_Value = [[none]],
+}
+
+dtutils.push = table.insert
```
wpferguson: gone

lib/dtutils.lua
```diff
@@ -0,0 +1,457 @@
...
+  Name = [[pop - pop a value from a stack]],
+  Synopsis = [[local du = require "lib/dtutils"
+
+    local result = du.pop(stack)
+      stack - table - a table being used as a stack]],
+  Description = [[Remove the last value pushed on the stack and return it]],
+  Return_Value = [[result = nil if stack isn't a table or the stack is empty, otherwise the value removed from the stack]],
+}
+
+dtutils.pop = table.remove
```
wpferguson: gone

lib/dtutils/debug.lua
```diff
@@ -0,0 +1,219 @@
...
+You should have received a copy of the GNU General Public License
+along with darktable.  If not, see <http://www.gnu.org/licenses/>.
+]]
+--[[
+DEBUG HELPERS
+A collection of helper functions to help debugging lua scripts.
+
+require it as
+
+dhelpers = require "official/debug-helpers"
```
wpferguson: Oops ;)

lib/dtutils/debug.lua
```diff
@@ -0,0 +1,219 @@
...
+      ... - arguments - variables to dump the contents of]],
+  Description = [[tracepoint prints its name and dumps its parameters using
+    dt.debug]],
+  Return_Value = [[result - ... - the supplied argument list]],
+}
+
+function M.tracepoint(name,...)
+  log.always(4, "*****  "..name.." ****")
+  params = {...}
+  print(dt.debug.dump(params,"parameters"))
```
wpferguson: I'll look at it and see what I can do

lib/dtutils/extensions.lua
```diff
@@ -0,0 +1,185 @@
...
+
+]]
+
+local dtutils_extensions = {}
+dtutils_extensions.libdoc = {
+  Sections = {"Name", "Synopsis", "Description", "License"},
+  Name = "dtutils.extensions - a library of extenstions to the lua libraries",
+  Synopsis = [[require "lib/dtutils.extensions"]],
+  Description = [[This library contains extensions to the lua libraries.]],
+  License = [[This program is free software: you can redistribute it and/or modify
```
wpferguson: Agreed.  Do we want the "This is part of darktable ..." lines included?

lib/dtutils/extensions.lua
```diff
@@ -0,0 +1,185 @@
...
+    result = string.strip_accents(str)
+      str - string - the string with characters that need accents removed]],
+  Description = [[strip_accents removes accents from accented characters returning the 
+    unaccented character.]],
+  Return_Value = [[result - string - the string containing unaccented characters]],
+}
+
+-- Strip accents from a string
+-- Copied from https://forums.coronalabs.com/topic/43048-remove-special-characters-from-string/
+function string.strip_accents( str )
```
wpferguson: I just copied these out of the scripts that I found them in because I thought they could/would be useful.  I didn't really apply a lot of thought about how they fit into the grand scheme of things.  That's why I like having others look over what I did and saying "Did you really think this out?" because sometimes I don't ;).  I kind of struggled with what to do with these routines anyway and dtutils.extensions was the best I could come up with.  dtutils.string makes more sense and gives us a home for future string manipulation routines.

lib/dtutils/file.lua
wpferguson: Maybe I don't understand gettext correctly.  My understanding of this line is that it tells where the .mo files, that are generated from the .po files in the script directories, are, not the path to the .po files.  Does darktable read the .po files and use those directly?  I wrote another tool, gen_mo.lua, to find all of the .po files, compile them to .mo files and place those in dt.configuration.config_dir .. "/lua/locale/", which I create.  Dumping ... just gave me the relative path, i.e. from lib/dtutils.lua I got lib/dtuils.  I wrote a dtutils.pwd() function to get the current directory but then scrapped it when I thought I had gettext figured out.  I'll look at the darktable code and see what it's doing...

lib/dtutils/file.lua
```diff
@@ -0,0 +1,523 @@
...
+      toFile - string - name of file to copy to]],
+  Description = [[copy a file using a succession of methods from operating system
+    to a pure lua solution]],
+  Return_Value = [[result - boolean - nil on error, true on success]],
+}
+
+function dtutils_file.file_copy(fromFile, toFile)
+  local result = nil
+  -- if cp exists, use it
+  if dtutils_file.check_if_bin_exists("cp") then
```
wpferguson: This stemmed from using os.rename to move a file from $HOME/.local/tmp to wherever the collection was located.  When we crossed file system boundaries os.rename didn't work since it actually needed to do a copy, not just a rename.  I replaced os.rename with file_move() which attempts a rename and if that doesn't work then os.execture("mv..."), then if that fails  file_copy which tries to copy the file using os.execute("cp...") then finally a pure lua solution. So, in other words I'm trying to move the file using the least expensive method I can, then supplying alternatives in order of increasing expense.

lib/libLog.lua
```diff
@@ -0,0 +1,378 @@
+--[[
```
wpferguson: dtutils came from existing code pulled from the scripts.  libLog was a new thing and I didn't want to get the two mixed up.  Moving it to dtutil.log is fine.  I just have to make sure that I don't get in a circular require loop, requiring libraries from each other.  That gets nasty real fast ;).

lib/libLog.lua
```diff
@@ -0,0 +1,378 @@
...
+
+local dt = require "darktable"
+
+-- set the default log levels
+
+libLog.debug = {"DEBUG:", false}
+libLog.info = {"INFO:", false}
+libLog.warn = {"WARN:", false}
+libLog.error = {"ERROR:", true}
+libLog.success = {"SUCCESS:", true}
```
wpferguson: I thought about the dictionary thing a couple of days ago, when I changed everything back to just printing to stdout.  At that point I needed to get this out and let my mind clear a bit.  I'll go ahead and do this...

lib/libLog.lua
```diff
@@ -0,0 +1,378 @@
...
+        libLog.debug   - debugging messages
+        libLog.info    - informational messages
+        libLog.warn    - warning messages 
+        libLog.error   - error messages 
+        libLog.success - success messages
+      ... - string(s) - the message to print, which could be a comma separated set of strings
+
+  DESCRIPTION
+    msg checks the level to see if it is enabled, then prints the level type and message if it is.
+    debug messages print straight to the console, info and warn messages to dt.print_error() and error
```
wpferguson: That's why I pulled everything back to just console.  When I first thought up the log library, I thought about it writing everything and wrapping dt.print_error() and dt.print().  The more I used it the more I thought my original idea might not have been so good, so I pulled it back.  Then I thought of the dictionary idea so that we could send it where we wanted, but I had too much in my head to do it then.  I like the writing policy.  It gives me an idea where to go with this.

lib/libLog.lua
```diff
@@ -0,0 +1,378 @@
...
+
+  RETURN VALUE
+    none
+
+  REFERENCE
+    http://stackoverflow.com/questions/7148678/lua-print-on-the-same-line
+
+]]
+
+function libLog.print(...)
```
wpferguson: join only works on strings.  If I pass a table or a function and don't wrap it with type, then I get an error.  In my development/debugging I would often just print things to see what came back since print() was pretty forgiving with what I threw at it.  Kind of cheap data dump.  I wish that I had known about dprint() when I was developing the plugin stuff.  It would have helped a lot.

lib/libLog.lua
```diff
@@ -0,0 +1,378 @@
...
+  local write = io.write
```
wpferguson: I did this because print put tabs between arguments and I needed finer control over the output.  If we use the dictionary and the printing policy, then this should be ok.  Or, I can preprocess the message before I send it to print and not do io.write.  While we're on this, can we change dt.print_error to dt.print_log?  It always says ERROR no matter what your printing out.  This was the reason that I created libLog, so I could separate errors from information.

lib/libLog.lua
```diff
@@ -0,0 +1,378 @@
...
+
+--[[
+  NAME
+    set_level - set the logging level
+
+  SYNOPSIS
+    local log = require "lib/libLog"
+
+    log.set_level(level)
+      level - string - a string specifying the level, one of: "debug", "info", "warn", "error", "success"
```
wpferguson: I don't know why I didn't think of that.  Good catch.

tools/get_lib_manpages.lua
```diff
@@ -0,0 +1,63 @@
+--[[
```
wpferguson: I figured that I would give you a full solution that you could use rather than writing half a solution and making someone else come up with the other half.  That way you could use what you have, build on what I built or just use it as is.  I figured you might be a little busy ;)

lib/dtutils.lua
boucman: could you have a fixed "order of sections" table in your generators instead ? instead of repeating it in each doc section.

lib/dtutils.lua
boucman: I am thinking of non-manpage generation here
* name : just the name
* synopsis : one line description 
* usage : mainly arguments, return values and their description
* example : actual code

lib/dtutils.lua
```diff
@@ -0,0 +1,457 @@
...
+  Synopsis = [[local du = require "lib/dtutils"
+
+    du.update_combobox_choices(combobox_widget, choice_table)
+      combobox_widget - lua_combobox - a combobox widget
+      choice_table - table - a table of strings for the combobox choices]],
+  Description = [[Set the combobox choices to the supplied list.  Remove any extra choices from the end]],
+  Return_Value = [[none]],
+}
+
+function dtutils.update_combobox_choices(combobox, choice_table)
```
boucman: sounds good

I have thought about that one, and I understand why it makes me uneasy...

this is a "replace in place" function for table, i.e it takes two tables and set the first table to hav e the same elements than the second table

this is disturbing because it's something you should never have to do in lua, but the widget API (and maybe other API) need it.

This function probably can stay in dtutils, but with a different name. dtutils.replace_table_elements or something similar...

lib/dtutils/extensions.lua
boucman: no, it's not really relevant anymore

lib/dtutils/file.lua
boucman: if mv doesn't work, I wouldn't push further.... the system we're one is too broken and it's safer not to insist.

lib/libLog.lua
boucman: ok, but now you have dprint, you might as well use join :)

lib/libLog.lua
boucman: hmm, interesting... I'll lead print_error as-is, but i'll add a new print_log with no prefix so you can do that

print_error will stay for two reasons
* backward compatibility
* standardized prefix that can be used to easily filter the output

lib/dtutils/processor.lua
boucman: yes, that's what I thought, but the explanation isn't clear...

lib/dtutils.lua
wpferguson: How about dtutils.widgets?

lib/libLog.lua
wpferguson: sounds good to me :)

lib/dtutils.lua
wpferguson: how about 

<libname>.libdoc = {
  "Name" = [[]],
  "Synopsis" = [[]],
  "Usage" = [[]],
  "Description" = [[]],
  "Return Value" = [[]],
  "Limitations" = [[]],
  "See Also" = [[]],
  "Reference" = [[]],
  "License" = [[]],
  "Copyright" = [[]],
  functions = {}
}

 for the lib and 

<libname>.libdoc.functions["name of routine"] = {
  "Name" = [[]],
  "Synopsis" = [[]],
  "Usage" = [[]],
  "Description" = [[]],
  "Return Value" = [[]],
  "Limitations" = [[]],
  "See Also" = [[]],
  "Reference" = [[]],
  "License" = [[]],
  "Copyright" = [[]],
}

for the functions. 

Fill in what you need and leave the rest empty.  When printing just skip the empty sections.

lib/libLog.lua
wpferguson: sounds good to me :)

lib/dtutils.lua
wpferguson: done

lib/dtutils.lua
```diff
@@ -0,0 +1,457 @@
...
+  RETURN VALUE
+    result - a table of strings on success, or an empty table on error
+
+  EXAMPLE
+    split("/a/long/path/name/to/a/file.txt", "/") would return a table like
+      {"a", "long", "path", "name", "to", "a", "file.txt"}
+
+]]
+
+dtutils.libdoc.functions[#dtutils.libdoc.functions + 1] = {
```
wpferguson: Used the name of the function

lib/dtutils.lua
```diff
@@ -0,0 +1,457 @@
...
+  Description = [[split separates a string into a table of strings.  The strings are separated at each
+    occurrence of the supplied pattern.]],
+  Return_Value = [[result - a table of strings on success, or an empty table on error]],
+  Example = [[split("/a/long/path/name/to/a/file.txt", "/") would return a table like
+      {"a", "long", "path", "name", "to", "a", "file.txt"}]],
+}
+
+-- Thanks to http://lua-users.org/wiki/SplitJoin
+function dtutils.split(str, pat)
+   local t = {}  -- NOTE: use {n = 0} in Lua-5.0
```
wpferguson: done

lib/dtutils.lua
wpferguson: I now return a status and either the lib or an error message

lib/dtutils.lua
wpferguson: In dtutils.processor for now...

lib/dtutils.lua
```diff
@@ -0,0 +1,457 @@
...
+    log.msg(log.debug, "Setting choice " .. i .. " to " .. name)
+    combobox[i] = name
+  end
+  if choices < items then
+    for j = items, choices + 1, -1 do
+      log.msg(log.debug, "Removing choice " .. j)
+      combobox[j] = nil
+    end
+  end
+  combobox.value = 1
```
wpferguson: done

lib/dtutils/debug.lua
wpferguson: done

lib/dtutils/extensions.lua
wpferguson: Did it where I could

lib/dtutils/extensions.lua
```diff
@@ -0,0 +1,185 @@
+--[[
+
+  NAME
+    dtutils/extensions.lua - a library of extenstions to the lua libraries
+
+  SYNOPSIS
+    require "lib/dtutils.extensions"
+
+  DESCRIPTION
+    This library contains extensions to the lua libraries.
```
wpferguson: It's now dtutils.string with a slightly longer description ;)

lib/dtutils/file.lua
wpferguson: I have an idea, but I need to play with it...

lib/libLog.lua
wpferguson: It's dtutils.log

lib/dtutils/file.lua
wpferguson: Changed file_move to stop after the mv.  Removed file_copy.

lib/libLog.lua
wpferguson: Can't use join from dtutils because dtutils requires the log library and if the log library required dtutils it would get real ugly real fast.  I worked around it though...

lib/libLog.lua
wpferguson: done

lib/libLog.lua
wpferguson: done

lib/libLog.lua
wpferguson: done

lib/dtutils/processor.lua
wpferguson: image_table it is...  ;)

lib/dtutils/file.lua
wpferguson: How about we break this up into 2 problems?  

Problem 1 is where do we put the .mo files.  I think we can solve that one now.  My vote is a locale directory in the lua-scripts directory.  That way we don't have all the different language directories scattered through out the base directory.

Problem 2 is how to get the .mo files to where we decide they should belong.  I'm still thinking on that one, but I have an idea...  ;)

lib/dtutils.lua
boucman: ok, let's go for that...

boucman:
At this point, it looks good for me @supertobi,@houz any opinions ?

supertobi:
Looks good.

tools/get_lib_manpages.lua
```diff
@@ -0,0 +1,67 @@
...
+  if mf then
+    mf:write(".TH " .. string.upper(name) .. " 3 \"\" \"\" \"Darktable " .. libname .. " functions\"\n")
+    for _,section in ipairs(keys) do
+      if d[section]:len() > 0 then
+        mf:write(".SH " .. string.upper(string.gsub(section, "_", " ")) .. "\n")
+        mf:write(d[section] .. "\n")
+      end
+    end
+    mf:close()
+    os.execute("groff -man " .. fname .. " | ps2pdf - " .. fname .. ".pdf")
```
supertobi: What about an check_if_bin_exists here to check for groff and ps2pdf?

wpferguson: good idea  :)

supertobi:
Lets go with this. If we find problems we can fix them later.

PR #75, opened 2016-10-18
copy attach detach add translation submitted by Holger Klemm

chrik5:
Hi,

Holger Klemm submitted a translated version of my script. 

Christian


supertobi:
Sorry, I clicked the wrong button fist.

PR #76, opened 2016-10-19
New storage: file to collection

wpferguson:
Created a new storage that exports an image then imports it back into the collection and groups it with the original.  Tags and ratings are copied from the original to the import.

### DIFF
--- a/contrib/export2collection.lua
+++ b/contrib/export2collection.lua
@@ -1,0 +1,191 @@
+--[[
+
+    export2collection.lua - export a file and import it to the collection
+
+    Copyright (C) 2016 Bill Ferguson <wpferguson@gmail.com>.
+
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 3 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+]]
+--[[
+    export2collection - export an image then import it into the collection
+
+    This script provides another storage (export target) for darktable.  Selected
+    images are exported in the specified format to temporary storage.   The exported 
+    files are moved into the current collection and imported into the database.  
+    The imported files then show up grouped with the originally selected images.
+    If there is a filename conflict on import, the incoming filename will have an 
+    increment added to it, i.e. _7D_1234.jpg would be _7D_1234_01.jpg.
+
+    USAGE
+    * require this script from your main lua file
+    * select an image or images to export
+    * in the export dialog select "file to collection" and select the format and bit depth for the
+      exported image
+    * Press "export"
+    * The exported image will be imported and grouped with the original image
+
+    BUGS, COMMENTS, SUGGESTIONS
+    * Send to Bill Ferguson, wpferguson@gmail.com
+
+]]
+
+local dt = require "darktable"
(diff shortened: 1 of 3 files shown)


contrib/export2collection.lua
```diff
@@ -0,0 +1,192 @@
...
+dt.configuration.check_version(...,{3,0,0})
+
+-- Tell gettext where to find the .mo file translating messages for a particular domain
+gettext.bindtextdomain("export2collection",dt.configuration.config_dir.."/lua/locale/")
+
+local function _(msgid)
+    return gettext.dgettext("export2collection", msgid)
+end
+
+local function split_filepath(str)
```
boucman: you should start using dtutils, since it's on its way in :)

contrib/export2collection.lua
```diff
@@ -0,0 +1,192 @@
...
+  -- for each of the image, exported image pairs
+  --   move the exported image into the directory with the original  
+  --   then import the image into the database which will group it with the original
+  --   and then copy over any tags other than darktable tags
+
+  for image,exported_image in pairs(image_table) do
+
+    local myimage_name = image.path .. "/" .. get_filename(exported_image)
+
+    while check_if_file_exists(myimage_name) do
```
boucman: this loop could be made into a utility function find_unique_file_name or something like that (in dtutils.file)

wpferguson: I agree.  Consider it done...

contrib/export2collection.lua
wpferguson: I actually copied and pasted from dtutils.file for the new check_if.. functions.  ;)

wpferguson:
There was a feature request in the mailing list and I looked at it and figured I could write that script in 30 minutes.  I took gimp.lua and ripped out the GIMP specific part and got this.  There are 2 use cases I can think of right now.  One is official/enfuse.lua requires ldr files, so this would be a way to create them then use the enfuse module.  The other is a user named Harry Durgin who has made several darktable videos.  His workflow is do a bunch of edits, then export to compress the history stack, then repeat.

Maybe, as Jeff Goldblum said in Jurassic Park, I should have asked "Should we?" instead of "Can we?".

houz:
I wonder, shouldn't we rather add a signal that gets fired when an image got exported, so you can use the regular export storage and just import/group when an image got exported? Or is this something that is only intended to be used once in a while?

boucman:
lua can already add storages (if I understand houz correctly) but the point is that we are talking about tools that do image=>external-tool=>reimport.

the general, vague, consensus is that such a tool shouldn't be a strorage (from a UI point of view) but something else... it can't really be a iop either (even by stretching the definition) because it can involve multiple images (enfuse)

I'm not sure how to call such tools, nor how to handle them in the UI. Right now we are (poorly) reproducing the export-lib UI in lua in order to separate them from storage. Maybe we could have a separate export-like UI for that (on the C side, I mean) I don't know...

@wpferguson : I am a bit confuse by the "export to compress the history stack" step you describe... what do you mean ? is that a workaround for a DT bug ?

houz:
I know that we can add Lua storages. What I meant was a new signal, similar to the intermediate-export-image one, which a script can register for, getting the filename of the exported image + image object of the original + the storage used. Then the script could check if the storage is "disk" and import the exported image and group it with the original.

boucman:
Ok, so right now there is no real point in doing this, i'll close the PR

wpferguson:
@boucman here is a link to one of harry durgin's videos - https://www.youtube.com/watch?v=QbQA2P2EM1w.  At 26 minutes he exports his raw to a tif and brings it back in.  My understanding of what he is doing is applying all the changes so that he doesn't have to wait while the image processes to update the display when he changes something.  It may also be that he doesn't know how to compress the history stack, so he thinks darktable is doing more processing than it actually is.

houz:
He does that in every video for two reasons:

- to make darktable faster as he tends to pile up duplicates of the modules
- to battle the fixed pipe order. By using intermediate exports he can use an early module on the result of a late one.

PR #80, opened 2016-12-14
Add new GUI entry in lighttable for renaming tags

GHswitt:
Rename tag: A new tag is added to all images containing old tag. After that, the old tag is removed from all images (and removed from tags). The GUI entry has two entries: Old tag, New tag and a Go button.

### DIFF
--- a/contrib/rename-tags.lua
+++ b/contrib/rename-tags.lua
@@ -1,0 +1,114 @@
+--[[
+Rename tags
+
+AUTHOR
+Sebastian Witt (se.witt@gmx.net)
+
+INSTALLATION
+* copy this file in $CONFIGDIR/lua/ where CONFIGDIR
+is your darktable configuration directory
+* add the following line in the file $CONFIGDIR/luarc
+  require "rename-tags"
+
+USAGE
+* In lighttable there is a new entry: 'rename tag'
+* Enter old tag (this one gets deleted!)
+* Enter new tag name
+
+LICENSE
+GPLv2
+
+]]
+
+local darktable = require "darktable"
+local debug = require "darktable.debug"
+
+-- GUI entries
+local old_tag = darktable.new_widget("entry") { tooltip = "Enter old tag" }
+local new_tag = darktable.new_widget("entry") { tooltip = "Enter new tag" }
+
+-- This function does the renaming
+local function rename_tags(widget)
+  -- If entries are empty, return
+  if old_tag.text == '' then
+    darktable.print ("Old tag can't be empty")
+    return
+  end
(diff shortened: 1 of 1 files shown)


GHswitt:
Just found out that it updates the image tags in the database, but the XMP files are not updated automatically. The XMP has still the old value in it. Only when I open an image in darkroom and switch back to lighttable, the new tag from the database gets written to the XMP. How can I do that from LUA?

supertobi:
For me this sounds like a bug. @boucman what do you think?

houz:
Maybe we should wire dt_tag_reorganize() into the Lua API instead – and check if it actually works as it's unused in current dt. 😄

chrik5:
Definitly, yes. This will make the proposed script very short. 
Finally, I will be able to fix all my tag typos. Thx @GHswitt for starting this!

GHswitt:
No problem, I just needed a way to do renaming more comfortable.

GHswitt:
I looked into dt_tag_reorganize, but couldn't find where it synchronizes the new tags to the XMP. Can anyone give me a hint?

GHswitt:
As far as I can see it, dt_control_write_sidecar_files() or dt_image_write_sidecar_file should also be available via Lua API.

houz:
It doesn't. It's been bitrotting for years.

boucman:
* feature freeze, no new api (obviously)
* even if we wire a tag renaming primitive in lua, we need to figure out what's going on with the XMP. this is a bug, your code should work
* maybe DT needs a core way to rename tags ?

spaceChRiS:
Maybe that's the right time and place to start another discussion regarding the topic. Renaming tags is a feature I am looking for as well. But it is something that I would need several times a year, but not very often. This is something that would perfectly fit into a regular menu structure, that darktable avoids so far. But that may not be a necessary restriction for lua scripts. For the functionality of tag renaming, I think about a menu entry “Rename tag” that opens a little window with a search text entry field, a list of tags found, where one tag could be marked, a text entry field that gives the new tag and an “apply” button. A new module that is always visible, at minimum in the collapsed state, does IMO not really fit to the purpose of the functionality of tag renaming.

supertobi:
FYI
darktable 2.2.2 released:
https://www.darktable.org/2017/01/darktable-2-2-2-released/

Bugfixes:
...
* Lua: sync XMPs after some tag manipulations

...

supertobi:

GHswitt:
Not from my side. I'm still using rename-tags.lua, works ok.

supertobi:
OK, then I'll pull this script.

PR #83, opened 2017-01-24
This scripts work with API version 5.0.0 & small fix for hugin

chrik5:
This scripts work with API version 5.0.0.

About the hugin error message i'm not sure. Evertime i close hugin after a succesfull export to hugin this popped up in darktable. i think it should not.

Christian


supertobi:
Looks good.

PR #87, opened 2017-02-28
Change abbreviation method of the quicktag script

spaceChRiS:
Abbreviate in a way that the final branch in the tag tree is never abbreviated but the tree branches are abbreviated to `<...>|<firstletter>..<lastletter>|<...>`. This is much more handy at least for me, since otherwise, e.g. names become hardly distinguishable within one family if the tags end `...|first_name last_name` and the last name is long.

### DIFF
--- a/contrib/quicktag.lua
+++ b/contrib/quicktag.lua
@@ -104,12 +104,24 @@ local function abbrevate_tags(t)
   for i=1,qnr do
     local taglength = string.len(t[i])
     if taglength > max_label_length then
-      local max_2 = math.floor(max_label_length/2)
-      quicktag_label[i] = string.sub(t[i],0,max_2).."..."..string.sub(t[i],taglength - max_2)
+
+      s = {}; k = 0
+      for str in string.gmatch(t[i], "([^|]+)") do
+        s[k] = str
+        k = k + 1
+      end
+
+      quicktag_label[i] = ""
+      k = 0
+      while s[k+1] do
+        quicktag_label[i] = quicktag_label[i] .. string.sub(s[k],0,1) .. ".." .. string.sub(s[k],-1) .. "|"
+        k = k + 1
+      end
+      quicktag_label[i] = quicktag_label[i] .. s[k]
+
     else
       quicktag_label[i] = t[i]
     end
-    -- print(quicktag_label[i])
   end
 end
 


contrib/quicktag.lua
```diff
@@ -104,12 +104,24 @@ local function abbrevate_tags(t)
...
+      end
+
+      quicktag_label[i] = ""
+      k = 0
+      while s[k+1] do
+        quicktag_label[i] = quicktag_label[i] .. string.sub(s[k],0,1) .. ".." .. string.sub(s[k],-1) .. "|"
+        k = k + 1
+      end
+      quicktag_label[i] = quicktag_label[i] .. s[k]
+
```
chrik5: But this does not handle long tags anymore or the case where the last subtag is too long?

spaceChRiS: That's true, but I think it's more the tree than the tag itself which causes a problem. The tree description (`...|...|...|...|`) becomes long really fast. And, even if you count letters, with a proportional font, all of them have different sizes and you only get a lose statistical connection between the length of a string on screen and the number of letters. One could think about special handling of the tag itself (opposed to tree branches), but I would rather test if the proposed approach isn't enough or even better in terms of readability/recognizability of the tags.

chrik5: Well, even if the correlation between the number of strings and width is not so good, I would include an option to shorten it. I have no idea about other languages as well. I would simply check if string.len(quicktag_label[i] .. s[k]) > max, calc how much is the rest for s[k] and split s[k] like it was before and concatenate it with the tree.

contrib/quicktag.lua
```diff
@@ -104,12 +104,24 @@ local function abbrevate_tags(t)
   for i=1,qnr do
     local taglength = string.len(t[i])
     if taglength > max_label_length then
-      local max_2 = math.floor(max_label_length/2)
-      quicktag_label[i] = string.sub(t[i],0,max_2).."..."..string.sub(t[i],taglength - max_2)
+
+      s = {}; k = 0
```
chrik5: s and k must be local

contrib/quicktag.lua
spaceChRiS: I wonder if it would be even better if we had just an on/off switch for this method, or a selection of several abbreviation methods, some with additional parameters (your original code) and some without. Is something like that feasible in the configuration dialogue?

supertobi:

chrik5:
No, not from my side. Can be closed I think.
It is on my list to look into this, because I will try to arrange the UI buttons in two columns.

supertobi:
I'll now close this ticket. Please reopen if needed.

PR #88, opened 2017-04-08
Initial version of PDF Slideshow.

TurboGit:
A simple PDF slideshow generator. Each image is on a single slide and
a title bar at the bottom is added with the slide show general title,
the image author and title if defined in the metadata.

This script requires pdflatex.

### DIFF
--- a/contrib/pdf_slideshow.lua
+++ b/contrib/pdf_slideshow.lua
@@ -159,6 +159,7 @@ dt.register_storage("pdf_slideshow",_("pdf slideshow"),
     local preamble = [[
     \documentclass[a4paper,10pt,landscape]{beamer}
     \usetheme{default}
+    \usepackage[utf8]{inputenc}
     \usepackage{graphicx}
     \usepackage[space]{grffile} % needed to support filename with spaces
     \pagestyle{empty}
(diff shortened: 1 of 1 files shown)


contrib/pdf_slideshow.lua
```diff
@@ -0,0 +1,229 @@
...
+         my_write(latexfile,"\n\\bigskip\n")
+         i = i+1
+      end
+      my_write(latexfile,ending)
+      latexfile:close()
+
+      -- convert to PDF
+      local dir=string.gsub(filename, "(.*/)(.*)", "%1")
+      local locfile=string.gsub(filename, "(.*/)(.*)", "%2")
+      local command = "pdflatex -halt-on-error -output-directory "..dir.." "..locfile
```
supertobi: You should check if pdflatex is installed. You can use the function from dtutils.file:
Just add "local df = require "lib/dtutils.file"" at the top and then do a check like this:
  if not df.check_if_bin_exists("pdflatex") then
    dt.print_error(_("pdflatex not found"))
    return
end	

Here you can see an example:
https://github.com/darktable-org/lua-scripts/blob/master/contrib/geoToolbox.lua#L313

TurboGit:
Thanks for the review. I have added a check for pdflatex and fixed some issues.

PR #90, opened 2017-04-10
general translation reorg.

TurboGit:


supertobi:
This is something I'd like to discuss first with the other developers, especially @boucman here before pushing. Comments from everyone are welcome.

TurboGit:
@supertobi, sure but note that current implementation does not work so either the way I fixed it or any other way but I think we need to do something for the organization of the translation artifacts.

supertobi:
OK, then I'll commit it now. If someone has a problem with this dir, we can change it later.

PR #95, opened 2017-08-07
Fixed quotation marks typos

elman22:
There are some extra quotation marks, that prevent the script from being loaded with Darktable.


supertobi:
thanks

PR #97, opened 2017-09-13
Selection of untagged images

jbvoelker:
When tagging my collection, I was looking for a way to find untagged images in darktable. All I found were older threads and a [feature request from five years ago](https://redmine.darktable.org/issues/8792).

So I wrote a lua script which checks the collection for untagged images and adds them to the selection. Images having no tags at all or only tags added by darktable (e.g. darktable|format|cr2) are considered untagged.

Maybe this is interesting for some other users, or did I miss an official way to show untagged images?

### DIFF
--- a/contrib/select_untagged.lua
+++ b/contrib/select_untagged.lua
@@ -1,0 +1,65 @@
+--[[
+    This file is part of darktable,
+    copyright (c) 2017 Jannis_V
+
+    darktable is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+    darktable is distributed in the hope that it will be useful,
+     but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+    You should have received a copy of the GNU General Public License
+    along with darktable.  If not, see <http://www.gnu.org/licenses/>.
+]]
+
+--[[
+Enable selection of untagged images (darktable|* tags are ignored)
+]]
+
+local dt = require "darktable"
+require "official/yield"
+local gettext = dt.gettext
+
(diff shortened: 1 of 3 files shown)


contrib/select_untagged.lua
```diff
@@ -0,0 +1,65 @@
...
+    You should have received a copy of the GNU General Public License
+    along with darktable.  If not, see <http://www.gnu.org/licenses/>.
+]]
+
+--[[
+Enable selection of untagged images (darktable|* tags are ignored)
+]]
+
+local dt = require "darktable"
+require "official/yield"
```
supertobi: I think this "require" is not needed in this script.

contrib/select_untagged.lua
```diff
@@ -0,0 +1,65 @@
...
+local gettext = dt.gettext
+
+dt.configuration.check_version(...,{3,0,0},{4,0,0},{5,0,0})
+
+-- Tell gettext where to find the .mo file translating messages for a particular domain
+gettext.bindtextdomain("select_untagged",dt.configuration.config_dir.."/lua/locale/")
```
supertobi: Do you have any translations, that you could add to the pr?

jbvoelker: Not yet, but I will add english and french translations.

jbvoelker: I have created a file containing a german translation (locale/de_DE/LC_MESSAGES/select_untagged.po).
The translation is not applied in darktable 2.2.5, gettext.dgettext always returns the original message, the other GUI elements which are not added by scripts are in german as expected.
According to inotifywait, the file is not even accessed by darktable, same with the translations of other scripts.
Does the translation in scripts work for you in 2.2.5?

jbvoelker: Should I include the .mo file, or is it left to the user of the script to build it?

supertobi: As we don't have a build system for the mo file I would add them.

jbvoelker: I have added them to my branch.

supertobi: then please make a new pull request

PR #99, opened 2017-10-11
Support for importing the in-camera ratings from Fujifilm cameras

Sitwon:


contrib/fujifilm_ratings.lua
```diff
@@ -0,0 +1,56 @@
...
+- exiftool (https://www.sno.phy.queensu.ca/~phil/exiftool/)
+
+--]]
+
+darktable = require "darktable"
+
+local function detect_rating(event, image)
+	local RAF_filename = tostring(image)
+	local JPEG_filename = string.gsub(RAF_filename, "%.RAF$", ".JPG")
+	local command = "exiftool -Rating " .. JPEG_filename
```
supertobi: You could add a check if exiftool is installed on the system like this:
if not checkIfBinExists("exiftool") then
   return
end

contrib/fujifilm_ratings.lua
```diff
@@ -0,0 +1,56 @@
...
+with this program; if not, write to the Free Software Foundation, Inc.,
+51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+Dependencies:
+- exiftool (https://www.sno.phy.queensu.ca/~phil/exiftool/)
+
+--]]
+
+darktable = require "darktable"
+
```
supertobi: Please check the version you have tested the script with, like this:

darktable.configuration.check_version(...,{4,0,0})

contrib/fujifilm_ratings.lua
```diff
@@ -0,0 +1,56 @@
...
+	local JPEG_filename = string.gsub(RAF_filename, "%.RAF$", ".JPG")
+	local command = "exiftool -Rating " .. JPEG_filename
+	darktable.print_error(command)
+	local output = io.popen(command)
+	local jpeg_result = output:read("*all")
+	output:close()
+	if string.len(jpeg_result) > 0 then
+		jpeg_result = string.gsub(jpeg_result, "^Rating.*(%d)", "%1")
+		image.rating = tonumber(jpeg_result)
+		darktable.print_error("Using JPEG Rating: " .. tostring(jpeg_result))
```
supertobi: You could earn some bonus point if you would make the strings translatable. But that is not really needed for error messages.
Have a look at:
 lua-scripts/examples/gettextExample.lua
or
 lua-scripts/contrib/gimp.lua

contrib/fujifilm_ratings.lua
```diff
@@ -0,0 +1,56 @@
...
+	if string.len(raf_result) > 0 then
+		raf_result = string.gsub(raf_result, "^Rating.*(%d)", "%1")
+		image.rating = tonumber(raf_result)
+		darktable.print_error("Using RAF Rating: " .. tostring(raf_result))
+	end
+end
+
+darktable.register_event("post-import-image", detect_rating)
+
+darktable.print_error("fujifilm_ratings loaded.")
```
supertobi: I think here is a print better then a print_error

houz:
Why do you want this as a Lua script? If you provide us with a sample file we can probably teach dt itself to understand that field.

Sitwon:
You're right, it would probably be better if darktable could handle it directly, like it does for Nikon NEF files. I guess I'm just used to having to fix my own issues with open source projects.

Some things I've learned from trial and error:
1) If you shoot in JPEG+RAW, the in-camera Rating will be written only to the JPEG file, so on import you need to look for the JPEG and read the Rating from there.
2) If you shoot in RAW only, the in-camera Rating will be written to the RAF file.
3) In my first pass at this problem, I wrote a Bash script to create a `<filename>.XMP` file, and discovered that darktable looks for `<filename>.xmp` (case-sensitive extension matching).
4) A lot of tools seem to not understand Fujifilm's MakerNotes fields, but exiftool does (at least in recent versions).

I have put some samples here: https://drive.google.com/open?id=0B0p0PF3h5F9kTkF1X3gycmkweXM

houz:
Bad news: At least my copy of exiv2 (version 0.25) doesn't understand that tag yet. Until that changed there isn't much we can do on darktable's side. So you will have to keep using your script. Once that changes we will of course add support for it.

contrib/fujifilm_ratings.lua
Sitwon: The version I have is 2.2.5, but I notice scripts in the repo claiming 4,0,0 and 5,0,0. Could you explain how the version numbering works? Should I use 2,2,5 or 4,0,0?

supertobi: The API Version is not the same then the darktable version. You can get the version with  dt.configuration.api_version_string. Or simply use this script:

https://github.com/darktable-org/lua-scripts/blob/master/examples/api_version.lua

Sitwon:
@supertobi I think I've addressed all of your feedback, thanks.

supertobi:
Here we go, your script is in.

contrib/fujifilm_ratings.lua
houz: Or just run `darktable --version`.

PR #103, opened 2018-01-03
GPX export script

jbvoelker:
Generates a GPX track from all selected images containing geo information. I mainly use this to write geo information directly to my RAW files using exiftool.


supertobi:
Thanks for the script.

PR #104, opened 2018-01-07
New script: "Labels to Tags"

schwerdf:
This is a script that I use extensively as a tagging aid. When invoked it will check the color labels and rating of each selected image against a provided list of patterns and, for each one that matches, attach a corresponding set of tags to the image. Tag mappings can be specified in `luarc` or elsewhere by calling a function provided in the module.

### DIFF
--- a/contrib/LabelsToTags.lua
+++ b/contrib/LabelsToTags.lua
@@ -1,0 +1,225 @@
+--[[
+   LABELS TO TAGS
+   Allows the mass-application of tags using color labels and ratings
+   as a guide.
+
+   AUTHOR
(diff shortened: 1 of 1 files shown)


supertobi:
Whouldn't it be better to define the mapping in the preferences dialog instead of the luarc?

schwerdf:
I tried to use the preferences at first, but the only way I could think of to hold mappings in a preference was just to serialize them all into a single-line string using JSON or something similar. I found it much easier to specify the mappings in a file as Lua tables.

supertobi:
OK, here we go.

PR #105, opened 2018-01-15
Updated KML/KMZ export script

elman22:
- correct handling of duplicates
- path is sorted by exif datetime taken
- added more characters for transliteration
- code formatting

### DIFF
--- a/contrib/kml_export.lua
+++ b/contrib/kml_export.lua
@@ -82,57 +82,108 @@ end
 -- Copied from https://forums.coronalabs.com/topic/43048-remove-special-characters-from-string/
 function string.stripAccents( str )
   local tableAccents = {}
+    -- A
     tableAccents["à"] = "a"
-    tableAccents["á"] = "a"
-    tableAccents["â"] = "a"
-    tableAccents["ã"] = "a"
-    tableAccents["ä"] = "a"
-    tableAccents["ç"] = "c"
-    tableAccents["è"] = "e"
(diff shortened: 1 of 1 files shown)


contrib/kml_export.lua
```diff
@@ -180,7 +244,7 @@ local function create_kml_file(storage, image_table, extra_data)
         return
     end
 
-    dt.print_error("Will try to export KML file now")
+    dt.print_log("Will try to export KML file now")
```
supertobi: I don't like the use of print_log here, because print_log is only available in the latest darktable version. There are still a lot of users with darktable 2.2 out there

elman22:
OK, reverted print_log to print_error and changed indentation from 4 spaces to 2 spaces

supertobi:
Th code looks good now. Thanks. Now I have to find some time for a real test.

Perhaps you could have a look into geoJSON_export.lua and make the duplicate change and the sorting change there too.

contrib/kml_export.lua
houz: In the past we forked the Lua repo when doing a dt release, so for people using stable or old darktable versions there is a matching set of compatible scripts. Maybe we should bring that back?

wpferguson: I think @houz has a good point.  Now that there is a windows port of DT we may have to introduce compatibility changes to get some of the scripts to run, such as using the new os detection feature.

PR #108, opened 2018-02-23
fixed a couple dtutils_file functions for windows platform

achienbsi:
Fixed a couple functions in the lib/dtutils/file are not compatible with Windows. Also fixed a bug in the create_unique_filename function.

### DIFF
--- a/lib/dtutils/file.lua
+++ b/lib/dtutils/file.lua
@@ -51,7 +51,13 @@ dtutils_file.libdoc.functions["check_if_bin_exists"] = {
 }
 
 function dtutils_file.check_if_bin_exists(bin)
-  local result = os.execute("which " .. bin)
+  local cmd
+  if package.config:sub(1,1) == '/' then
+	cmd = 'which '
+  else
+    cmd = 'where '
+  end
+  local result = os.execute(cmd .. bin)
   if not result then
     result = false
   end
@@ -194,10 +200,13 @@ dtutils_file.libdoc.functions["check_if_file_exists"] = {
 }
 
 function dtutils_file.check_if_file_exists(filepath)
-  local result = os.execute("test -e " .. filepath)
-  if not result then
-    result = false
+  local file = io.open(filepath, "r")
+  local result = false
+  if file then
+    result = true
+	file:close()
   end
+
   return result
 end
 
@@ -384,7 +393,7 @@ function dtutils_file.create_unique_filename(filepath)
   while dtutils_file.check_if_file_exists(filepath) do
     filepath = dtutils_file.filename_increment(filepath)
     -- limit to 99 more exports of the original export
-    if string.match(dtfileutils.get_basename(filepath), "_(d-)$") == "99" then 
+    if string.match(dtutils_file.get_basename(filepath), "_(d-)$") == "99" then
       break 
     end
   end


wpferguson:
I read the issue, shell command "which" does not work on macOS, not too long ago and started thinking about how to get the lua scripts to run on windows.  With the delivery of 2.4, we now have built in OS detection, which @supertobi used in his prototype fix for macOS.  I set up a windows environment with dt 2.4, gimp, hugin, and enfuse and started trying to figure out how to find and use the executables.  I figured out using where instead of which, but then I ran into another problem.  Windows executables all have .exe appended to the filenames.  Therefore, check_if_bin_exists("gimp") wont find gimp.exe.  Even if it somehow managed to find gimp.exe, gimp.lua still wont work because it's going to try and execute "gimp", unless the gimp.exe executable is in the path.  At this point I started trying to figure out the best way to get the scripts working on macOS, windows, and unix/linux.

First, check_if_bin_exists needs to be a whole lot smarter.  Maybe we need to call check_if_bin_exists which in turn calls OS specific versions to find the executable.  check_if_bin_exists currently returns the executable, but we ignore that and just check for a value.  We need to modify all the scripts that use an external executable to use the return value from check_if_bin_exists for the executable.

check_if_file_exists may need a similar OS specific implementation, or we can just open the file to see if it exists.

Thoughts @houz, @boucman, @supertobi?

achienbsi:
Yea I think these functions doing OS calls need OS specific implementation. I was actually just about to close this request because I've just discovered that file_move is doing the same thing. A proper overhaul is probably needed.

chrik5:
@wpferguson On windows a file chooser in the preferences maybe is the only option to find the executeable in a save way. I use TeXstudio and there is a long list for of external commands in the preferences which need to be setup correctly. 
Only some programs set the PATH variable on windows. darktable itself does it not during install. Progams can be installed in different locations including different drive letters. Executeables can have different names. E.g. what happens if you install different versions of Gimp. How to chose the correct one automatically? They later will apply to all OS I guess?

supertobi:
1. @achienbsi Could you create a new PR for the bug fix.
2. I think @chrik5 is right the only sane solution for window is a preference where you can select the exe.
Then we can do a simple check if the file exists.

wpferguson:
@chrik5 I thought the problem of having different versions of an executable installed and didn't come up with an answer.  I think you're right about setting preferences.  That would also solve the macOS issue.

@chrik5, @supertobi, @achienbsi Any suggestions for an implementation?  Do we set it in each individual script, or do we have a global configuration script that contains settings for all external executables?

achienbsi:
Unless I am missing something (still relatively new to DT), I think it makes more sense for individual script to register their preferences since there are multiple strings a preference registration need that is specific to each script  And in that case do we even need separate check_if_bin_exists and check_if_file_exists? They do the same thing.

achienbsi:
@wpferguson @chrik5 @supertobi ran into another problem:

1. I made the change to check_if_file_exists to use io.read instead of test which seems to fail for checking files in c:\program files\* directory. I can fix this issue by using the native windows command "if exist file (cmd /c exit 0) else (cmd /c exit 1)"  I'm not sure if OSX can continue to use the test command or not.

2. Both dt.control.execute and os.execute seem to have trouble launching an external program when it is specified using the full path. I am already putting quotes around the exe to deal with the space character but it doesn't seem to get passed on. Here's the DT output:

LUA ERROR "C:\Program Files\Hugin\bin\hugin.exe" "C:\Users\achien\test.jpg" 
'C:\Program' is not recognized as an internal or external command,
operable program or batch file.

supertobi:
@achienbsi 
For 1. I think we'll need special code for every OS here. Something like this:
```
function check_if_bin_exists(bin)
  local result
  if (dt.configuration.running_os == "windows") then
    local f=io.open(bin,"r")
    if f~=nil then 
	  io.close(f) 
	  return true 
	else
	  return false 
	end  
  elseif (dt.configuration.running_os == "linux") then
    result = os.execute("which " .. bin)
  elseif (dt.configuration.running_os == "macos") then
    result = os.execute() -- ToDo: We need something here
  else
    result = false
  end
  if (not result or result == nil) then
    result = false
  end
  return result
end
```

For 2. Could you try this:
`"C:\\Program Files\\Hugin\\bin\\hugin.exe" `

supertobi:
@wpferguson 
Until your question I thought about handling it in every script. But there are some tools that are used in more then one script. (e.g. imagemagic, ffmpeg, ...) It would be annoying to set the path several times.
Now I thing it would be good to do something like this:
- A new utility function: "create_preference_path_if_not_exist("program_name")" 
- A new utility function: "get_preference_path("program_name")" 
- In every individual script we can now call create_preference_path_if_not_exist(). If an other script already created such an preference nothing happens, otherwise the new entry in the preference dialog will be added.
Of cause the new create_preference_path_if_not_exist() should only create the preferences on Windows and MacOS.
- get_preference_path() should then return the value from the parameter on Windows and MacOS and on Linux just the input parameter.

What do you think?

wpferguson:
@supertobi 
There is actually  a use case in linux for this also.  When you are calling a binary running under wine (i.e. the nik collection) you need to supply the full path.  Calling from the script could take this into account (i.e. the only way I can run the nik collection on linux is with wine) and show the file selector.

Now some questions...

If we already have a preference, do we show the preference in case the user wants to change it, or do we just hide it?  If we hide it, how should we update the preferences?

On linux, if a preference doesn't exist should we create one with the result from check_if_bin_exists or should we not fix it because it isn't broke?

I'll try and code something up in the next couple of days.

wpferguson:
[lua.zip](https://github.com/darktable-org/lua-scripts/files/1770074/lua.zip)

@achienbsi Here's a working gimp and a hugin for you to test.  The only modified files are lib/dtutilis/file.lua, contrib/hugin.lua and contrib/gimp.lua.  Look them over, try them out, see what you think.

One thing I discovered while testing on windows is if you install the git package, it supplies a git bash shell.  You can run darktable -d lua in the shell and the debugging prints in the shell while it's running rather than searching for a log file.

PR #109, opened 2018-02-23
improved a few aspects of the hugin script

achienbsi:
* handle all the stiching steps from hugin_executor without the gui
* automatically create the panorama file name based on the first selected file plus _pano suffix
* automatically detects if output file exists, if so use an unique name
* import the pano file after generation
* delete temporarily exported file


supertobi:
Thanks, that loogs good.

PR #110, opened 2018-02-26
hugin plugin improvements

achienbsi:
* added preference requiring user to specify the paths for hugin, hugin_executor and pto_gen needed for this plguin
* added preference to allow user to choose to launch hugin in GUI mode
* minor fix to the way files are handled; due to some versions of hugin_executor not writing out to the path specified in --prefix, assume we are working in the tmp directory and then move the file afterwards


achienbsi:
found some issues

PR #111, opened 2018-02-26
hugin plugin improvements

achienbsi:
* added preference requiring user to specify the paths for hugin, hugin_executor and pto_gen needed for this plguin
* added preference to allow user to choose to launch hugin in GUI mode
* minor fix to the way files are handled; due to some versions of hugin_executor not writing out to the path specified in --prefix, assume we are working in the tmp directory and then move the file afterwards

### DIFF
--- a/contrib/hugin.lua
+++ b/contrib/hugin.lua
@@ -29,7 +29,7 @@ ADDITIONAL SOFTWARE NEEDED FOR THIS SCRIPT
 
 USAGE
 * require this file from your main luarc config file
-* set the hugin tool paths in preferences
+* set the hugin tool paths (on some platforms)
 * if hugin gui mode is used, save the final result in the tmp directory with the first file name and _pano as suffix for the image to be automatically imported to DT afterwards
 
 This plugin will add a new storage option and calls hugin after export.
@@ -39,7 +39,13 @@ local dt = require "darktable"
 local df = require "lib/dtutils.file"
 require "official/yield"
 local gettext = dt.gettext
+
 local namespace = 'module_hugin'
+local user_pref_str = 'prefer_gui'
+local user_prefer_gui = dt.preferences.read(namespace, user_pref_str, "bool")
+local hugin_widget = nil
+local exec_widget = nil
+local executable_table = {"hugin", "hugin_executor", "pto_gen"}
 
 -- works with darktable API version from 2.0.0 to 5.0.0
 dt.configuration.check_version(...,{2,0,0},{3,0,0},{4,0,0},{5,0,0})
@@ -51,9 +57,14 @@ local function _(msgid)
   return gettext.dgettext("hugin", msgid)
 end
 
+local function user_preference_changed(widget)
+  user_prefer_gui = widget.value
+  dt.preferences.write(namespace, user_pref_str, "bool", user_prefer_gui)
+end
+
 local function show_status(storage, image, format, filename,
   number, total, high_quality, extra_data)
-  dt.print("exporting to Hugin: "..tostring(number).."/"..tostring(total))
+  dt.print("exporting to hugin: "..tostring(number).."/"..tostring(total))
 end
 
 local function create_panorama(storage, image_table, extra_data) --finalize
@@ -62,19 +73,21 @@ local function create_panorama(storage, image_table, extra_data) --finalize
 -- We need pto_gen to create pto file for hugin_executor
 -- http://hugin.sourceforge.net/docs/manual/Pto_gen.html
 
-  local hugin = '"'..dt.preferences.read(namespace, "hugin", "file")..'"'
-  local hugin_executor = '"'..dt.preferences.read(namespace, "hugin_executor", "file")..'"'
-  local pto_gen = '"'..dt.preferences.read(namespace, "pto_gen", "file")..'"'
-  local user_prefer_gui = dt.preferences.read(namespace, "hugin_prefer_gui", "bool")
-
-  local cmd_line_available = false
-  if df.check_if_bin_exists(hugin_executor) and df.check_if_bin_exists(pto_gen) then
-    cmd_line_available = true
-  end
+  local hugin = df.check_if_bin_exists("hugin")
+  local hugin_executor = df.check_if_bin_exists("hugin_executor")
+  local pto_gen = df.check_if_bin_exists("pto_gen")
 
   local gui_available = false
-  if df.check_if_bin_exists(hugin) then
+  if hugin then
     gui_available = true
+  else
+    dt.print(_("hugin is not found, did you set the path?"))
+    return
+  end
+
+  local cmd_line_available = false
+  if hugin_executor and pto_gen then
+    cmd_line_available = true
   end
 
   -- list of exported images
@@ -93,6 +106,11 @@ local function create_panorama(storage, image_table, extra_data) --finalize
   local first_file
   for _, k in ipairs(img_set) do
     first_file = k break
+  end
+
+  if first_file == nil then
+    dt.print("no file selected")
+    return
   end
 
   local pto_path = dt.configuration.tmp_dir..'/project.pto'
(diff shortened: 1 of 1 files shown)


wpferguson:
I've been brainstorming for several days trying to figure out a solution that will fit all the operating systems as well as be standard across scripts so that it's easy to use and works with check_if_bin_exists.  In order to implement it, I'm thinking of adding some library routines on the lines of those suggested by @supertobi.  This would also give us a standard naming convention for executable paths so that another script could run a check and use an existing preference instead of having multiple preferences all referencing the same executable. In addition, I'm thinking of adding a gui element to get the preference, rather than requiring the user to manually edit preferences.  

As coded, this implementation would require linux/unix users to edit preferences and specify paths to executables in order to use the script, which would be a step backward for linux/unix users.

I'll try and code up a set of routines for the path preferences this week (maybe tonight).  I can test with windows and and linux, but we really need someone with macOS to help with testing.

wpferguson:
@achienbsi I'll look at that too, and include it in the changes.

wpferguson:
Coded up a solution for a single executable.  Still need to handle multiple executables, add it to the libraries, then test on windows.

wpferguson:
[lua.zip](https://github.com/darktable-org/lua-scripts/files/1771891/lua.zip)

@achienbsi here is a working implementation.  I tested gimp.lua, but didn't have anything to test hugin.lua with.  The only files changed in this are contrib/gimp.lua, contrib/hugin.lua, and lib/dtutils/file.lua.  Try it and let me know what you think.

Bill

achienbsi:
@wpferguson thanks for that. I've only tested with hugin on Windows, here are a few thoughts:

1. I'm not sure why it's necessary to check for linux platform for plugin's widget registration. Wouldn't it be perfectly valid for a user to want to choose his/her locally built version of the executable?

2. It's nice to have executable widget registered automatically, but does that also means the plugin cannot have other widget registered? For example, I wanted to add a check box for user to choose whether to always launch Hugin in GUI mode vs having it automatically done in command line.

3. check_if_bin_exists seems redundant. Each plugin would need to call check_if_bin_exists and then call get_executable_path_preference to get the actual path. From plugin's perspective, it seems simpler to just call get_executable_path_preference and then check_if_file_exists to do the same thing. Maybe check_if_bin_exists should return the search result plus the path where it is found?

4. nitpick, DT's UX don't use capital letters and the widget registered is breaking that

wpferguson:
@achienbsi I'll answer in order

1. On linux locally built executables are usually put in /usr/local/bin which comes before /usr/bin in the path, so which finds the locally built executable first.

2. I coded it that way since there were no other widgets used by gimp or hugin.  If you wanted to use multiple widgets, you would create a box widget to hold them all.

widget1 = df.executable_path_widget....
widget2 = dt.new_widget...
widget3 = dt.new_widget....

widgets = {widget1, widget2, widget3}

hugin_widget = dt.new_widget("box"){
  orientation = "vertical",
  table.unpack(widgets)
}

3. linux doesn't use the executable_path_preferences, or need them (except in the case of wine executables).  This way doesn't break anything that already exists.  This way the linux execution path doesn't change, and we've added accommodations for windows and macOS without having to build os specific implementations.  I'll revisit it and see if there is a cleaner way.

4.  My english teachers did their job too well.  :-D  I'll fix that.

achienbsi:
@wpferguson The only reason I brought up #1 & #3 is because I think it's preferable for plugins not having to worry about target platform. So if other platforms need to use executable_path_preferences, it would be better if doing so on linux would just return the path given by which command?

supertobi:
Hi, I have a cold at the moment and don't have the power to help here. But I think you two will find a good solution. 👍

wpferguson:
@supertobi hope you start feeling better soon.  

@achienbsi I've thought about changing the behavior of check_if_bin_exists and having it return the executable path.  So, if we got it down to displaying the gui element as the only os aware part, would that work?  I selfishly don't want to give up the screen real estate for an option that I don't need and wont use.  And, at the worst it could confuse a linux/unix user and have them set a value that stops the script from working.

houz:
For what it's worth, I never put stuff into /usr/local/. It always goes to its own subdirectory of /opt/.

achienbsi:
@wpferguson Does DT offer collapse-able widget like the one used by modules? Seems like if it can be hidden in there that makes the most sense to me instead of have another unspoken widget rule to exclude linux platform.

wpferguson:
@houz I have 9 different versions of darktable in /opt.  :-D  I use that a lot for development, and then I add a bash script to start it in ~/bin so that which still works.   

@achienbsi At this time widgets don't collapse.  I thought of that solution too, but for right now it's not on the board.  We can ask and maybe it'll end up in 2.5.  My thought on the widget is that if a script provides a widget that can be changed, then it's meant to be changed.  It doesn't make sense to provide a widget that isn't meant to be changed.  And, we're only talking a couple lines of code, so it's not a huge amount of work.

Script authors are still free to write their scripts however they choose.  You may choose to write a script that only runs on windows.  I might look at it and think it's a really good idea and modify it to run on linux.  That's the beauty of open source.  No author has to write cross platform compatible scripts, but we're providing some tools to ease the work in case they want to.

wpferguson:
[lua.zip](https://github.com/darktable-org/lua-scripts/files/1773366/lua.zip)

@achienbsi Here is a cleaner check_if_bin_exists and a new gimp.lua and hugin.lua reflecting the changes.  Check them out and see what you think.

I looked at a couple of other scripts, slideshowMusic.lua and kml_export.lua.  Both of these scripts use external binaries that aren't available on windows.  So, do we start marking scripts operating system compatible (i.e. do a check up front and if we can't run on that os then just return)?  slideshowMusic.lua uses rhythmbox-client and there is no windows version.  kml_export.lua uses xdg-open and again there is no windows version.  Thoughts?

achienbsi:
@wpferguson The modified file.lua works for me, thanks! I'll update and submit a new hugin.lua later.

I think a check_os call similar to check_version would be a good idea. It'd be less confusing due to user error.

wpferguson:
@achienbsi I'll finish documenting the library routine then do a PR on the library and gimp.lua.  I'll let you add the changes to hugin.lua.  I'll start thinking on check_os.

supertobi:
1. I'm with @achienbsi:
I think a check_os call similar to check_version would be a good idea. 

2. kml_export.lua uses xdg-open
xdg-open opens a file with the standard program. A fast search showes, that there is "open" for MacOS and start for Windows:
https://superuser.com/questions/550138/open-file-with-default-application-from-command-line#550144
https://apple.stackexchange.com/questions/212583/how-to-open-files-via-terminal
I think we should wrap this into a helper function too.

wpferguson:
@supertobi PR submitted for check_os

Should we create lib/dtutils/cross_platform.lua and start populating it with solutions to whatever cross platform problems we run into?

supertobi:
@wpferguson Sound good.

supertobi:
@achienbsi What have you done? All lines are different from the original file.

achienbsi:
@supertobi hmm, weird. I started using visual source code as my editor maybe it's doing something to the line ending? I'll have a look later.

achienbsi:
@supertobi ok, fixed. I know what I did... I copied the file directly from Windows to Linux machine to test and then submitted from Linux machine so it still had Windows line endings.

contrib/hugin.lua
```diff
@@ -38,35 +40,54 @@ local df = require "lib/dtutils.file"
...
 
+local namespace = 'module_hugin'
+local user_pref_str = 'prefer_gui'
+local user_prefer_gui = dt.preferences.read(namespace, user_pref_str, "bool")
+local hugin_widget = nil
+local exec_widget = nil
+local executable_table = {"hugin", "hugin_executor", "pto_gen"}
+
 -- works with darktable API version from 2.0.0 to 5.0.0
 dt.configuration.check_version(...,{2,0,0},{3,0,0},{4,0,0},{5,0,0})
```
wpferguson: Once we make the changes to run on windows, specifically using dt.configuration.running_os or dtutils.check_os(), then the API version that this works with is only 5.0.0+

supertobi:
@achienbsi if you make the changes @wpferguson asked for I can commit this.

supertobi:
@wpferguson can you please do a quick test if this runs on all platforms.

wpferguson:
Ran into a problem on macos.  Hugin can't find the files and won't accept them when I pick them.  I'll debug and see what I can figure out.

Bill

wpferguson:
Works fine on linux.

supertobi:
Ok, then is this version better then the old one and there is at least one person waiting for the windows support. I'll merge this pr. Please make a new pr for the MacOS fix.

PR #112, opened 2018-03-02
Windows and macOS external executable compatibility

wpferguson:
Modified lib/dtutils/file.lua 
   Added get_executable_path_preference to retrieve the location of an executable from preferences
   Added set_executable_path_preference to store the local of an executable to a preference
   Added executable_path_widget to create a widget for the user to identify the location of executables
   Modified check_if_bin_exists to be aware of the path preferences and use them.

Modified gimp.lua to be windows and macOS compatible.

These changes break api compatibility with any api less than 5.0.0, since we are using dt.configuration.running_os().

@supertobi I recommend tagging the last commit that was still api 4.0.0 compatible for users of older versions of darktable.

### DIFF
--- a/contrib/gimp.lua
+++ b/contrib/gimp.lua
@@ -67,109 +67,38 @@ local dt = require "darktable"
 local df = require "lib/dtutils.file"
 require "official/yield"
 local gettext = dt.gettext
+local gimp_widget = nil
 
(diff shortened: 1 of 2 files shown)


wpferguson:
I'll write a script for examples that explains how to write scripts that should run across all the operating systems.

PR #117, opened 2018-05-05
Fix autostyle

marco44:
checkIfBinExists doesn't seem to exist in latest releases anymore

### DIFF
--- a/contrib/autostyle.lua
+++ b/contrib/autostyle.lua
@@ -37,7 +37,7 @@ GPLv2
 ]]
 
 local darktable = require "darktable"
-
+local filelib = require "lib/dtutils.file"
 
 -- Forward declare the functions
 local autostyle_apply_one_image,autostyle_apply_one_image_event,autostyle_apply,exiftool_attribute,capture
@@ -68,7 +68,7 @@ function autostyle_apply_one_image (image)
 	  darktable.print("style name not found in " .. darktable.preferences.read("autostyle","exif_tag","string"))
 	  return
   end
-  if not checkIfBinExists("exiftool") then
+  if not filelib.check_if_bin_exists("exiftool") then
      return
   end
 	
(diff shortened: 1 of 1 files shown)


contrib/autostyle.lua
```diff
@@ -87,15 +87,17 @@ function autostyle_apply_one_image (image)
...
   local ok,auto_dr_attr= pcall(exiftool_attribute,image.path .. '/' .. image.filename,tag)
+  --darktable.print_error("dr_attr:" .. auto_dr_attr)
   -- If the lookup fails, stop here
   if (not ok) then
+    darktable.print("Couldn't get attribute" )
     return
   end
   if auto_dr_attr==value then
---	  darktable.print("Image " .. image.filename .. ": autostyle automatically applied " .. darktable.preferences.read("autostyle","exif_tag","string") )
+	  darktable.print_error("Image " .. image.filename .. ": autostyle automatically applied " .. darktable.preferences.read("autostyle","exif_tag","string") )
```
wpferguson: Changing darktable.print to darktable.print_error takes feedback away from the user and puts it in the debugging stream, which the user isn't likely to look at unless there is a problem and they are know to look there.

contrib/autostyle.lua
```diff
@@ -87,15 +87,17 @@ function autostyle_apply_one_image (image)
...
 	  darktable.styles.apply(style,image)
---  else
---	  darktable.print("Image " .. image.filename .. ": autostyle not applied, exif tag " .. darktable.preferences.read("autostyle","exif_tag","string")  .. " not matched: " .. auto_dr_attr)
+  else
+	  darktable.print_error("Image " .. image.filename .. ": autostyle not applied, exif tag " .. darktable.preferences.read("autostyle","exif_tag","string")  .. " not matched: " .. auto_dr_attr)
```
wpferguson: Same as above

wpferguson:
I'm assuming that you use autostyle.lua because you caught the CheckIfBinExists/check_if_bin_exists problem.  I noticed that you moved the darktable.print messages to darktable.print_error messages which only prints them in the debug log.  Was darktable.print causing too many messages to the screen?  If this is the case, maybe keeping track of how many images have the autostyle applied, then using darktable.print to print a summary message would give the user feedback that the styles were applied.  Also, with API 5.0.0 we can now use darktable.print_log for informative messages and darktable.print_error for error messages, which might help remove some confusion when people are reading the debug messages.

marco44:
I changed a bit the way to display messages. Most of it will go to print, except details such as why an image hasn't received a style. What I would have liked is to be able to display several messages in a row… while here, the latest print overwrites the previous one, so if I send several print in a row, only the last one is displayed

wpferguson:
I read through the autostyle.lua code.  I must have overlooked the author line...  *blush*

As far as printing several messages...  If you embed newlines in the message you can print several lines.  It doesn't look good since darktable.print only expects one line and sizes the box to contain it.  I'll look for the code and see if I can adapt it to contain multiple lines.

PR #119, opened 2018-05-15
Replace placeholder for dt.print_log with dt.print_log

wpferguson:
lib/dtutils/log.lua was written with the knowledge that the darktable.print_log function was being developed, so a placeholder was inserted in the code.  Now that the function exists the placeholder is being replaced with the function.


supertobi:
That looks OK, any reason you have not pushed it?

wpferguson:
I was letting it set overnight just to see if anyone had a comment.

PR #120, opened 2018-05-15
hugin fails on images with spaces in the filenames

wpferguson:
single quoted filename arguments to the hugin commands for the supplied image files.  Single quoted any filenames derived from the supplied image files in case they contained spaces.  Added logging at various levels so that the next time I have to work on this I can figure it out quicker.  Normal logging level is log.error which only prints error information.  Changing log level to log.info provides much more insight about what is occurring.  Fixed gui not returning generated file to film strip.


supertobi:
That looks fine, please pull.

PR #130, opened 2018-07-01
Add enfuse_pro_2

BzKevin:
Adds enfuse_pro_2 plug-in to "contrib". This is a version of Holger Klemm's well-known script that adds features as well as cross-platform compatibility.


BzKevin:
I am not very familiar with Github, if there is a better way to do this other than making the changes on my own fork and then requesting my fork be pulled in please let me know. Additionally, I was told by Bill Ferguson it was best to add a single script at a time, hence what you see above where I added 3, then deleted all three and re-added just one specifically. If there is a better way to do that also let me know.

supertobi:
1. @houz perhaps you can answer on the best practice git question. I would suggest a branch for every new commit (in your own fork), and then pull it.
2. @wpferguson is right. Separate and small commits are easier to review and
3. here are some anti closed source people around, so the PS script will not be accepted.

Now to the script. Wow that beast is big and @wpferguson any help with reviewing is welcome.
There are some strings marked as translatable. If you have the translations files you should push them too.
Some strings are not marked as translatable, is that intended?
Unused code should be removed and not commented out.

And I'm not sure about the name "enfuse pro 2". The "2" should be remove, because we don't have a "1" and even then versioning should be done with the version control system (git) and not by filenames. And the "pro" sounds a little bit like marketing and we don't have a not "pro" version or is there a "light/beginner" version too? @BzKevin What do you think?

wpferguson:
@supertobi I'll start looking at it tonight.  

If we trim the name down to just enfuse, then we'll introduce some confusion between official/enfuse and contrib/enfuse.  official/enfuse supports enfuse 4.1 while this script requires 4.2.  official/enfuse may run on 4.2, but I haven't tested it.  I'm on Ubuntu 16.04, so my enfuse is 4.1.4.  I'll download 4.2 and test with it to see what happens.

BzKevin:
@supertobi I don't have any strong feelings about the name, it can be changed to whatever you all feel is best. I only did enfuse_pro_2 because this follows from Holger Klemm's original enfuse_pro script (which was never added to this repo. Perhaps enfuse_advanced would be a good alternative to differentiate it from the more basic official/enfuse (as this does offer a lot more control over the enfuse process from directly within dt)?

In regards to translatable or not: I believe the original script by Holger Klemm did support translations. Quite frankly, when I started working on this script I was confused by what all that was, so I think I ripped some of it out not knowing what it was. I still don't really know how the translation works to be honest, I have not really paid a lot of attention to it. And you are correct, there is now probably a jumble. I do not have any translations, and I do not think the original translations would work anymore anyways.

Sure, commented code can be removed. 

Bill, you may notice in my build_execute_command() function I had to add in remove "open -W -a" from the MacOS command as it was causing the script to fail, oddly. After adding in the removal I have verified it works on both my wife's mac as well as user Hevii_Guy at pixls.us.

wpferguson:
@BzKevin when you run the command on the mac, what does it look like?  Is it just the path to the executable without any additional information?

I loaded up a copy of the script and I'm not having much luck with it.  It keeps core dumping on align_image_stack, but if I do it by hand it works.

contrib/enfuse_pro_2.lua
```diff
@@ -0,0 +1,1050 @@
...
+	exiftool_path = df.check_if_bin_exists("exiftool")
+	if not exiftool_path then
+		dt.print_error("exiftool not found")
+		dt.print("ERROR - exiftool not found")
+		remove_temp_files()
+		return
+	end
+	
+--Ensure proper version of Enfuse installed--
+	--[[
```
wpferguson: This script requires enfuse 4.2.  If you try running it against 4.1 it crashes on the options.  If you're running CentOS, or any Ubuntu LTS except the latest one, then you have version 4.1 by default.  To get around windows not having a grep command you could use find, or you could run enfuse --version in an io.popen() and grab the result, then loop through the lines and use string.match to check for enfuse 4.2

supertobi:
I had forgotten the official/enfuse script.  Is the official/enfuse script still needed if we have this script?
@houz What do you think, you wrote official/enfuse?

wpferguson:
@supertobi Ubuntu 16.04 is supported until 2021 and CentOS 7 is supported until 2024.  Both only have enfuse 4.1.  Users would have to download, compile and install enfuse 4.2 on their own.  So there are potential users for official/enfuse until 2024.  My vote is keep official/enfuse and call this one contrib/enfuse42.

wpferguson:
I got align_image_stack working finally and confirmed that this script does not work with enfuse 4.1.  

@BzKevin the script chokes on filenames with spaces, so you should quote them.

contrib/enfuse_pro_2.lua
```diff
@@ -0,0 +1,1050 @@
...
+ADDITIONAL SOFTWARE NEEDED FOR THIS SCRIPT
+* align_image_stack
+* enfuse
+* exiftool
+
+USAGE
+* require this file from your main luarc config file.
+
+This plugin will add the new export modul "fusion to DRI or DFF image".
+]]
```
wpferguson: I would add that the first time you run darktable with it enabled you need to go to the lua options under preferences and set the executable paths, then restart darktable.

contrib/enfuse_pro_2.lua
```diff
@@ -0,0 +1,1050 @@
...
+	local extension = string.match(full_path, "%.%w*")
+    return path, filename, identifier, extension
+end
+local function truncate(x)
+      return x<0 and math.ceil(x) or math.floor(x)
+end 
+local function replace_comma_to_dot(s)
+	return string.gsub(s, "%,", ".")
+end
+local function file_exists(name)
```
wpferguson: You could just df.check_if_file_exists() and save yourself some code

contrib/enfuse_pro_2.lua
```diff
@@ -0,0 +1,1050 @@
...
+
+This plugin will add the new export modul "fusion to DRI or DFF image".
+]]
+
+local dt = require "darktable"
+local df = require "lib/dtutils.file"
+local gettext = dt.gettext
+
+-- works with LUA API version 4.0.0 and 5.0.0
+dt.configuration.check_version(...,{4,0,0},{5,0,0})
```
wpferguson: You're using dt.print_log() so the compatible api version is only 5.0.0

contrib/enfuse_pro_2.lua
```diff
@@ -0,0 +1,1050 @@
...
+	dt.print_error("rm "..images_to_align)
+	result_to_align_images=dt.control.execute("rm "..images_to_align)
+	dt.print_error("rm "..tmppath.."/aligned_*.tif")
+	result_aligned_images=dt.control.execute("rm "..tmppath.."/aligned_*.tif")
+	]]
+end
+local function build_execute_command(cmd, args, file_list)
+	local result = false
+
+	if dt.configuration.running_os == "macos" then
```
wpferguson: I was kind of afraid something like this would happen when I changed check_if_bin_exists.  I should probably change check_if_bin_exists to not add the open -W -a on MacOS and let each individual script add it if it's necessary since it doesn't work in every instance.

contrib/enfuse_pro_2.lua
```diff
@@ -0,0 +1,1050 @@
...
+AIS_path = df.get_executable_path_preference("align_image_stack")
+if not AIS_path then 
+	AIS_path = ""
+end
+AIS_path_widget = dt.new_widget("file_chooser_button"){
+	title = "Select align_image_stack[.exe] file",
+	value = AIS_path,
+	is_directory = false,
+}
+dt.preferences.register("module_enfuse_pro", "ais_path",	-- name
```
wpferguson: if you change this to dt.preferences.register("executable_paths", "align_image_stack", then you can get away from the workarounds that you were doing.  This fits in with [get|set]_executable_path_preference, works with check_if_bin_exists, and leverages the other scripts that are storing executable path information.

contrib/enfuse_pro_2.lua
```diff
@@ -0,0 +1,1050 @@
...
+enfuse_path = df.get_executable_path_preference("enfuse")
+if not enfuse_path then 
+	enfuse_path = ""
+end
+enfuse_path_widget = dt.new_widget("file_chooser_button"){
+	title = "Select enfuse[.exe] file",
+	value = enfuse_path,
+	is_directory = false,
+}
+dt.preferences.register("module_enfuse_pro", "enfuse_path",	-- name
```
wpferguson: Same as above - dt.preferences.register("executable_paths", "enfuse",

contrib/enfuse_pro_2.lua
```diff
@@ -0,0 +1,1050 @@
...
+exiftool_path = df.get_executable_path_preference("exiftool")
+if not exiftool_path then 
+	exiftool_path = ""
+end
+exiftool_path_widget = dt.new_widget("file_chooser_button"){
+	title = "Select exiftool[.exe] file",
+	value = exiftool_path,
+	is_directory = false,
+}
+dt.preferences.register("module_enfuse_pro", "exiftool_path",	-- name
```
wpferguson: Same as above  dt.preferences.register("executable_paths", "exiftool",

BzKevin:
Thanks for the review, sorry for my absence, I hope to get on some of these this weekend!

contrib/enfuse_pro_2.lua
BzKevin: Yea, I think I have that on my repo page, but adding it in the comments of hte file is defenitely the best way to go, you're right.

contrib/enfuse_pro_2.lua
BzKevin: Didn't know that, great! Will remove 4.0.0 from the list.

contrib/enfuse_pro_2.lua
BzKevin: Yea I was surprised to find this actually. You were the one who originally clue me in to needing the open -W -a, and reading literiture online it seems like that would be the proper way to do it, but I get alignimagestack doesn't count as an actual application, but more an executable...?

Also, when removing that from the command, I was not able to succesfully do string.gsub("open -W -a","") and I don't know why. It wouldn't replace. That's why I had to break it up into three individual statements. Any idea why that might be?

contrib/enfuse_pro_2.lua
BzKevin: Also, in regards to the spaces, this was a big trick for me. I found that if I quoted that paths in the arguments than windows would choke and not run AIS, enfuse, exiftool, etc properly. That's why I didn't quote them. I could do an if statement and only quote on mac and linux machines. Windows users would still have the limitation though. I will try again and see if for some reason quoting works under windows this time around, but that may be a limitation of this program for win. users, unless someone else knows a way around?

Additionally, I would love if there was a way to make the cmd windows that pop up on windows hidden or not have focus... I've asked around previously and didn't get much in response regarding that so I gave up.

contrib/enfuse_pro_2.lua
BzKevin: I just replaced the only location the file_exists() function was called with df.create_unique_filename, so this has been removed entirely.

contrib/enfuse_pro_2.lua
```diff
@@ -0,0 +1,1050 @@
...
+	--Set output path and add filename
+	if (chkbtn_source_location.value) then
+		cmd_output_path = source_path
+	else
+	cmd_output_path=file_chooser_button_path.value --FUTURE FEATURE: add gui option for "same as source images"
+	end
+	path_with_filename = cmd_output_path..os_path_seperator..first_filename.."-"..last_id.."."..cmd_suffix_output_format
+	
+	--Create unique name with index if GUI selection create unique name (don't overwrite), also do this if user selected to make variants from presets
+	if (cmbx_existing_file.selected == 1) or (from_preset) then
```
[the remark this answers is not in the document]
BzKevin: You're right, works like a dream.

contrib/enfuse_pro_2.lua
wpferguson: I originally did this because Mac users were trying to run GIMP and when GIMP opened, the script finished running instead of waiting for GIMP to finish.  The commands you're running don't have a GUI, but don't run with "open -W -a".  So it seems to me the best choice is for me to take it out of the library and each script writer needs to be aware that on Macs they need to do "open -W -a" if they want to run a GUI command.  I'll make that change and submit a PR  To get gsub to work you needed to escape the spaces and dashes, i.e. string.gsub(mac_command, "open% %-W% %-a ", "").

contrib/enfuse_pro_2.lua
wpferguson: I meant quote the filenames of the image files, i.e. /usr/bin/align_image_stack '/path/to/file name1.cr2' '/path/to/file name2.cr2' '/path/to/file name3.cr2'.  If you don't quote this example align_image_stack sees the files as 
/path/to/file
name1.cr2
/path/to/file
name2.cr2
/path/to/file
name3.cr2

contrib/enfuse_pro_2.lua
wpferguson: I made the change and submitted a PR.  I'll leave it overnight for comment, then merge it in the morning.

contrib/enfuse_pro_2.lua
BzKevin: Yea, the behavior is really odd on windows. Maybe you can help me solve what I am seeing. Here it is, with an example command and two scenarios below:
CMD = "C:\Program Files\Hugin\bin\align_image_stack.exe"  -i -C --distortion -g 5 -c 8 -t 3 --corr=0.9 --gpu -a "C:\Users\Kevin\AppData\Local\Temp\aligned_"  "C:\Users\Kevin\AppData\Local\Temp\_MG_5798.tif" "C:\Users\Kevin\AppData\Local\Temp\_MG_5800.tif" "C:\Users\Kevin\AppData\Local\Temp\_MG_5797.tif" "C:\Users\Kevin\AppData\Local\Temp\_MG_5799.tif"

NOTE: ALL paths are now quoted in the command including the executable path, output path, and source file paths.

1) Copy and paste the command into a cmd window and press enter, things run great, no complaints.
2) Modify enfuse_pro_2 script to build its command identical to above and it fails with the following error:
'C:\Program' is not recognized as an internal or external command, operable program or batch file.

contrib/enfuse_pro_2.lua
BzKevin: Great, thanks for the update and explanation about gsub. I should have realized that was my problem...

BzKevin:
I made mode of the changes per your suggestions. Notable differences are:
1) I did not get around to adding code that checks the enfuse version for 4.2 only
2) I only quote the paths in run commands on Linux and MacOS, due to issues I expereince (as explained above) on Windows. If you know a way around this I would love to hear it.

wpferguson:
I'll pull a new copy and look at it.  When you quote things use single quotes (') instead of double quotes (").  Double quotes allow the command executor to interpret the arguments, single quotes force the argument to be used as is.  For instance

echo "$HOME" prints out your home directory and echo '$HOME' prints out $HOME

contrib/enfuse_pro_2.lua
wpferguson: I merged the change to check_if_bin_exists() so you can take out the lines getting rid of "open -W -a"

contrib/enfuse_pro_2.lua
```diff
@@ -0,0 +1,1026 @@
...
+--Create Images String--
+	images_to_align = ""
+	images_to_blend = ""
+	counted_images=0
+	first_id = "999999999999999999999999"
+	last_id = "0"
+	for source_image,image_path in pairs(image_table) do
+		counted_images=counted_images+1
+		if dt.configuration.running_os == "windows" then
+			images_to_align = images_to_align..image_path..' '
```
wpferguson: The windows version of align_image_stack will  not accept a filename with spaces in it, no matter how you quote it or escape it.  The linux version accepts filenames with spaces as long as they're wrapped in single quotes, so I think this is probably a bug in the windows implementation.

Your can scan the filename for spaces and then....

1. print an error to the screen about filenames not having spaces and exit

or 

2.  rename the exported files to something without spaces and modify the argument list to reflect that

or 

3.  leave it as is and report a bug to the hugin people and hope the fix comes out before somebody trips over the problem :-)

or 

4. leave it as is

contrib/enfuse_pro_2.lua
```diff
@@ -0,0 +1,1026 @@
...
+		else
+			images_to_align = images_to_align..'"'..image_path..'" '
```
wpferguson: Turn your quotes around, i.e. "'" and "' "

contrib/enfuse_pro_2.lua
BzKevin: I'll probably go with #3 and then put a note in the readme / header warning windows users of this known issue.

wpferguson: I checked this a little further and discovered it's not a bug with align_image_stack.  You can pass filenames with spaces, on windows, as long as they are surrounded with double quotes.  You can't, however, pass a double quoted executable path and a double quoted filename to dt.control.execute() because it fails with an error.  So, it appears that, until we find a work around, that you can't open filenames with spaces in external executables on windows.

supertobi:
What is the status here?

wpferguson:
Waiting on me to get back to @BzKevin.  I should have time this weekend.

BzKevin:
There's a few things that need to be added for proper function when there
are spaces in the file paths. Bill created some new functions and added
them to file utils, I need to incorporate them. Also need to work with bill
to implement the preferences in such a way that they can be used by other
scripts he is working on at the moment as well.

Have had a lot going on lately. Hopefully can get around to things this
week and next!

cryptomilk:
It looks like that some rm command tries to work on a directory ...

```sh: /home/asn: Is a directory
rm: missing operand
Try 'rm --help' for more information.```

**Comment by wpferguson** [comment]
Could you run it with -d lua, and paste the output?

**Comment by cryptomilk** [comment]
Looks like it is gone now. Strange.

**Comment by cryptomilk** [comment]
What I miss is an option to import the picture after it has been created.

**Comment by BzKevin** [comment]
That should still be there under settings>lua options> "enfuse advanced: add fused image to database"
image

**Comment by cryptomilk** [comment]
I missed that option, sorry.

**Comment by cryptomilk** [comment]
Ok, when choosing 32bit float tiffs as output, it fails and I reproduce the "rm" issues, it fails to delete the temporary files.

```
1,077539 LUA enfuseAdvanced - Executable Path Preference: /usr/bin/align_image_stack
1,083352 LUA enfuseAdvanced - Executable Path Preference: /usr/bin/enfuse
1,087852 LUA enfuseAdvanced - Executable Path Preference: /usr/bin/exiftool
64,944886 LUA enfuseAdvanced - Starting Image Fusion
64,950030 LUA enfuseAdvanced - Align Start Command: "/usr/bin/align_image_stack"  -i -C --distortion -g 5 -c 8 -t 3 --corr=0.9 --gpu -a '/tmp/aligned_'  '/tmp/2018-07-25__7M33733.tif' '/tmp/2018-07-25__7M33734.tif' '/tmp/2018-07-25__7M33731.tif' '/tmp/2018-07-25__7M33735.tif' '/tmp/2018-07-25__7M33732.tif' 
align_image_stack: using graphics card: X.Org AMD Radeon (TM) RX 470 Graphics (POLARIS10, DRM 3.26.0, 4.18.15-1-default, LLVM 6.0.1)
Invalid focallength
        No valid distortion data found in lens database.
Invalid focallength
        No valid distortion data found in lens database.
Invalid focallength
        No valid distortion data found in lens database.
Invalid focallength
        No valid distortion data found in lens database.
Invalid focallength
        No valid distortion data found in lens database.
ERROR: 16:18:41.027760 (/home/abuild/rpmbuild/BUILD/hugin-2018.0.0/src/hugin_base/vigra_ext/Correlation.h:404) subpixelMaxima(): Bad polynomial fit results
ERROR: 16:18:41.028021 (/home/abuild/rpmbuild/BUILD/hugin-2018.0.0/src/hugin_base/vigra_ext/Correlation.h:404) subpixelMaxima(): Bad polynomial fit results
terminate called after throwing an instance of 'vigra::PreconditionViolation'
  what():  
Precondition violation!
correlateImageFFT(): kernel larger than image.
(/home/abuild/rpmbuild/BUILD/hugin-2018.0.0/src/hugin_base/vigra_ext/Correlation.h:105)

68,005921 LUA enfuseAdvanced - Completed Align
68,006034 LUA ERROR Align Image Stack Failed
68,006042 LUA enfuseAdvanced - Deleting temp files...
rm: missing operand
Try 'rm --help' for more information.
68,079629 LUA enfuseAdvanced - Done deleting
```

cryptomilk:
Looks like "align_image_stack" doesn't support 32bit tiffs.

wpferguson:
@supertobi I say we merge this and open an issue to make align_image_stack optional.  @BzKevin can do it if he wants, or I'll take it.

BzKevin:
Thanks Bill! Sorry I never got back to you. Appreciate you merging this in.

BzKevin:
cryptomilk, I agree, an option to bypass AIS would be a worthwhile addition. 

There is already an option to bypass distortion correction, it's a checkbox under "align image options". 

Improving the in-program help wouldn't be bad, until that happens though you can always refer to the Hugin documentation if you have questions, this plug-in does nothing more than run the existing programs with command line arguments based on your selections; that is to say that the functionality and computations are identical to as if you were running it through Hugin so the help files should correlate close to 1:1.
http://enblend.sourceforge.net/enfuse.doc/enfuse_4.2.xhtml/enfuse.html
https://wiki.panotools.org/Align_image_stack

PR #132, opened 2018-07-11
Cross platform external command capability

wpferguson:
Added system library to handle running external commands on different operating systems in a transparent manner.  Added file and path quoting that is cross platform compatible.  Changed gimp.lua to use the new utilities.

### DIFF
--- a/contrib/gimp.lua
+++ b/contrib/gimp.lua
@@ -95,17 +95,7 @@ local function group_if_not_member(img, new_img)
     dt.print_log("Added to group")
   end
 end
---[[
-local function sanitize_filename(filepath)
-  local path = df.get_path(filepath)
-  local basename = df.get_basename(filepath)
-  local filetype = df.get_filetype(filepath)
 
-  local sanitized = string.gsub(basename, " ", "\\ ")
-
-  return path .. sanitized .. "." .. filetype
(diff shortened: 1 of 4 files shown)


contrib/gimp.lua
```diff
@@ -104,7 +105,7 @@ local function sanitize_filename(filepath)
 
   return path .. sanitized .. "." .. filetype
 end
-
+]]
```
supertobi: Please delete the unused code.

supertobi:
I don't have MacOs or Windows for testing this. I put my trust in you that it works.
Perhaps you can add an example script to show how to write a multi OS script.

contrib/gimp.lua
wpferguson: Thanks

supertobi:
> Removed leftover unused code 

Wow, that was fast. 👍

wpferguson:
I have an idea for an example script.  Lots of comments with a little bit of code :-) and it will actually do something useful.

wpferguson:
@supertobi , just tested a new script, written with these changes, on linux, windows and macos without any problems.  I think it's ready to go.

supertobi:
Then push it.

PR #133, opened 2018-08-03
corrected "dt" to "darktable" in line 63 of cr2hdr

sjjh:
object "darktable" is defined in line 35; usage of "dt" was probably a typo or left over from a refactoring.


wpferguson:
Thanks for the catch.  I guess I need to load magic lantern on my camera and generate some dual iso images for testing.

PR #134, opened 2018-08-08
Multiple OS Example

wpferguson:
Created a cross platform script that calls an external executable.  It also performs a service that some users might find useful, extracting the embedded jpeg image from a raw file, then importing and grouping it with the raw.  Updated README.md to reflect the new file


supertobi:
Nice example nice one, please push.

PR #135, opened 2018-08-09
Image Stacking

wpferguson:
Created image_stack.lua based on Pat David's blog entry, https://patdavid.net/2013/05/noise-removal-in-photos-with-median_6.html.  Updated README.md to reflect the new script.

This script adds a new export target, image_stack.  A stack of images are selected, exported, optionally aligned, then processed using the specified operator.  Mean and median are the two most common, but others are supplied.  The resulting image is imported back into the collection.  

An option exists to tag the source images used to create the stacked result.  Images tagged in this manner can be collected using collect images and then used again to create another result.

This script has been tested on Linux, MacOS, and Windows.

I tested this using the two highest ISO's my camera has, 6400 and 12800.  I shot 10 images at ISO 6400 and 12 at ISO 12800.  I first tested just stacking with no preprocessing and got a 2 - 3 stop improvement in noise.  Some preprocessing (demosaic - amaze w/color smoothing 2, profiled denoise wavelets chroma, profiled denoise wavelets lightness) resulted in pretty much noise free images.  I used both mean and median.  Mean was quick, usually less than 10 seconds to process a stack.  Median took between 1 and 2 minutes to process the same stack.  I tried some other operators and their processing time fell between mean and median.

### DIFF
--- a/contrib/image_stack.lua
+++ b/contrib/image_stack.lua
@@ -1,0 +1,582 @@
+--[[
+
+    image_stack.lua - process a stack of images
+
+    Copyright (C) 2018 Bill Ferguson <wpferguson@gmail.com>.
+    Copyright (C) 2016,2017 Holger Klemm 
+
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 3 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+]]
+--[[
+    image_stack - export a stack of images and process them, returning the result
+
+    This script provides another storage (export target) for darktable.  Selected
+    images are exported in the specified format to temporary storage.  The images are aligned
+    if the user requests it. When the images are ready, imagemagick is launched and uses
+    the selected evaluate-sequence operator to process the images.  The output file is written
+    to a filename representing the imput files in the format specified by the user.  The resulting 
+    image is imported into the film roll.  The source images can be tagged as part of the file 
+    creation so that  a user can later find the contributing images.
+
+    ADDITIONAL SOFTWARE NEEDED FOR THIS SCRIPT
+    * align_image_stack - http://www.hugin.org
+    * imagemagick - http://www.imagemagick.org
+
+    USAGE
+    * require this script from your main lua file
+    * select the images to process with image_stack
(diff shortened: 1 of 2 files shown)


wpferguson:
@supertobi - Thoughts and/or suggestions?

contrib/image_stack.lua
```diff
@@ -0,0 +1,582 @@
...
+    local result = dtsys.external_command(convert_command)
+    if result == 0 then
+      dt.print(_("image stack processed"))
+      cleanup(img_list)
+      job.percent = job.percent + percent_step
+
+      -- import image
+      dt.print(_("importing result"))
+      local film_roll_path = extract_collection_path(image_table)
+      local import_filename = df.create_unique_filename(film_roll_path .. "/" .. df.get_filename(output_filename))
```
supertobi: Does this work on Windows?

contrib/image_stack.lua
```diff
@@ -0,0 +1,582 @@
...
+      df.file_move(output_filename, import_filename)
+      imported_image = dt.database.import(import_filename)
+      local created_tag = dt.tags.create("Created with|image_stack")
+      dt.tags.attach(created_tag, imported_image)
+      for img,_ in pairs(image_table) do
+        copy_image_attributes(img, imported_image)
+        break
```
supertobi: I'm not sure this break is OK or not. But I think so.

contrib/image_stack.lua
```diff
@@ -0,0 +1,582 @@
...
+      end
+      job.percent = job.percent + percent_step
+
+      -- tag images if requested
+
+      if tag_source then
+        dt.print(_("tagging source images"))
+        local source_tag = dt.tags.create("Source file|" .. imported_image.filename)
```
supertobi: Should this be translatable?

contrib/image_stack.lua
```diff
@@ -0,0 +1,582 @@
...
+-- Normally this shouldn't be a problem, but if you use this code just beware.
+-- If you want to do it non recursively, then remove the /s argument from dir
+-- and grab the path component from the search string and prepend it to the files
+-- found.
+
+local function list_files(search_string)
+  local ls = "ls "
+  local files = {}
+  local dir_path = nil
+  local cnt = 1
```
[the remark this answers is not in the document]
wpferguson: Fortran?  All my command line loops start with for i in ...  :-)

contrib/image_stack.lua
wpferguson: It actually does as long as you don't send it to the shell.  But, I should probably do the PS  substitution just to be safe.

contrib/image_stack.lua
wpferguson: I'm just copying the attributes from the first image, since they should be the same for all images except for time

wpferguson: I'll add a comment

contrib/image_stack.lua
wpferguson: Yes

supertobi:
Looks good. Feel free to push it.

PR #137, opened 2018-08-15
Alter enfuse arguments based on version

wpferguson:
Added a version check to modify the arguments based on the version of enfuse.  Enfuse 4.2 no longer accepts the exposure-mu argument.  The closest function is exposure-optimum. For versions of enfuse less than 4.2 exposure-mu is used.  For versions of enfuse greater than 4.2, exposure-optimum is used.


supertobi:
looks good, please push

PR #138, opened 2018-08-17
Manage external executables used by the lua scripts

wpferguson:
Created tools/executable_manager.lua for managing external executables used by the lua scripts.  Updated the README to reflect the new script.


supertobi:
Looks good, please push.

PR #139, opened 2018-08-18
darktable script management interface

wpferguson:
Created tools/script_manager.lua to provide an interface from darktable for managing the lua scripts.  Updated README.md to reflect the added script.

This file is designed to replace the luarc file in the darktable configuration directory.  Alternatively it can be required from the luarc file.  This file is dependent on the libraries and therefore needs the lua-scripts distribution installed.  A standalone version is available at https://github.com/wpferguson/script_manager.

script_manager is cross platform compatible and has been tested on Windows 7, MacOS High Sierra, and Linux (Ubuntu 16.04).

### DIFF
--- a/tools/script_manager.lua
+++ b/tools/script_manager.lua
@@ -247,6 +247,7 @@ local function activate(script, scriptname)
   local status, err = prequire(sm.script_paths[script])
   -- do it
   if status then
+    dt.preferences.write("script_manager", script, "bool", true)
     dt.print("Loaded " .. scriptname)
   else
     dt.print(scriptname .. " failed to load")
--- a/tools/script_manager.lua
+++ b/tools/script_manager.lua
@@ -679,7 +679,7 @@ if dt.configuration.running_os == "windows" or sm.need_git then
 end
 
 if dt.configuration.running_os ~= "windows" then
-  --sm.configuration_widgets[#sm.configuration_widgets + 1] = sm.use_lua_scripts_version
+  sm.configuration_widgets[#sm.configuration_widgets + 1] = sm.use_lua_scripts_version
   sm.configuration_widgets[#sm.configuration_widgets + 1] = sm.link_downloads_directory
 end
 sm.configuration_widgets[#sm.configuration_widgets + 1] = sm.apply_configuration
--- a/tools/script_manager.lua
+++ b/tools/script_manager.lua
@@ -29,10 +29,11 @@
     scripts directory that matched the current repository the categories would be contrib, examples, 
     offical, and tools.  Each script has an Enable/Disable button to enable or disable the script.
 
-    A link is created to the user's Downloads directory on linux, unix and mac os.  Windows users must create the 
+    A link is created to the user's Downloads directory on linux, unix and MacOS.  Windows users must create the 
     link manually using mklink.exe.  Additional "un-official" scripts may be downloaded 
     from other sources and placed in the users Downloads directory.  These scripts all fall in a downloads category.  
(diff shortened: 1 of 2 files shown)


tools/script_manager.lua
```diff
@@ -0,0 +1,764 @@
...
+     dt.configuration.running_os == "windows" then
+    use_lua_scripts_version()
+  end
+
+  return result
+end
+
+local function add_script_data(script_file)
+
+  local pattern = "(.-)/(.-)(([^\\/]-)%.?([^%.\\/]*))$"
```
supertobi: Please add documentation to pattern like this. This looks like write only code. ;)

tools/script_manager.lua
```diff
@@ -0,0 +1,764 @@
...
+            if not string.match(script_file, "yield") then -- special case, because everything needs this
+              add_script_data(script_file)
+            else
+              prequire(script_file) -- load yield.lua
+            end
+          end
+        end
+      end
+    end
+  end
```
supertobi: Please consider using "and" instead of the nested "if". I think that is easier to read and the code would look cleaner.

tools/script_manager.lua
```diff
@@ -0,0 +1,764 @@
...
+  -- work around because we can't dynamically add a new stack child.  We create an empty child that will be
+  -- populated with downloads as they occur.  If there are already downloads then this is just ignored
+
+  add_script_data("downloads" .. PS)
+end
+
+-- get the script documentation, with some assumptions
+local function get_script_doc(script)
+  local description = nil
+  local ps = dt.configuration.running_os == windows and "\\" or "/"
```
supertobi: Cant you use the PS from the top of the file here?

tools/script_manager.lua
```diff
@@ -0,0 +1,764 @@
...
+      for _,sname in ipairs(sm.script_names[cat]) do
+        local req = du.join({cat, sname}, "/")
+        local btext = "Enable "
+        if dt.preferences.read("script_manager", req, "bool") then
+          status, lib = prequire(sm.script_paths[req])
+          if status then 
+            btext = "Disable "
+          else
+            dt.print_error("Error loading " .. sname)
+  --          dt.print_error("Error message: " .. lib)
```
supertobi: If this line is no longer needed please remove it.

tools/script_manager.lua
```diff
@@ -0,0 +1,764 @@
...
+sm.script_widgets = {}
+sm.script_categories = {}
+sm.script_names = {}
+sm.script_paths = {}
+sm.main_menu_choices = {}
+sm.main_stack_items = {}
+
+-- figure out the path separator
+
+local ps = dt.configuration.running_os == "windows" and "\\" or "/"
```
supertobi: Can you use PS from line 60 here?

supertobi:
Please push it, I haven't seen any big problems in the script.
Perhaps it would be good to make the strings translatable.

tools/script_manager.lua
wpferguson: I like the and better too

tools/script_manager.lua
wpferguson: The second or third time I did the local ps =... I figured out that it should be a global, but I forgot to go back and clean up the other instances.

tools/script_manager.lua
wpferguson: I think this pattern came from you originally.  :-)

supertobi: That may be, but it is still evil code. ;)

PR #146, opened 2018-09-18
Update German translation of hugin and kml_export

simon04:


supertobi:
Looks good, please add the updated.mo files too.

wpferguson:
See my comment on PR 147.

PR #147, opened 2018-09-18
Delete compiled .mo files

simon04:


supertobi:
I think we should not delete the .mo files until we have a make file (or a script, or something like that) to generate them.

supertobi:
And if we remove the .mo files we will need some kind of release press, because it is then no longer possible to just clone the git repro and use the scripts.

wpferguson:
We probably need to do press releases anyway to let people know what's there.  I could tweak script_manager to automatically regenerate the .mo files on update.

wpferguson:
@supertobi PR 148 has the translation tool, tools/gen_i18n_mo.lua to generate the .mo files and put them in the proper location.  What are your thoughts?

supertobi:
The script is a nice and welcome improvement. But I'm not sure what to do next. :(

wpferguson:
The only script supplying mo files is contrib/select_untagged.lua.  We could just leave this as is for the moment, and merge the translation tool.  Once the translation tool has been used for awhile, we can merge this and let the translation tool take over generating this mo file too.

supertobi:
OK, then lets keep this .mo files for now. I think the best time for removing them is the next dt release Then we could add a note to the release notes.

PR #148, opened 2018-09-27
Translation tool

wpferguson:
Added tool to compile source translation files and place the results in the correct locale directory

If anyone has a better name than gen_mo, I'm open to it. :-)


christte:
gen_i18n_mo? (makes it almost impossible to miss its purpose :-) )

supertobi:
OK, I'm fine with this script.

PR #149, opened 2018-11-03
'LabelsToTags' is standalone.

schwerdf:
The table of scripts in `README.md` has a question mark in the "Standalone" column for my script, `LabelsToTags`. This change substitutes "Yes" and adds an explicit clarification in the script header.


supertobi:
Looks good, merged.

PR #150, opened 2018-11-19
Hugin windows fix

wpferguson:
Hugin doesn't display the files when the GUI opens due to different quoting requirements on windows versus linux/macos.  Changed windows path quoting to double quotes instead of single.  Changed dt.control.execute to dtsys.external_command to handle the windows quoted arguments.


wpferguson:
I tested it on windows and linux (to make sure I didn't break something else)...

PR #153, opened 2019-01-06
enfuse: convert raws to tiff automatically

cigno5:
it uses darktable-cli (if installed) to convert to a temporary file the
raw files that enfuse is not able to process.

### DIFF
--- a/official/enfuse.lua
+++ b/official/enfuse.lua
@@ -49,7 +49,12 @@ end
 -- add a new lib
 -- is enfuse installed?
 local enfuse_installed = df.check_if_bin_exists("enfuse")
-local darktable_cli_installed = df.check_if_bin_exists("darktable-cli")
+
+-- instance of DT tiff exporter
+local tiff_exporter = dt.new_format("tiff")
+tiff_exporter.bpp = 16
+tiff_exporter.max_height = 0
+tiff_exporter.max_width = 0
 
 -- check the version so that we can use the correct arguments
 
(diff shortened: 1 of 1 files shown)


official/enfuse.lua
```diff
@@ -149,6 +150,14 @@ dt.register_lib(
             cnt = cnt + 1
             f:write(i.path.."/"..i.filename.."\n")
             target_dir = i.path
+          elseif i.is_raw and darktable_cli_installed then
+            local tmp_exported = os.tmpname()..".tif"
+            os.execute("darktable-cli \""..i.path.."/"..i.filename.."\" \""..tmp_exported.."\"")
```
supertobi: @wpferguson does this construct with windows?

official/enfuse.lua
```diff
@@ -149,6 +150,14 @@ dt.register_lib(
...
+            dt.print("Raw file "..i.filename.." automatically converted to tiff")
```
supertobi: If you would use "string.format(" here instead of joining the strings with ".." it would make it easier to translate.  And then using "_()" would it even make it possible to translate the string. (Yes I know that the rest of the script is not translatable. But we have to start somewhere.)

https://github.com/darktable-org/lua-scripts/blob/3a5dfbc842c27596fe7cd7b1fa58030380df9eb5/contrib/image_stack.lua#L253

cigno5: Not a problem at all, and it's a good thing to start from there! Shall I fix the string (and I'd also move the `print` before the conversion command) and reissue the pull request?

supertobi: Yes, please.

official/enfuse.lua
cigno5: @wpferguson that's definitely a better approach to mine. I used the same approach used for invoking `enfuse`, and I carefully checked if `darktable-cli` is in the path.

I will change it asap!

wpferguson:
Also

local command = "enfuse --depth...

should be 

local command = enfuse_installed .. " --depth

df.check_if_bin_exists() returns the path to the executable.

wpferguson:
@cigno5 There need to be a few more changes to make enfuse.lua cross platform compatible.  I can tell you the changes and let you make them, or you can push the above changes and I'll copy the file from your tree and make them, then send it back to you.  Or we can merge this and then I'll do the changes as a separate PR.  Thoughts?   @supertobi thoughts?

supertobi:
Push the stuff first and make then a second commit.

cigno5:
@wpferguson I agree, I will fix the `print` statement and the conversion in tiff first, then I'd like to make `enfuse.lua` cross platform compatible (if you can suggest any guidelines you're welcome!)

I do not mind to take care of the "whole package", it's a nice job to contributing, and my knowledge of DT Lua API will improve. If you agree I'll do it!

wpferguson:
Sounds good to me.

wpferguson:
For cross platform lets take care of paths first

somewhere near the top add

-- path separator
local PS = dt.configuration.running_os == "windows" and "\\" or "/"

then go through the code and replace instances of "/" with PS

so for instance 

f:write(i.path.."/"..i.filename.."\n")

would be 

f:write(i.path..PS..i.filename.."\n")

cigno5:
@wpferguson shall I continue on my fork and pushing commits on this PR or it's better to close (and merge) this one and proceed with a new one?

And, second question...why we should make this changes only on `enfuse.lua` script? I think it should be better to include some utility functions in `dtutils.lua`, like a global `PS` var or common functions to write paths or call external program. Does it make sense to you?

wpferguson:
My recommendation is to continue making changes to your branch rather than merging something that only works on some systems.  @supertobi, your thoughts?

There are more libraries than lib/dtutils.lua.  In the lib/dtutils directory there are libraries for debug, file, log, processor, string, and system functions.  lib/dtutils/system.lua includes the functions to launch an external process, i.e.

local dtsys = require "lib/dtutils.system"

then replace dt.control.execute with dtsys.external_command

I'm still working on figuring out the constants and the best way to deal with them.  Maybe a lib/dtutils/constants.lua

supertobi:
>  shall I continue on my fork and pushing commits on this PR

Yes, please do so.

PR #157, opened 2019-01-21
Timelapse plugin based on ffmpeg

theres:
Plugin that register additional storage `timelapse` and create video based on selected images. It's tested only on Linux.


supertobi:
Nice, thanks for the PR.
Could you perhaps rename it to "video_ffmpeg.lua" because we already have a "video_mencoder.lua". That would make the naming a little bit more consistent.
@wpferguson Could you test the script on Windows?
@wpferguson What do you think about moving the functions "open_file", "mkdir" and "rm" to the lib?

contrib/timelapse.lua
```diff
@@ -0,0 +1,463 @@
...
+
+local dt = require "darktable"
+local df = require "lib/dtutils.file"
+local dsys = require "lib/dtutils.system"
+local gettext = dt.gettext
+
+dt.configuration.check_version(...,{5,0,0})
+
+local MODULE_NAME = 'timelapse'
+gettext.bindtextdomain(MODULE_NAME, dt.configuration.config_dir..'/lua/')
```
wpferguson: I would move the OS_PATH_SEPARATOR statement above this, and maybe use PS for the name with a comment specifying what it's for, then code the directory as

dt.configuration.config_dir..PS..'lua'..PS..'locale'..PS

contrib/timelapse.lua
```diff
@@ -0,0 +1,463 @@
...
+
+local function _(msgid)
+  return gettext.dgettext(MODULE_NAME, msgid)
+end
+
+local OS_PATH_SEPARATOR = dt.configuration.running_os == "windows" and  "\\"  or  "/"
```
wpferguson: See above comment

contrib/timelapse.lua
```diff
@@ -0,0 +1,463 @@
...
+table.sort(res_list)
+
+format_list = {}
+for i,v in pairs(formats) do
+  table.insert(format_list, i)
+end
+
+table.sort(format_list)
+
+codec_list = formats['AVI']['codecs']
```
wpferguson: Perhaps a comment explaining why you used this instead of the loop as above.  I figured it out, but a comment would make it easier for someone else

contrib/timelapse.lua
```diff
@@ -0,0 +1,463 @@
...
+
+local ffmpeg_path_widget = df.executable_path_widget({'ffmpeg'})
+dt.preferences.register("executable_paths", "ffmpeg",
+  "file",
+  MODULE_NAME .. ': ffmpeg location',
+  'Install location of ffmpeg',
+  "ffmpeg",
+  ffmpeg_path_widget
+)
+
```
wpferguson: if you use dt.preferences.register then the user has to start darktable, go to settings, change the preference, kill darktable and restart it.

on the other hand, if you do something like

local timelapse_widgets = {}

if not df.check_if_bin_exists("ffmpeg") then
  table.insert(timelapse_widgets, df.executable_path_widget("ffmpeg"))
end

then add your other widgets to timelapse_widgets and just unpack it into the final widget.  That way if you are on linux, ffmpeg is found in the path and nothing happens, otherwise if it's not found the widget is built and placed at the top of the widgets in the exporter.  Once you set the widget the value is available for use so you don't have to restart darktable.

contrib/timelapse.lua
```diff
@@ -0,0 +1,463 @@
...
+local function format(label, symbols)
+  local es1, es2 = "\u{ffe0}", "\u{ffe1}" -- for simplicity, just some strange utf characters 
+  local result = label:gsub("\\{", es1):gsub("\\}", es2)
+  for s,v in pairs(symbols) do
+    result = result:gsub("{"..s.."}", v)
+  end
+  return result:gsub(es1, "{"):gsub(es2, "}")
+end
+
```
wpferguson: I'm not sure I understand this.  Is this a windows issue?  Could you give me an example of what it's trying to fix?

contrib/timelapse.lua
```diff
@@ -0,0 +1,463 @@
...
+local function open_file(filename)
```
wpferguson: could we change this to something like launch_default_app?  open_file makes it sound like you're opening a file for  reading or writing, not launching an application that's associated with the file.

@supertobi this could probably end up in the system library.

contrib/timelapse.lua
```diff
@@ -0,0 +1,463 @@
...
+local function mkdir(path) 
+  local mkdir_cmd = dt.configuration.running_os == 'windows' and 'mkdir' or 'mkdir -p'
+  return dsys.external_command(mkdir_cmd..' '..df.sanitize_filename(path))
+end
+
+local function rm(path)
+  local rm_cmd = dt.configuration.running_os == 'windows' and 'rmdir /S /Q' or 'rm -r'
+  return dsys.external_command(rm_cmd..' '..df.sanitize_filename(path))
+end
+
```
wpferguson: @supertobi - both of these will probably end up in the dtutils.file library

contrib/timelapse.lua
```diff
@@ -0,0 +1,463 @@
...
+local function string_pref_write(name, widget_attribute)
+  widget_attribute = widget_attribute or "value"
+  local writer = function(widget)
+    dt.preferences.write(MODULE_NAME, name, "string", widget[widget_attribute])
+  end
+  return writer
+end
+
```
wpferguson: I like these functions.  Nice!

contrib/timelapse.lua
```diff
@@ -0,0 +1,463 @@
...
+    else
+      dt.print(_('SUCCESS'))
+      if extra_data['open_after_export'] then
+        open_file(path)
+      end
+    end
+
+    rm(tmp_dir)
+end
+
```
wpferguson: You could combine the above 3 functions into a function called process_timelapse and just do everything.  It would simplify your code because you wouldn't need any of the extra_table stuff.  You could just declare variables, set them and use them.  Put nil in place of the export and finalize_export functions and process_timelapse in place of the init_export function.

contrib/timelapse.lua
```diff
@@ -0,0 +1,463 @@
...
+dt.register_storage(
+  'module_timelapse', 
+  _(MODULE_NAME), 
+  nil,
```
wpferguson: I recommend that you steal the show_status function from another of the storage modules, such as contrib/gimp.lua.  If you put show_status here, it gives the user feedback as the export is progressing, so they get a warm fuzzy that something is happening and maybe some idea of how long it will take.  If you're exporting several hundred images it's going to take awhile, so a message on the screen of "Exporting 621/1000" lets the user know it's working.

contrib/timelapse.lua
```diff
@@ -0,0 +1,463 @@
...
+    end
+
+    rm(tmp_dir)
+end
+
+dt.register_storage(
+  'module_timelapse', 
+  _(MODULE_NAME), 
+  nil, 
+  finalize_export,
```
wpferguson: put process_timelapse here

contrib/timelapse.lua
```diff
@@ -0,0 +1,463 @@
...
+  support_format,
```
wpferguson: nil

contrib/timelapse.lua
```diff
@@ -0,0 +1,463 @@
...
+    rm(tmp_dir)
+end
+
+dt.register_storage(
+  'module_timelapse', 
+  _(MODULE_NAME), 
+  nil, 
+  finalize_export,
+  support_format, 
+  init_export,
```
wpferguson: nil

contrib/timelapse.lua
```diff
@@ -0,0 +1,463 @@
...
+  output_box,
+  open_after_export_cb
+}
+
+---- EXPORT & REGISTRATION
+
+local function support_format(storage, format)
+  return true
+end
+
```
wpferguson: You really don't need this function unless you are going to restrict the formats that are available to the user.  You can remove this and just put nil for the argument to dt.register_storage

wpferguson:
One last thought on the code is there should be a cleanup process where the exported files get removed after the video is built.  The cleanup process should also run if there is an error causing premature exit.

contrib/timelapse.lua
```diff
@@ -0,0 +1,463 @@
...
+  for i, name in ipairs(new_items) do
+    combobox[i] = name
+  end
+  if old_elements_count > #new_items then
+    for j = old_elements_count, #new_items + 1, -1 do
+      combobox[j] = nil
+    end
+  end
+
+  combobox.value = to_select_idx
```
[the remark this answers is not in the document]
theres: yes, I'm aware and it's expected - I want to update preferences when value change.

contrib/timelapse.lua
theres: This function allow to format string by substitute tokens like `{label}` by value provided in `symbols` table
  ex:
```lua
  format_string("I like {bananas} and {potatos}. {bananas}!", {["bananas"]: "darktable", ["potatos"]: "lua"}) 
-> "I like darktable and lua. darktable!"
```
I you want to preserve given curly braces you need to escape it by backslash like so:

```lua
format_string("it preserve \\{label\\} but substitute {label}", {["label"]: "this"}) 
-> "It preserve {label} but substitute "
```
I'll change name to a bit more meaningful and add some comments.

contrib/timelapse.lua
theres: I'm afraid I cannot do this:
1. first thing is that user might want to run/schedule a few bigger exports with different settings in one time. If I move reading of settings to the post-export phase, it will read the last settings, not the settings provided when export was initiated.
2. as far as I know finalize_export gets dictionary of exported images, that map `image file -> exported file name` but do **not** preserve original files order. This may be quite a big issue in video.

theres:
@wpferguson @supertobi thanks for revew, sorry it took me so long to respond. I'll push post-review fixes today. A few notes:

> Could you perhaps rename it to "video_ffmpeg.lua" because we already have a "video_mencoder.lua". That would make the naming a little bit more consistent.

I can do that for sure, consistency is good argument, but I don't really like `video_*` name in the first place. `timelapse_*` seems more specific. I've never search for _video plugin_, but for _timelapse_ one to be honest. This is probably why I've missed `video_mencoder` and started this one.

> One last thought on the code is there should be a cleanup process where the exported files get removed after the video is built. The cleanup process should also run if there is an error causing premature exit.

I'm pretty sure `rm(tmp_dir)` from :451 should do the job, at least on linux. Did I miss something?

About moving `rm`/`mkdir`/`launch_default_app` to the file & system libs - I'd love to, but I haven't really tested it outside Linux (windows/macos version is based on stackoverflow ;)). So testing needed.

contrib/timelapse.lua
theres: done

contrib/timelapse.lua
theres: done

contrib/timelapse.lua
theres: nice! done.

contrib/timelapse.lua
theres: renamed & moved to system lib

contrib/timelapse.lua
theres: done

contrib/timelapse.lua
theres: done

contrib/video_ffmpeg.lua
```diff
@@ -0,0 +1,448 @@
...
+local destination_label = dt.new_widget("section_label"){
+  label = _("output file destination"),
+  tooltip = _("settings of output file destination and name")
+}
+
+local output_directory_chooser = dt.new_widget("file_chooser_button"){
+  title = _("Select export path"),
+  is_directory = true,
+  tooltip =_("select the target directory for the timelapse. \nthe filename is created automatically."),
+  value = string_pref_read("export_path", os.getenv("HOME")),
```
supertobi: One small improvement could be to set the default export directory to the Video folder and not home. Have a look into the "video_mencoder.lua" how we do it there:
https://github.com/darktable-org/lua-scripts/blob/master/contrib/video_mencoder.lua#L107

supertobi:
@wpferguson for me it looks fine now. I think every thing else can be improved later. If you can't see any big problem please merge it.

supertobi:
And after we have merged this script we need to add a big warning into the memcoder script that it is deprecated and only runs on Linux.

contrib/video_ffmpeg.lua
theres: good idea, I'll do that. I just need to figure out how to do it on Windows/Mac, I'm pretty sure `xdg-user-dir` wont work ;)

wpferguson:
I haven't had a chance to test on windows yet.  I'm shooting games tonight, so it'll probably be tomorrow night.

contrib/video_ffmpeg.lua
wpferguson: On a mac it's $home/Movies

contrib/video_ffmpeg.lua
```diff
@@ -0,0 +1,448 @@
...
+  tooltip = _("open video file after successful export"),
+  value = dt.preferences.read(MODULE_NAME, "open_after_export", "bool"),
+  clicked_callback = function (widget)
+    dt.preferences.write(MODULE_NAME, "open_after_export", "bool",  widget.value)
+  end
+}
+
+local widgets_list = {}
+if not df.check_if_bin_exists("ffmpeg") then
+  table.insert(timelapse_widgets, df.executable_path_widget("ffmpeg"))
```
wpferguson: Instead of timelapse_widgets, it should be widgets_list and executable_path_widget expects a table of executables so it should be {"ffmpeg"}

theres: eh, yes, should be fixed now.

contrib/video_ffmpeg.lua
theres: hope it's more or less working now.

I've found that on Windows default video location can be changed to any directory you want. If I'm not mistaken, the only way to retrieve this information is to read registry. I don't know how frequent case it is. I'm open for contributions ;)

wpferguson:
I'm not sure what the contrib/info.lua has to do with video_ffmpeg.  Was adding this file a mistake?

theres:
@wpferguson yes, removed, sorry.

contrib/video_ffmpeg.lua
```diff
@@ -0,0 +1,448 @@
...
+local function show_status(enf_storage, image, format, filename, number, total, high_quality, extra_data)
+  dt.print(_("export ")..tostring(number).." / "..tostring(total))   
+end
+
+local function init_export(storage, img_format, images, high_quality, extra_data)
+  -- store filename preference here cause there is no changed_callback on entry yet
+  string_pref_write("filename_entry", "text")(filename_entry)
+
+  extra_data["images"] = images -- needed, to preserve images order
+  extra_data["tmp_dir"] = dt.configuration.tmp_dir .. "/"..MODULE_NAME .. "_" .. os.time()
```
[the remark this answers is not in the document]
theres: done

lib/dtutils/file.lua
```diff
@@ -553,5 +554,50 @@ function dtutils_file.sanitize_filename(filename)
...
+  Limitations = [[]],
+  Example = [[]],
+  See_Also = [[]],
+  Reference = [[]],
+  License = [[]],
+  Copyright = [[]],
+}
+function dtutils_file.mkdir(path) 
+  local mkdir_cmd = dt.configuration.running_os == "windows" and "mkdir" or "mkdir -p"
+  return dsys.external_command(mkdir_cmd.." "..dtutils_file.sanitize_filename(path))
```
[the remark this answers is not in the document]
theres: done

contrib/video_ffmpeg.lua
```diff
@@ -272,11 +272,22 @@ local destination_label = dt.new_widget("section_label"){
...
+  defaultVideoDir = handle:read()
+  handle:close()
+end
+
 local output_directory_chooser = dt.new_widget("file_chooser_button"){
   title = _("Select export path"),
   is_directory = true,
   tooltip =_("select the target directory for the timelapse. \nthe filename is created automatically."),
-  value = string_pref_read("export_path", os.getenv("HOME")),
+  value = string_pref_read("export_path", defaultVideoDir),
```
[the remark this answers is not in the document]
theres: done

PR #158, opened 2019-01-23
Lua script for tracking the Git status of files.

danielschudel:
I use Git to archive/distribute all my XMP files and my best RAW files. I wanted visual indicators in Darktable that would tell me an image's Git status (untracked, staged, committed, etc.). This Lua add-on does that with the color flags (red/yellow/green) and the "Image Information" box.

Additionally - simple "git add/remove" widget is available to manage individual files.

More details are in the comments of the new git.lua file.

### DIFF
--- a/contrib/git.lua
+++ b/contrib/git.lua
@@ -1,0 +1,245 @@
+--[[
+   git.lua - Report Git status for image files and sidecar files.
+             Perform basic repo operations: add, rm
+
+   Copyright (C) 2018 Daniel Schudel <dan.schudel@gmail.com>.
+
+   This program is free software: you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 3 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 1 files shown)


contrib/git.lua
```diff
@@ -0,0 +1,245 @@
...
+  local sidecarFilename
+  local result
+
+  for _,image in ipairs(sel_images) do
+    sidecarPath, sidecarFilename = string.match(image.sidecar, "(.-)([^/]-([^%.]+))$")
+
+    dt.print_log(" ", image.path .. " / " .. image.filename)
+    dt.print_log(" ", sidecarPath .. " / " .. sidecarFilename)
+
+    result = os.execute("git -C " .. image.path .. " " .. action .. " " .. image.filename)
```
supertobi: Have you tested this with filenames with spaces? we have a the utility function df.sanitize_filename() in "lib/dtutils.file". Perhaps you can use it here.

contrib/git.lua
```diff
@@ -0,0 +1,245 @@
...
+local function git_status_of_file(path, filename)
+  local result
+  local cmd
+  local f
+  local output
+
+--	dt.print_log(path .. " / " .. filename)
+  result = os.execute("git -C " .. path .. " rev-parse --git-dir 2>&1 > /dev/null 2>&1 ")
+  if result == nil then
+    return "No Git repo"
```
supertobi: You should make your strings translatable, we have an example how to do that here:
https://github.com/darktable-org/lua-scripts/blob/master/examples/gettextExample.lua

PR #162, opened 2019-02-13
Add CollectHelper

BzKevin:
Adds plugin CollectHelper which gives the user a few helpful buttons to change collections based on selected image.

### DIFF
--- a/contrib/CollectHelper.lua
+++ b/contrib/CollectHelper.lua
@@ -152,27 +152,27 @@ end
 
 -- GUI --
 dt.gui.libs.image.register_action(
-	"Return to Previous Collection",
+	"collect: previous",
 	function() PreviousCollection() end,
 	"Sets the Collect parameters to be the previously active parameters"
 )
 if dt.preferences.read('module_CollectHelper','folder','bool') then
 	dt.gui.libs.image.register_action(
-		"Collect on image's Folder",
+		"collect: folder",
 		function() CollectOnFolder() end,
 		"Sets the Collect parameters to be the selected images's folder"
 	)
 end
 if dt.preferences.read('module_CollectHelper','colors','bool') then
 	dt.gui.libs.image.register_action(
-		"Collect on image's Color Label(s)",
+		"collect: color label(s)",
 		function() CollectOnColors() end,
 		"Sets the Collect parameters to be the selected images's color label(s)"
 	)
 end
 if dt.preferences.read('module_CollectHelper','all_and','bool') then
 	dt.gui.libs.image.register_action(
-		"Collect on All (AND)",
+		"collect: all (AND)",
 		function() CollectOnAll_AND() end,
 		"Sets the Collect parameters based on all activated CollectHelper options"
 	)
--- a/contrib/CollectHelper.lua
+++ b/contrib/CollectHelper.lua
@@ -1,0 +1,199 @@
+--[[Collect Helper plugin for darktable
+
+  copyright (c) 2019  Kevin Ertel
+  
+  darktable is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+  
+  darktable is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+  
+  You should have received a copy of the GNU General Public License
+  along with darktable.  If not, see <http://www.gnu.org/licenses/>.
+]]
+
+--[[About this plugin
+This plugin adds the button(s) to the "Selected Images" module:
+1) Return to Previous Collection
+2) Collect on image's Folder
+3) Collect on image's Color Label(s)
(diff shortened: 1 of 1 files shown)


contrib/CollectHelper.lua
```diff
@@ -0,0 +1,199 @@
...
+
+Select the photo you wish to change you collection based on.
+In the "Selected Images" module click on "Collect on this Image"
+
+----KNOWN ISSUES----
+]]
+
+local dt = require "darktable"
+previous = nil
+all_active = false
```
supertobi: Please add a local in front of the variables.

contrib/CollectHelper.lua
```diff
@@ -0,0 +1,199 @@
...
+
+-- FUNCTION --
+local function CheckSingleImage(selection)
+	if #selection ~= 1 then
+		dt.print("Please select a single image")
```
supertobi: Please consider to make the strings translatable.

contrib/CollectHelper.lua
```diff
@@ -0,0 +1,199 @@
...
+	end
+	return 0
+end
+local function PreviousCollection()
+	if previous ~= nil then
+		previous = dt.gui.libs.collect.filter(previous)
+	end
+end
+local function CollectOnFolder()
+	local images = dt.gui.selection()
```
wpferguson: The API manual recommends using dt.gui.action_images() instead of dt.gui.selection().

wpferguson:
@BzKevin I was thinking of writing a compare function similar to lightoom's using the collect filter to do it.  I figured that I was going to have to play with collect to figure out how it worked, but you did it for me.  Thanks

BzKevin:
@wpferguson Awesome, if it is similar enough in content then maybe adding it in to this same script makes sense? Certainly it wouldn't be hard to duplicate the functionality I've already implemented for the other collect parameters, like filmstrips, day, place, etc.

But if you're thinking of making a module with checkboxes for each parameter and then filtering on the selected properties with the selected image's value that my be better served in a new script. Which would be an awesome script by the way.

contrib/CollectHelper.lua
BzKevin: Yea, lets talk about that real quick. Here are my thoughts:
1) I use to get some glitches when scrolling real fast through big collections, or moving my mouse in/out of the lightable grid very quickly. In these cases an image was staying hovered when it shouldn't. This was frustrating. This doesn't seem to happen anymore though.
2) Theoretically the mouse will be on the button when activated, so nothing could ever be hovered

But it boils down to: personally I get frustrated a lot when using dt by the selected/hovered behavior and feel that selected works better for this script in case of any hovered glitches. Also this script requires a single image be selected/active to run. So if I included hovered behavior than it would force the user to either hover over the selected image or none at all, so it's really only making the script potentially more restrictive on user behavior than needed. 

If you all feel very strongly this is really bad then we can change it for inclusion in the repo. (I'll probably revert it back on my local machine though)

contrib/CollectHelper.lua
BzKevin: I'll comb through and do this once I get a chance.

contrib/CollectHelper.lua
BzKevin: I can do this at some point, could be a while though. At the moment I don't really know how the translation side of things works. Maybe it's real simple and easy, I just haven't looked so I don't know.

contrib/CollectHelper.lua
BzKevin: This is done in my latest commit.

contrib/CollectHelper.lua
wpferguson: Making it translatable just means including the gettext stuff, defining the function _("msg") and wrapping your strings with _().  You can look at the top of gimp.lua or in the examples at the gettextExample.lua.

contrib/CollectHelper.lua
```diff
@@ -0,0 +1,198 @@
...
+	else
+		previous = dt.gui.libs.collect.filter(rules)
+	end
+end
+local function CollectOnColors(all_rules, all_active)
+	local images = dt.gui.selection()
+	if CheckSingleImage(images) then
+		return
+	end
+	for _,image in pairs(images) do
```
wpferguson: Checking that the image has a colorlabel is probably a good idea, although the function survived and just put a blank rule in the recently used collections module.

contrib/CollectHelper.lua
wpferguson: I've not personally seen the hover issue where an image remains hovered, but I could see it happening if a machine got busy.  Your reasoning is sound, especially about not wanting the hover, so just leave it as it is.

contrib/CollectHelper.lua
BzKevin: that's easy enough to do. I'll make that change sometime this week and update.

contrib/CollectHelper.lua
BzKevin: I have updated the latest script with what I believe to be the correct implementation.

contrib/CollectHelper.lua
BzKevin: recently uploaded version adds a check to ensure at least one color label is applied to the selected image before proceeding. if a label is not active on the image a message is displayed to the user.

wpferguson:
Looks good.

PR #163, opened 2019-02-13
Add OpenInExplorer

BzKevin:
Adds plugin OpenInExplorer which, for Windows OS users, adds a button to open the selected files in File Explorer

### DIFF
--- a/contrib/OpenInExplorer.lua
+++ b/contrib/OpenInExplorer.lua
@@ -1,0 +1,80 @@
+--[[OpenInExplorer plugin for darktable
+
+  copyright (c) 2018  Kevin Ertel
+  
+  darktable is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+  
+  darktable is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+  
+  You should have received a copy of the GNU General Public License
+  along with darktable.  If not, see <http://www.gnu.org/licenses/>.
+]]
+
+--[[About this plugin
+This plugin adds the module "OpenInExplorer" to darktable's lighttable view
+
+----REQUIRED SOFTWARE----
+Microsoft Windows Operating System
+
+----USAGE----
+Install: (see here for more detail: https://github.com/darktable-org/lua-scripts )
+ 1) Copy this file in to your "lua/contrib" folder where all other scripts reside. 
+ 2) Require this file in your luarc file, as with any other dt plug-in
+
+Select the photo(s) you wish to find in explorer and press "Go to Folder". 
+A file explorer window will be opened for each selected file at the file's location; the file will be highlighted.
+
+----KNOWN ISSUES----
+]]
+
+local dt = require "darktable"
+local df = require "lib/dtutils.file"
+local dsys = require "lib/dtutils.system"
+require "official/yield"
+
+--Detect OS and modify accordingly--	
+local proper_install = 0
+if dt.configuration.running_os == "windows" then
+	proper_install = 1
+end
+
+-- FUNCTION --
+local function OpenInExplorer() --Open in Explorer
+	--Inits--
+	if proper_install ~= 1 then
+		return
+	end
+	local images = dt.gui.selection()
+	local curr_image = ""
+	
+	for _,image in pairs(images) do 
+		curr_image = image.path..'\\'..image.filename
+		local run_cmd = "explorer.exe /select, "..curr_image
+		dt.print_log("OpenInExplorer run_cmd = "..run_cmd)
+		resp = dsys.external_command(run_cmd)
+	end
+end
+
+-- GUI --
+OpenInExplorer_btn_run = dt.new_widget("button"){
+	label = "Go to Folder",
+	tooltip = "Opens selected image(s) location(s) in file explorer",
+	clicked_callback = function() OpenInExplorer() end
+	}
+dt.register_lib( --OpenInExplorer
+	"OpenInExplorer_Lib",	-- Module name
+	"Open In Explorer",	-- name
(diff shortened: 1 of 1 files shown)


supertobi:
I have now Windows installed on an other machine and this script works fine. 

I have two comments for later improvements:
1.  Make the strings translatable. Have a look at the gettext example:
https://github.com/darktable-org/lua-scripts/blob/master/examples/gettextExample.lua
2. Have a look into the openInNautilus.lua pull request and merge the two scripts:
https://github.com/darktable-org/lua-scripts/pull/73/files

contrib/OpenInExplorer.lua
```diff
@@ -0,0 +1,69 @@
...
+		resp = dsys.external_command(run_cmd)
+	end
+end
+
+-- GUI --
+if proper_install then
+	dt.gui.libs.image.register_action(
+		"show in file explorer",
+		function() OpenInExplorer() end,
+		"Opens File Explorer at the selecte image's location"
```
wpferguson: typo, should be selected

contrib/OpenInExplorer.lua
```diff
@@ -0,0 +1,69 @@
...
+-- FUNCTION --
+local function OpenInExplorer() --Open in Explorer
+	--Inits--
+	if not proper_install then
+		return
+	end
+	local images = dt.gui.selection()
+	local curr_image = ""
+	
+	for _,image in pairs(images) do
```
wpferguson: Do you want to set a maximum for the number of copies of explorer that you open.  For instance, most of my filmrolls are 1000+ pictures.  If I hit select all, then "show in file explorer", I'm going to have a horrible mess on my hands.  Maybe a limit of 10?

contrib/OpenInExplorer.lua
```diff
@@ -0,0 +1,69 @@
...
+	--Inits--
+	if not proper_install then
+		return
+	end
+	local images = dt.gui.selection()
+	local curr_image = ""
+	
+	for _,image in pairs(images) do 
+		curr_image = image.path..'\\'..image.filename
+		local run_cmd = "explorer.exe /select, "..curr_image
```
wpferguson: macos is like windows in that there is one file manager.  You can open a directory with _open /path/to/directory/_ or open with an image selected with _open -R /path/to/image.xxx_.  

Nautilus is the default installed file manager on linux.  You could start with this.  With nautilus you can open the directory and select the file with _nautilus file:///path/to/image.xxx_  With nautilus you can have multiple arguments and a file manager window will open with each one.

I've figured out a method to determine the default file manager in linux, but I have to build a VM and test it.  I didn't want to clutter my install with multiple file managers.

contrib/OpenInExplorer.lua
```diff
@@ -0,0 +1,69 @@
...
+end
+
+-- FUNCTION --
+local function OpenInExplorer() --Open in Explorer
+	--Inits--
+	if not proper_install then
+		return
+	end
+	local images = dt.gui.selection()
+	local curr_image = ""
```
wpferguson: What about the case where the user doesn't select an image.  You could call dt.collection(), use the first image, and just open the directory without an image selected.  Or, you can do nothing, or pop a message that no images were selected.

contrib/OpenInExplorer.lua
```diff
@@ -0,0 +1,69 @@
...
+		local run_cmd = "explorer.exe /select, "..curr_image
+		dt.print_log("OpenInExplorer run_cmd = "..run_cmd)
+		resp = dsys.external_command(run_cmd)
+	end
+end
+
+-- GUI --
+if proper_install then
+	dt.gui.libs.image.register_action(
+		"show in file explorer",
```
wpferguson: If you make it cross platform, you can change explorer to manager.

contrib/OpenInExplorer.lua
```diff
@@ -0,0 +1,69 @@
...
+
+local dt = require "darktable"
+local dsys = require "lib/dtutils.system"
+
+--Detect OS and modify accordingly--	
+local proper_install = false
+if dt.configuration.running_os == "windows" then
+	proper_install = true
+end
+
```
wpferguson: I would check the os then pop an error message and not install if it 's not windows (unless you make it cross platform).  That would be better than having a button that does nothing.

contrib/OpenInExplorer.lua
BzKevin: Yea, that's a good idea

wpferguson:
@BzKevin what do you use this for?  The one use case I can see for me is that sometimes I want to open the directory, but I don't really care about having an image selected.  I'm more interested in getting to the export directories.

@supertobi as soon as I saw this I was thinking about merging PR #73 into it.  If we merge the two, then we just close #73, with a note that we merged it?

supertobi:
> @supertobi as soon as I saw this I was thinking about merging PR #73 into it. If we merge the two, then we just close #73, with a note that we merged it?

Yes, I think that's the best solution.

contrib/OpenInExplorer.lua
BzKevin: I think it would make sense to eventually make this cross platform compatible as you suggest. I might not do it immediately. but this does make sense to do.

contrib/OpenInExplorer.lua
BzKevin: When I make the other changes I'll add a popup message. Good thought.

BzKevin:
@wpferguson I admittedly don't use this as much as I once did. Previously I would use it if I wanted to edit an image in PS, say to remove a glitch from enfuse, or to remove a streetlight, bird, person, etc. I do all those type edits in PS. Or sometimes I'll do this on a group of images to bring in to hugin, or that I want to copy to a thumb drive for sharing, etc. 

Other scripts and features of dt I didn't know about originally have meant I use this far less then previously, but nonetheless, I still have rare occasions where I find I use this and clicking a button is so much easier than looking at the path an navigating there.

contrib/OpenInExplorer.lua
BzKevin: I made a pop up message in my latest commit. PR pending.

However, I do not know what you mean by "...and not install if it's not windows..." How do I disable the script from installing itself?

PR #164, opened 2019-02-13
Add HDRMerge

BzKevin:
Adds a plugin for HDRMerge capabilities. This should be cross-platform compatible.

### DIFF
--- a/contrib/HDRMerge.lua
+++ b/contrib/HDRMerge.lua
@@ -308,31 +308,17 @@ dt.register_lib(
 )
 
 -- PREFERENCES --
-HDRMerge_entry_widget_tags = dt.new_widget("entry"){
-	tooltip = "Seperate with commas, all spaces will be removed",
-	text = nil,
-	placeholder = "Enter default tags",
-	editable = true
-}
 dt.preferences.register("module_HDRMerge", "add_tags",	-- name
 	"string",	-- type
 	'HDRMerge: Defualt additional tags',	-- label
 	'Changes DEFAULT entry in the additional tags option. Requires restart to take effect.',	-- tooltip
-	"",	-- default
-	HDRMerge_entry_widget_tags
+	""	-- default
 ) 
-HDRMerge_entry_widget_style = dt.new_widget("entry"){
-	tooltip = "Enter the style name exactly as it is",
-	text = nil,
-	placeholder = "Enter Style name",
-	editable = true
-}
 dt.preferences.register("module_HDRMerge", "style",	-- name
 	"string",	-- type
 	'HDRMerge: Defualt Style',	-- label
 	'Changes DEFAULT entry in the Style option. Requires restart to take effect.',	-- tooltip
-	"",	-- default
-	HDRMerge_entry_widget_style
+	""	-- default
 ) 
 dt.preferences.register("module_HDRMerge", "copy_tags",	-- name
 	"bool",	-- type
@@ -358,6 +344,5 @@ dt.preferences.register("executable_paths", "HDRMerge",	-- name
 	"file",	-- type
 	'HDRMerge: Binary Location',	-- label
 	'Install location of HDRMerge[.exe]. Requires restart to take effect.',	-- tooltip
-	"HDRMerge",	-- default
-	HDRMerge_path_widget
+	"HDRMerge"	-- default
 )
--- a/contrib/HDRMerge.lua
+++ b/contrib/HDRMerge.lua
@@ -52,14 +52,9 @@ local dsys = require "lib/dtutils.system"
 require "official/yield"
 
 --Check if HDRMerge is installed--
-local not_installed = 0
-dt.print_log("HDRMerge - Executable Path Preference: "..df.get_executable_path_preference("HDRMerge"))
-local HDRMerge_path = df.check_if_bin_exists("HDRMerge")
-if not HDRMerge_path then
-	dt.print_error("HDRMerge not found")
-	dt.print("ERROR - HDRMerge not found")
-	not_installed = 1
-end
+local first_run = true
+local not_installed = false
+local HDRMerge_path
 
 --Detect OS and modify accordingly--	
 local os_path_seperator = "/"
@@ -101,7 +96,16 @@ local function build_execute_command(cmd, args, file_list)
 	return result
 end
 local function HDRMerge()
-	if not_installed == 1 then
+	if first_run then
+		HDRMerge_path = df.check_if_bin_exists("HDRMerge")
+		if not HDRMerge_path then
+			dt.print_error("HDRMerge not found")
+			dt.print("ERROR - HDRMerge not found")
+			not_installed = true
+		end
+		first_run = false
+	end
+	if not_installed then
 		dt.print("Required software not found")
 		dt.print_log("Required software not found - HDRMerge did not run")
 		return
@@ -350,15 +354,6 @@ dt.preferences.register("module_HDRMerge", "bits_per_sample",	-- name
 	"16",	-- default
 	"24","32"	--value
 )
-
-if not HDRMerge_path then 
-	HDRMerge_path = ""
-end
-HDRMerge_path_widget = dt.new_widget("file_chooser_button"){
-	title = "Select HDRMerge[.exe] file",
-	value = HDRMerge_path,
-	is_directory = false,
-}
 dt.preferences.register("executable_paths", "HDRMerge",	-- name
 	"file",	-- type
 	'HDRMerge: Binary Location',	-- label
--- a/contrib/HDRMerge.lua
+++ b/contrib/HDRMerge.lua
@@ -51,312 +51,312 @@ local mod = 'module_HDRMerge'
 local os_path_seperator = '/'
 if dt.configuration.running_os == 'windows' then os_path_seperator = '\\' end
 
-dt.preferences.register("executable_paths", "HDRMerge",	-- name
-	"file",	-- type
-	'HDRMerge: binary location',	-- label
-	'Install location of HDRMerge. Requires restart to take effect.',	-- tooltip
-	"HDRMerge"	-- default
+dt.preferences.register("executable_paths", "HDRMerge", -- name
+    "file", -- type
+    'HDRMerge: binary location',    -- label
+    'Install location of HDRMerge. Requires restart to take effect.',   -- tooltip
+    "HDRMerge"  -- default
 )
 local temp
 local HDRM = { --HDRMerge Program Table
-	name = 'HDRMerge',
-	bin = '',
-	first_run = true,
-	install_error = false,
-	arg_string = '',
-	images_string = '',
-	args = {
-		bps			={text = '-b ', style = 'integer'},
-		size		={text = '-p ', style = 'string'},
-		batch		={text = '-B ', style = 'bool'},
-		gap			={text = '-g ', style = 'integer'}
-	}
+    name = 'HDRMerge',
+    bin = '',
+    first_run = true,
+    install_error = false,
+    arg_string = '',
+    images_string = '',
+    args = {
+        bps         ={text = '-b ', style = 'integer'},
+        size        ={text = '-p ', style = 'string'},
+        batch       ={text = '-B ', style = 'bool'},
+        gap         ={text = '-g ', style = 'integer'}
+    }
 }
 local GUI = { --GUI Elements Table
-	HDR = {
-		bps 		={},
-		size		={},
-		batch		={},
-		gap			={}
-	},
-	Target = {
-		style		={},
-		copy_tags	={},
-		add_tags	={}
-	},
-	run = {}
+    HDR = {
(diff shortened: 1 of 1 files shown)


supertobi:
I have no experiences with HDRMerge, could you perhaps add a folder with an example image set for testing?
@wpferguson could you have a look into the cross platform stuff?

BzKevin:
I could not attached due to file size/type so here is a link to some raws for testing:
https://drive.google.com/drive/folders/18RuMOg-iWg9d0xR_TqA2TgIYu4iPa-HA?usp=sharing

HDRMerge is basically supposed to take in a stack of bracketed raw photos and merge them in to a single DNG with a wide dynamic range.

contrib/HDRMerge.lua
```diff
@@ -0,0 +1,348 @@
...
+Auto-import Options:
+Select a style, whether you want tags to be copied from the original, and any additional tags you desire added when the new image is auto-imported
+
+----KNOWN ISSUES----
+]]
+
+local dt = require "darktable"
+local df = require "lib/dtutils.file"
+local dsys = require "lib/dtutils.system"
+require "official/yield"
```
wpferguson: you don't need the require "official/yield".  It's included in lib/dtutils.system

contrib/HDRMerge.lua
```diff
@@ -0,0 +1,348 @@
...
+		HDRMerge_chkbtn_cpytags,
+		HDRMerge_lbl_tags,
+		HDRMerge_entry_tag,
+		HDRMerge_btn_run,
+		HDRMerge_lbl_note
+	}
+)
+
+-- PREFERENCES --
+dt.preferences.register("module_HDRMerge", "add_tags",	-- name
```
wpferguson: You don't need to register the preferences.  You can just use a callback on your gui elements and dt.preferences.write() to save them.  This gets rid of the necessity to restart the program just to use the module.

contrib/HDRMerge.lua
BzKevin: Ok, good to know. Thanks! I'll make this change when I rework the preferences per below.

contrib/HDRMerge.lua
BzKevin: I'll make this change.

BzKevin:
Let me make the few changes requested here. Similarly, I have a much better grasp on LUA and the dt API then when I first wrote this, so I may take some time to clean a few other things up. I also had some glitchy behavior on my Win 10 dt 2.6 machine the other day I need to investigate. I'll update back here with changes.

contrib/HDRMerge.lua
BzKevin: this has been removed

contrib/HDRMerge.lua
BzKevin: I refactored this script and the new version only registers a preference for the binary location. However, I was thinking that even the binary location may not need to be registered on all OS types. I know it is needed on Windows, I believe it is NOT needed on linux, I have no clue about MacOS. If we make that determination I can wrap the binary preference registry in an if statement based on OS type.

BzKevin:
I have uploaded a refactored version.

A few weeks back, after upgrading to dt 2.6 I had some issue with my enfuseAdvanced script. That code was a mess and a hack-job from Holger Klemms original so troubleshooting was a nightmare. I decided it was best to refactor that code and developed a clear structure for programs and GUI elements. (I'll be making a PR for the refactored enfuseAdvanced soon, once I've got more testing on it). I maintained that same structure in this newly refactored HDRMerge. 

The new structuring was probably overkill for this relatively simple script, but I am hoping the consistency across scripts will prove useful in the long run and make maintenance and understanding easier.

contrib/HDRMerge.lua
wpferguson: I think the only time you should use dt.register_preference() is if you don't have a user interface, i.e. it's only a shortcut, or a button on the selected images module. If it's an exporter or a module, then put the preferences in the UI. The problem with dt.register_preference() is that you have to exit darktable in order for the preference to take effect.  If you have a user interface, then put the executable path widget in the UI and save the preference right away so the use can use the script without having to restart darktable.  That's why I created df.executable_path_widget(), so all that could be done and included in the UI with a minimum of coding.  The only thing that would make it better is if you ran the script without the executable path and it threw an error and drew a box around the executable_path_preference in red so that the user knew exactly where to go to fix it.

BzKevin: Ok, I can add an executable path widget to the module rather than in preferences. I just though it was cleaner long-term to have the executable in the preferences. You only have to set it once and then never again, so it seemed to me having it displayed in the UI of the widget for just a single use was extraneous. And you already have to go in to the Luarc file to install new widgets, so I figured one more "install step" was a minor price to pay for the cleaner UI. 

There may be a best of both worlds, I'll have to play. I notice preferences.register() has a "set_callback" I wonder if this can be used to trigger updating something that contains the executable path.

wpferguson: I don't get the luarc/new widget thing.  Did you mean scripts?  If so, copy tools/script_manager.lua to your luarc, in effect making it a replacement.  Now you never have to edit your luarc again.  You just enable and disable scripts from the script_manager interface.  As a bonus for developers, if you have a syntax error and your script fails to load, you can fix the syntax error, then click enable <script> again and it will try and load it again without having to stop and start darktable.  If you have a logic error, then you have to stop and start darktable because I still haven't figured out how to unload a script (actually unload a registered anything because there aren't functions to do that, until I figure you how to add unregister :-D)

Part of my thought process is big picture.  Let's say  we have 50 scripts using on average 1.5 executables which would work out to 75 preferences.  Unfortunately of those 75, 30 of them are multiple definitions of the same executable.  For instance align_image_stack is used by contrib/image_stack.lua, official/enfuse.lua, contrib/enfuseAdvanced.lua, and maybe more.  exiftool is also used multiple times.  So, the user has to figure out which one to change because they don't realize they are really the same.  Programmers have to put them in there because they can't rely on the users knowing the one exiftool executable preference is as good as another. And, the user has to go away from what he was doing and dig through the configuration data to find where to make the change, then make the change.  And then he has to restart darktable, which is my biggest hangup about register preferences.  My belief is that the user should start darktable once and use it until they're done and not have to restart it when they want to use a script, which is the reason I wrote script manager.  User's shouldn't have to know lua in order to enable the scripts.

The whole gui element can really be solved with a call to check_if_bin_exists().  If it does, don't load the executable_path_widget.  So therefore I could write a new script maybe called compare :-) and use exiftool and if it's already defined by another script then I don't ever need to ask the user for it, but the code's there just in case.  

If you decide to do it this way, then to test it you'll have to edit your darktablerc file and remove the "lua/executable_paths/<script>=" lines, or just remove the values.

BzKevin: I don't use the script manager; I wasn't really aware of it or what it does. Sounds useful though. I'll have to load it up. But, as I don't use it right now I go in an manually modify my LUARC file for every new script. Your point on the big picture is a very valid one. I admit I still don't thing a single restart once you install a new script is a big deal, however, I certainly agree that having an over abundance of preferences, and likely a bunch of dupilcates, is bad, so I'm on board with changing the way executables are defined, especially if it can be done in a way that is minimally invasive to the normal module GUI. My past scripts have certainly been culprits of over-utilizing preferences and that was part of the driving reason I wanted to refactor HDRMerge and enfuseAdvanced. 

I had previously thought about wrapping the preferences.register() in an if statement that checks if it already exists. The only problem I thought about with that was the following: what if the user accidentally (or intentionally) assigns a different executable, and then the ability to correct it disappears? For example: On my dev laptop I didn't have HDRMerge installed for a long time, but I still tested my HDRMerge script by just pointing to hugin.exe. Maybe this is a corner case we don't need to care about...

If we either don't care about that, or find a way around it then I could imagine doing the following:
putting all gui elements in box containers inside a stack. One box holds your normal GUI elements, the other just executable choosers. When the script loads it checks if the bin exists. If it doesn't then the executable chooser container is made active, otherwise the normal GUI is active. Once the user sets the executable path and it's verified with a check_if_bin_exists() the GUI container is made active. Thus the path choosers are always hidden by being not active in the stack unless we detect they need to be shown at startup.

wpferguson: You really should load script_manager.  It lets you see all that is available, and if you hover over the button it tells you what it does.  Then you could look at tools/executable_manager.lua which provides a way to manipulate all the executable path preferences and redefine the location or clear it if necessary.

The nice thing about script manager is it rescans the lua directory every time you start darktable, so it picks up new scripts automatically.  You can also add extra directories like development, testing, mynewtoys, etc. and it picks those up too and just adds new categories.  The best thing about script_manager, in my opinion, is that a misbehaving script doesn't stop other scripts from loading.  If you use the luarc file and your first script has a syntax error, then nothing loads.  

I remember when I first tried to use the scripts and I didn't know lua.  It took 20 or 30 minutes before I finally figured it out, starting with # isn't a comment :-).  At that point I realized that the lua-scripts weren't really accessible to normal users, unless you had a programming background or a lot of perseverance.  I wrote the first script_manager a few months later and have been using that as my lua-script interface ever since.  It went through several iterations before I thought it was fit for public consumption, but I finally got there.

The other feature I use a lot, is that I only turn on what I'm using and when I'm done I turn it off. If I'm in the middle of an editing session and I decide I want to finish an image in gimp, I just to go script_manager, turn it on, then use it, all without having to restart darktable and lose my train of thought. :-D

BzKevin: I just loaded it up and had a quick play with it. That is very nice, very impressive! I'll start using this. I'll have to setup GIT on my main computer for that.

Let me know if you have any thoughts on how to avoid the situation described above where a user might accidentally select the wrong BIN and then the availability to change it disappears. If you think that's rare enough of a case that we can ignore it and be fine, I'll go ahead and start working on these changes across multiple of my scripts.

wpferguson: If a user accidentally chooses the wrong executable, they can use tools/executable_manager to fix it.  executable_manager lets you change any executable that has been used in the lua scripts.  It has a combobox for all of the executables it knows about.  Select the executable and an executable path widget is displayed along with a clear button.  Select the correct executable, or clear the entry.

BzKevin: Ahh perfect! Well then I'll start working on a different way of prompting user for executable paths in my scripts.

BzKevin:
Just a note that I'll be working on modifying how the executable gets requested from the user. I have suddenly gotten busy with other stuff, so this may take me a bit of time to get around to.

BzKevin:
These latest commits are ready for review. This incorprates changes which eliminate the executable chooser in preferences and instead moves it into the module's GUI.

contrib/HDRMerge.lua
```diff
@@ -0,0 +1,417 @@
...
+    orientation = 'vertical',
+    GUI.exes.HDRMerge,
+    GUI.exes.update
+}
+GUI.stack = dt.new_widget('stack'){
+    GUI.options,
+    exes_box
+}
+if dt.preferences.read(mod, 'bin_exists', 'bool') then
+   GUI.stack.active = 1
```
supertobi: Not a real review, but here is a space missing.

supertobi:
Do you plan to mark the sting for translation? This is something I would say is missing.

BzKevin:
> 
> 
> Do you plan to mark the sting for translation? This is something I would say is missing.

Yea, I do need to do that still.

contrib/HDRMerge.lua
BzKevin: Added space in latest commit

BzKevin:
@supertobi or @wpferguson I'm pretty sure this latest version of the script is all good to go. Have a look when you all get a chance

wpferguson:
My head's been in "code mode" for the last few weeks while I've been trying to finish the API extensions.  That's done now.  I have a couple of life things to take care of that I put off while coding, but I should have those all cleaned up by the start of next week so I can tackle the PR list again.

@supertobi I've been thinking of ways to use the tools on this site such as the projects and wiki to try and communicate better with the developers/users.  We could use the project tools to track each PR through the "stages" of acceptance.  It might also help us "formalize" the process somewhat so that we're not so hit or miss (at least me).  I know that when I'm jumping back and forth I sometimes forget where I was and end up repeating my work. @BzKevin you're welcome to chime in on this as well as any other developers/users.

BzKevin:
@wpferguson no big rush from my standpoint, just wanted to make sure you all were aware I had made what I think to be the last of changes for a while.

I'm unfamiliar with those aspects of Github, I'll take a look and familiarize myself with them so I can offer an informed opinion. Ironically, at my day job we are working heavily to nail down proper processes via a new system we've just switched over to.

BzKevin:
Just a reminder that this should be good to go.

supertobi:
This looks good to me. If there are still problems we can fix it later. I'll merge it now.
@wpferguson I hope this was OK for you.

PR #165, opened 2019-02-15
Update OpenInExplorer

BzKevin:
Changes:
1) Add check if no image selected or more than 15 images selected
2) Add popup error message if not a windows OS detected
3) Fixed a typo

### DIFF
--- a/contrib/OpenInExplorer.lua
+++ b/contrib/OpenInExplorer.lua
@@ -42,6 +42,7 @@ if dt.configuration.running_os == "windows" then
 	proper_install = true
 else
 	dt.print_error('OpenInExplorer plug-in only supports Windows OS at this time')
+	return
 end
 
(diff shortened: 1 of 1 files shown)


contrib/OpenInExplorer.lua
```diff
@@ -40,6 +40,8 @@ local dsys = require "lib/dtutils.system"
 local proper_install = false
 if dt.configuration.running_os == "windows" then
 	proper_install = true
+else
+	dt.print_error('OpenInExplorer plug-in only supports Windows OS at this time')
 end
```
wpferguson: Just add return above this and the script won't install

BzKevin: I opted to add it below the error message so the user would get some visible feedback that the script wasn't loaded

PR #166, opened 2019-02-19
check_min_api_version

wpferguson:
Added check_min_api_version function to dtutils.lua to check the minimum version of the lua API required for a script to run.  This is intended to replace the API check done with darktable.configuration.check_version(), which requires changing each time the API version changes.  check_min_api_version should work for all API version changes unless backward compatibility is broken.  All the scripts using darktable.configuration.check_version() were changed to use check_min_api_version().  check_min_api_version returns true or false letting the author deal with the result appropriately rather than raising an error which would stop further processing or loading by the luarc file.

### DIFF
--- a/contrib/LabelsToTags.lua
+++ b/contrib/LabelsToTags.lua
@@ -48,7 +48,12 @@
 ]]
 
 local darktable = require("darktable")
-darktable.configuration.check_version(...,{3,0,0},{4,0,0},{5,0,0})
+local du = require "lib/dtutils"
+
+if not du.check_min_api_version("3.0.0") then
+  darktable.print("ERROR:LabelsToTags failed to load.  Lua API version 3.0.0 or later required.")
+  return
+end
 
 -- Lua 5.3 no longer has "unpack" but "table.unpack"
 unpack = unpack or table.unpack
--- a/contrib/LabelsToTags.lua
+++ b/contrib/LabelsToTags.lua
@@ -50,10 +50,7 @@
 local darktable = require("darktable")
 local du = require "lib/dtutils"
 
-if not du.check_min_api_version("3.0.0") then
-  darktable.print("ERROR:LabelsToTags failed to load.  Lua API version 3.0.0 or later required.")
-  return
-end
+du.check_min_api_version("3.0.0", "LabelsToTags") 
 
 -- Lua 5.3 no longer has "unpack" but "table.unpack"
 unpack = unpack or table.unpack
--- a/lib/dtutils.lua
+++ b/lib/dtutils.lua
@@ -32,7 +32,40 @@ local dt = require "darktable"
 
 local log = require "lib/dtutils.log"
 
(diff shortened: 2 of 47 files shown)


contrib/LabelsToTags.lua
```diff
@@ -48,7 +48,12 @@
 ]]
 
 local darktable = require("darktable")
-darktable.configuration.check_version(...,{3,0,0},{4,0,0},{5,0,0})
+local du = require "lib/dtutils"
+
+if not du.check_min_api_version("3.0.0") then
+  darktable.print("ERROR:LabelsToTags failed to load.  Lua API version 3.0.0 or later required.")
```
supertobi: I have two comments:
1. I don't like the double  space here: "load.  Lua"
2. Have you tried to generate the error massage in du.check_min_api_version()? Perhaps extend the method with two parameters "showErrorMessage" and "scriptName". What do you think?

wpferguson: Ideally I could exit the routine from the function call if the API isn't compatible.  I tried a couple of things.  Did you know that calling os.exit() kills darktable?  I also tried return -2 to try and pop the stack, but that didn't seem to have an effect.  So, at present, I'm stuck with returning true or false and then the caller has to act on it accordingly.  If I generate the error message in the function then the caller fails to act on it (by doing a return) then the function will continue to execute until it crashes on a part of the API that's missing.  I'll see if I can figure out a way to exit the caller from the version check.  If I can, then I'll move the error message.  Or, I can move the message and we can enforce the syntax on anything we include in the lua-scripts.  We could just code it as 

if not check_min_api_version(<version>,<script>) return end

As for the double spaces after the period, my typing teacher beat it into me along with too many years of using a typewriter.  I don't care either way, but muscle memory often wins out.  If we keep the error messages this way, I'll fix it with a perl 1 liner...  :-)

wpferguson:
It turns out that the only reliable way to stop a script from loading is to throw an error, so I changed check_min_api_version() to take the script name as an argument, print out error messages, and throw an error if the minimum API version is not satisfied.  I updated all the scripts to use the new functionality.

lib/dtutils.lua
```diff
@@ -32,7 +32,42 @@ local dt = require "darktable"
...
+  License = [[]],
+  Copyright = [[]],
+}
+
+function dtutils.check_min_api_version(min_api, script_name)
+  local current_api = dt.configuration.api_version_string
+  if min_api > current_api then
+    dt.print_error("This application is written for lua api version " .. min_api .. " or later.")
+    dt.print_error("The current lua api version is " .. current_api)
+    dt.print("ERROR:" .. script_name .. " failed to load. Lua API version " .. min_api .. " or later required.")
```
supertobi: just one small thing. I would add a space after the ":" here:
dt.print("ERROR: " .. script_name .. " failed to load. Lua API version " .. min_api .. " or later required.")

supertobi:
Looks good, feel free to push it.

PR #168, opened 2019-02-24
Compatibility updates

wpferguson:
Fixed compatibility issues. Added new version check to scripts. Renamed lib/dtutils/file/rm() to lib/dtutils/file/rmdir since removing a file in windows with rmdir /q/s results in an error. Fixed os.tmpname() issue in enfuse. Under windows os.tmpname() doesn't put the file in a temporary directory, so fixed it to use darktable.configuration.tmp_dir. Cleaned up documentation in lib/dtutils/system.lua.


supertobi:
Do you have an reason for not pushing it?

wpferguson:
Because I usually let you look them over and be my second set of eyes just in case I did something that seemed like a good idea at the time :-)

PR #169, opened 2019-02-24
Remove video mencoder

wpferguson:
Remove video_mencoder because mencoder is no longer included by default in most major distributions and contrib/video_ffmpeg.lua offers a better supported solution.


supertobi:
Have you mixed your PRs? This is not just the file removal.

wpferguson:
Not sure what happened here.  I'll close it and redo it correctly.

PR #170, opened 2019-02-24
Remove contrib/video_mencoder.lua

wpferguson:
Removed video_mencoder.lua since mencoder is no longer distributed with most major distributions and video_ffmpeg provides better support.


wpferguson:
@supertobi Fixed it.

wpferguson:
Merged by hand since there were conflicts

PR #171, opened 2019-02-25
make kml_export multi OS

supertobi:
replaced mkdir, xdg-open, xdg-user-dir

I've coded this without windows for testing.
@wpferguson or @BzKevin could you please test if this works.

### DIFF
--- a/contrib/kml_export.lua
+++ b/contrib/kml_export.lua
@@ -286,7 +286,7 @@ local function create_kml_file(storage, image_table, extra_data)
       path = exportDirectory..PS..exportKMLFilename
     end
 
-	dsys.launch_default_app(path)
+    dsys.launch_default_app(df.sanitize_filename(path))
   end
 
 end
--- a/contrib/kml_export.lua
+++ b/contrib/kml_export.lua
@@ -21,7 +21,7 @@ darktable KML export script
 
 ADDITIONAL SOFTWARE NEEDED FOR THIS SCRIPT
 * zip (at the moment Linux only and only if you create KMZ files)
-* convert (ImageMagick)
+* magick (ImageMagick)
 * xdg-user-dir (Linux)
 
 WARNING
@@ -230,9 +230,9 @@ end
 
 local function create_kml_file(storage, image_table, extra_data)
 
-  local convertPath = dt.preferences.read("kml_export","convertPath","string")
-  if not df.check_if_bin_exists(convertPath) then
-    dt.print_error(_("convert not found"))
+  local magickPath = dt.preferences.read("kml_export","magickPath","string")
+  if not df.check_if_bin_exists(magickPath) then
+    dt.print_error(_("magick not found"))
     return
   end
   if dt.configuration.running_os == "linux" then
@@ -244,7 +244,8 @@ local function create_kml_file(storage, image_table, extra_data)
   dt.print_error("Will try to export KML file now")
 
   local imageFoldername
-  if ( dt.preferences.read("kml_export","CreateKMZ","bool") == true ) then
+  if ( dt.preferences.read("kml_export","CreateKMZ","bool") == true 
+       and dt.configuration.running_os == "linux") then
     if not df.check_if_bin_exists("zip") then
       dt.print_error(_("zip not found"))
       return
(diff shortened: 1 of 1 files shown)


BzKevin:
I don't use the map feature in dt at all, so I don't really know what exactly this is supposed to do or what I need (aside from ImageMagick) for this to work. If you explain a bit I can try to test sometime this week.

contrib/kml_export.lua
```diff
@@ -230,23 +229,16 @@ function addDuplicateIndex( index, filename )
 end
 
 local function create_kml_file(storage, image_table, extra_data)
-  if not df.check_if_bin_exists("mkdir") then
-    dt.print_error(_("mkdir not found"))
-    return
-  end
   if not df.check_if_bin_exists("convert") then
```
BzKevin: At first glance I suspect check_if_bin_exists('convert') will fail on windows. I have always had to put a file chooser in the preferences so the user can point to the program and then perform a dt.preferences.write('executable_paths', 'someprogram', dir) for check_if_bin_exists('someprogram') to work on windows. It seems as though when installing programs on windows they are not automatically visible to dt in the way they are on linux machines. 

@wpferguson please correct me if I'm wrong here as I'm not 100% certain. If there is a way around this I would love to know, as I'm currently trying to access dts exiv2 but can't figure out how.

contrib/kml_export.lua
```diff
@@ -230,23 +229,16 @@ function addDuplicateIndex( index, filename )
...
   end
-  if not df.check_if_bin_exists("xdg-open") then
-    dt.print_error(_("xdg-open not found"))
-    return
-  end
-  if not df.check_if_bin_exists("xdg-user-dir") then
-    dt.print_error(_("xdg-user-dir not found"))
-    return
+  if dt.configuration.running_os == "windows" then
+    if not df.check_if_bin_exists("xdg-user-dir") then
```
BzKevin: Similar to my comment above, my experience has been this will fail unless you provide the user a way to specifically define a directory for this and at some point do a dt.preferences.write.

contrib/kml_export.lua
supertobi: @BzKevin  sounds reasonable. I'll look into that and a preference.

supertobi:
This script creates an KML file (XML file for google earth) with the selected images. It works only with imagers, that have GPS information. 

I blogged some time ago about this script here.
https://dablogter.blogspot.com/2016/03/tutorial-creating-maps-from-geo-tagged.html

contrib/kml_export.lua
supertobi: this was stupid, the right syntax must be 
if dt.configuration.running_os == "linux" then

contrib/kml_export.lua
supertobi: I pushed a new version

supertobi:
The next step on my ToDo list is to refactor the preferences.

@wpferguson I had a change to test with Windows and it looks like we have a problem with dsys.launch_default_app(path). launch_default_app() adds double quotes around the filename and that does not work here.
`start c:\temp\test.jpg` works
`start "c:\temp\test.jpg"` does not work

wpferguson:
I almost pulled that back when I merged video_ffmpeg, because I thought that might happen.  I have a fix in mind, so let me code it, test it, and I'll push it in 10 or 12 hours.

wpferguson:
Fixed, pushed, and merged.

BzKevin:
I can do some testing over the weekend unless @wpferguson feels like doing it sooner. I'll need some files with GPS coordinates though if you can provide some @supertobi

supertobi:
@BzKevin Thanks, I'll add some example images and my last version tomorrow.

supertobi:
Here the test images:
https://drive.google.com/drive/folders/1SOA0Ok9r1cbm6Kc9ElNYsyb99PdYa5Id?usp=sharing

BzKevin:
Ok. I tested it. There is a small error I noted up already (an 'e' missing from line 96). It appears to work, however, ther amy be an issue. I am to unfamiliar to know. I opened the exported kml in Google Earth and the images thumbnails show up properly. However, when I click on an image I don't get a larger view, just a '?'

kmlexport example

contrib/kml_export.lua
```diff
@@ -263,8 +100,7 @@ local function create_kml_file(storage, image_table, extra_data)
 
     -- Creates dir if not exsists
     imageFoldername = "files"..PS
-    local mkdirCommand = "mkdir -p "..exportDirectory..PS..imageFoldername
-    dt.control.execute(mkdirCommand)
+    df.mkdir(df.sanitize_filename(xportDirectory..PS..imageFoldername))
```
BzKevin: you are missing an 'e' from 'xportDirectory'

supertobi: Thanks fixed

supertobi:
@BzKevin I fixed the bug you fund and now I think it works on Windows. (If PR #184 and #185 are merged.) Next step is to test if I it still works on Linux.

PR #172, opened 2019-02-26
fix lib/dtutils/system/open_default_app()

wpferguson:
Added a space after the commands and removed quotes around the path argument in the command invocation.

### DIFF
--- a/lib/dtutils/system.lua
+++ b/lib/dtutils/system.lua
@@ -120,13 +120,13 @@ dtutils_system.libdoc.functions["launch_default_app"] = {
   Copyright = [[]],
 }
 function dtutils_system.launch_default_app(path) 
-  local open_cmd = "xdg-open"
+  local open_cmd = "xdg-open "
   if (dt.configuration.running_os == "windows") then
-    open_cmd = "start"
+    open_cmd = "start "
   elseif  (dt.configuration.running_os == "macos") then
-    open_cmd = "open"
+    open_cmd = "open "
   end   
-  return dtutils_system.external_command(open_cmd..' "'..path..'"')
+  return dtutils_system.external_command(open_cmd .. path)
 end
 
 


lib/dtutils/system.lua
```diff
@@ -120,13 +120,13 @@ dtutils_system.libdoc.functions["launch_default_app"] = {
...
+  local open_cmd = "xdg-open "
   if (dt.configuration.running_os == "windows") then
-    open_cmd = "start"
+    open_cmd = "start "
   elseif  (dt.configuration.running_os == "macos") then
-    open_cmd = "open"
+    open_cmd = "open "
   end   
-  return dtutils_system.external_command(open_cmd..' "'..path..'"')
+  return dtutils_system.external_command(open_cmd .. path)
```
theres: @wpferguson path is not sanitized, this breaks Linux version when there is space in the path

wpferguson: The caller should send a sanitized path to the function, rather than the function assuming the path isn't sanitized.  Also windows and linux/macos have different quoting requirements.  lib/dtutils/file/sanitize_filename() is provided for that purpose.  The author of the script calling the function should know whether the path being passed needs to be sanitized or not and pass the argument appropriately.  Or, they can be safe and sanitize it anyway.  

Think of the libraries as little black boxes.  There is documentation about what the input should be and what output is expected.  So, I decide to use this function.  I pass it a path and sanitize it because it includes a space.  Unknown to me, unless I went in an looked at the code, the function attempts to sanitize my sanitized string.  The quotes cancel out and the command crashes because the string wasn't sanitized.  I would be very confused if this happened, because I knew that I had passed a sanitized string and my script crashed because the string wasn't sanitized.  And the solution wouldn't really make sense because I would have to not sanitize my string in order for it to be sanitized.

theres: For me both options (expect sanitized path or not) would work if it's covered in documentation, but I personally prefer to have this in the lib itself. It was my fault that it wasn't documented properly.  Anyway, seems like #173  fix problem in `video_ffmpg`.

I think, we should then remove sanitize from lib/dtutils/file/rm and lib/dtutils/file/mkdir to be consistent or change sanitize method to be idempotent (preferred).

PR #176, opened 2019-03-01
OpenInExplorer Tap to Space

supertobi:
Here I'll add the Nautilus to the script


supertobi:
As an bonus, you can now rename it to open_in_filemanager or something like this.

PR #178, opened 2019-03-04
Add AutoGrouper

BzKevin:
Adds AutoGrouper script which allows the user to select multiple images (or an entire collection) and automatically group them based on timestamp. The separation between groups is user defined via a slider.

### DIFF
--- a/contrib/AutoGrouper.lua
+++ b/contrib/AutoGrouper.lua
@@ -1,0 +1,155 @@
+--[[AutoGrouper plugin for darktable
+
+  copyright (c) 2019  Kevin Ertel
+  
+  darktable is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+  
+  darktable is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+  
+  You should have received a copy of the GNU General Public License
+  along with darktable.  If not, see <http://www.gnu.org/licenses/>.
+]]
+
+--[[About this Plugin
+This plugin adds the module "Auto Group" to darktable's lighttable view
+
+----REQUIRED SOFTWARE----
+None
+
+----USAGE----
+Install: (see here for more detail: https://github.com/darktable-org/lua-scripts )
+ 1) Copy this file in to your "lua/contrib" folder where all other scripts reside. 
+ 2) Require this file in your luarc file, as with any other dt plug-in
+
+Set a gap amount in second which will be used to determine when images should no 
+longer be added to a group. If an image is more then the specified amount of time
+from the last image in the group it will not be added. Images without timestamps 
+in exif data will be ignored.
+
+There are two buttons. One allows the grouping to be performed only on the currently
+selected images, the other button performs grouping on the entire active collection
+]]
+
+local dt = require "darktable"
+local mod = 'autogroup'
+local gettext = dt.gettext
+-- Tell gettext where to find the .mo file translating messages for a particular domain
+gettext.bindtextdomain("AutoGrouper",dt.configuration.config_dir.."/lua/locale/")
+local function _(msgid)
+    return gettext.dgettext("AutoGrouper", msgid)
+end
+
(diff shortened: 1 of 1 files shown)


contrib/AutoGrouper.lua
```diff
@@ -0,0 +1,155 @@
...
+longer be added to a group. If an image is more then the specified amount of time
+from the last image in the group it will not be added. Images without timestamps 
+in exif data will be ignored.
+
+There are two buttons. One allows the grouping to be performed only on the currently
+selected images, the other button performs grouping on the entire active collection
+]]
+
+local dt = require "darktable"
+local mod = 'autogroup'
```
wpferguson: 1. A style thing.  I would make mod MOD.  It's a carryover from C #defines where you define a constant.  In this case you're just setting the module name, but if it's all in caps then it flags people that it's probably a constant.

2. I would make autogroup autogrouper just to be consistent.

contrib/AutoGrouper.lua
```diff
@@ -0,0 +1,155 @@
...
+	if on_collection then 
+		local col_images = dt.collection
+		for i,image in ipairs(col_images) do --copy images to a standard table, table.sort barfs on type dt_lua_singleton_image_collection
+			table.insert(images,i,image)
+		end
+	else
+		images = dt.gui.selection()
+	end
+	dt.preferences.write(mod, 'active_gap', 'integer', GUI.gap.value)
+	if #images < 2 then return end
```
wpferguson: Maybe print a message to the screen (and/or log) about insufficient images instead of just silently doing nothing

contrib/AutoGrouper.lua
```diff
@@ -0,0 +1,155 @@
...
+	clicked_callback = function() main(false) end
+}
+GUI.collection = dt.new_widget("button"){
+	label = _('auto group: collection'),
+	tooltip =_('auto group the entire collection'),
+	clicked_callback = function() main(true) end
+}
+dt.register_lib(
+	"AutoGroup_Lib",	-- Module name
+	_("Auto Group"),	-- name
```
wpferguson: All the other module names are lower case including the first letter.

contrib/AutoGrouper.lua
```diff
@@ -0,0 +1,155 @@
...
+	selected = 		{},
+	collection = 	{}
+}
+temp = dt.preferences.read(mod, 'active_gap', 'integer')
+if not InRange(temp, 1, 3600) then temp = 3 end
+GUI.gap = dt.new_widget('slider'){
+	label = _('group gap [sec.]'),
+	tooltip = _('minimum gap, in seconds, between groups'),
+	soft_min = 1,
+	soft_max = 30,
```
wpferguson: Is this high enough?  Should it maybe be a minute or two, or five?  I tried it on a baseball game I shot and I ended up at 200 seconds.  I tried 300, but that was a little too much.  I also tried a 2 basketball games shot back to back.  With a 30 second gap I got 45 groups.  I think the right gap for that is probably 60 seconds, which would group between timeouts. Lastly I tried a senior picture session I did.  The 30 second grouping was just about right and the grouping was pretty close to what I did manually when I processed them.

BzKevin: The hard max is 3600 so the user can enter up to a full 1hr gap. I don't really know what would make a good soft max. 

My only thought was to avoid a frustration I have with some other sliders where the adjustments I am trying to make are difficult due to being stuck on the far end of the slider range. This happens to me frequently when using the right-click method for adjusting white balance, It's hard to make fine tune adjustments when lowering WB because I'm already so close to the left-hand boundary. 

That being said, I don't have a strong preference here, so if you all feel a higher soft max would be more usable by the majority of people then let's make it whatever you think is best.

wpferguson: I'm not sure either.  I toyed with the idea of 3 sliders, hours, minutes, and seconds, but I'm not sure that's any better.  How about we make it 60 seconds, and then see if the users provide any feedback.  Or, we can leave it as is and see if the user's have feedback.  Up to you.  Let me know.

BzKevin: 60 sounds good, a nice even 1 minute. I'll do that and take care of these other notes you've made when I get some time to tinker with these again. Glad you got some use out of it though.

BzKevin: I made it 60, I also increased the hard max all the way up to a full day. I saw no drawbacks to having a very large hard max.

PR #179, opened 2019-03-04
Refactored enfuseAdvanced

BzKevin:
I was having issue with this script periodically. The original version of this was real a hack-job of Holger Klemms original. This new version utilizes tables in a manner that I hope will make the code easier to follow and understand, but also allows for certain functions to be re-used throughout (and potentially in different scripts as well). In general:

Each program (align image stack, enfuse, exiftool) now has a table that contains information about that program, this includes a sub-table that contains argument data for the program.
There is a GUI table which contains all GUI elements. Those elements are split into sub-tables within GUI based on what aspect of the scrpt behavior they relate to (AIS, ENF, Target File, Presets, etc...)
These two structures (and the similarities between them) have allows me to create loops which populate the arguments from the GUI and vice-versa, this is all done through an intermediary preference, the "active" preference. 
Whenever a change to a GUI element occurs the associated member of hte active preference is updated. then when it becomes time to build an argument string for a program command the active preference is read from and the value contained within is used. This works fine for all GUI elements that have a "changed" or "clicked" callback, as they can remain up-to-date real time. Sliders and Entry boxes do not have this, so at a certain point withing the script the GUI values are read directly for these elements and written to the associated active preference field.

As you will see about 50% of this code is simply building all the GUI elements and making sure that the initialization value for that element is valid (via the InRange() function). The actual heart of the program and it's functions really only takes up about 300 lines (starting at 180 and going through 476). That is broken down into multiple fairly concise functions. Hopefully this makes maintenance, feature enhancement, and troubleshooting much easier in the future.

I also made a few UI tweaks. This script use to create such a massive list of GUI elements that it was a nuisance when actually using it. They are now grouped into three containers for AIS options, Enfuse options, and target options. Only one of these options containers is displayed at a time. Additionally, I removed all the registered preferences I had before, so now I only register 3 options to enable user to point towards the program binaries.

### DIFF
--- a/contrib/enfuseAdvanced.lua
+++ b/contrib/enfuseAdvanced.lua
@@ -45,7 +45,7 @@ If you have a specific set of parameters you frequently like to use, you can sav
 
 target file:
 Select your file destination path, or check the 'save to source image location' option.
-Unless 'Create unique filename' is selected, it will overwrite existing files
+'Create Unique Filename' is enabled by default at startup, the user can choose to overwrite existing
 Set any tags or style you desire to be added to the new image (only available if the auto-import option is enabled). You can also change the defaults for this under settings > lua options
 
 format options:
@@ -860,7 +860,7 @@ temp = dt.preferences.read(mod, 'active_on_conflict_ind', 'integer')
 if not InRange(temp, 1, 2) then temp = 1 end 
 GUI.Target.on_conflict = dt.new_widget('combobox'){
     label = _('on conflict'), 
-    selected = temp,  
+    selected = 1,  
     'create unique filename','overwrite',           
     changed_callback = function(self)
         dt.preferences.write(mod, 'active_on_conflict', 'string', self.value) 
--- a/contrib/enfuseAdvanced.lua
+++ b/contrib/enfuseAdvanced.lua
@@ -18,7 +18,7 @@
 ]]
 
 --[[About this plugin
-This plugin will add the new export module "fusion to DRI or DFF image".
+This plugin will add the new export module 'fusion to DRI or DFF image'.
    
 ----REQUIRED SOFTWARE----
 align_image_stack
@@ -27,12 +27,12 @@ exiftool
 
 ----USAGE----
 Install: (see here for more detail: https://github.com/darktable-org/lua-scripts )
- 1) Copy this file in to your "lua/contrib" folder where all other scripts reside. 
+ 1) Copy this file in to your 'lua/contrib' folder where all other scripts reside. 
  2) Require this file in your luarc file, as with any other dt plug-in
 On the initial startup go to darktable settings > lua options and set your executable paths and other preferences, then restart darktable
 
 DRI = Dynamic Range Increase (Blend multiple bracket images into a single LDR file)
-DFF = Depth From Focus ("Focus Stacking" - Blend multiple images with different focus into a single image)
+DFF = Depth From Focus ('Focus Stacking' - Blend multiple images with different focus into a single image)
 Select multiple images that are either bracketed, or focus-shifted, set your desired operating parameters, and press the export button. A new image will be created. The image will
 be auto imported into darktable if you have that option enabled. Additional tags or style can be applied on auto import as well, if you desire.
 
@@ -44,8 +44,8 @@ See enfuse documentation for further explanation of how it specifically works an
 If you have a specific set of parameters you frequently like to use, you can save them to a preset. There are 3 presets available for DRI, and 3 for DFF.
 
 target file:
-Select your file destination path, or check the "save to source image location" option.
-Unless "Create unique filename" is selected, it will overwrite existing files
+Select your file destination path, or check the 'save to source image location' option.
+Unless 'Create unique filename' is selected, it will overwrite existing files
 Set any tags or style you desire to be added to the new image (only available if the auto-import option is enabled). You can also change the defaults for this under settings > lua options
 
 format options:
@@ -56,15 +56,22 @@ Same as other export modules
 
 ]]
 
-local dt = require "darktable"
-local df = require "lib/dtutils.file"
-local dsys = require "lib/dtutils.system"
-local du = require "lib/dtutils"
+local dt = require 'darktable'
+local df = require 'lib/dtutils.file'
+local dsys = require 'lib/dtutils.system'
+local du = require 'lib/dtutils'
 local mod = 'module_enfuseAdvanced'
 local os_path_seperator = '/'
 if dt.configuration.running_os == 'windows' then os_path_seperator = '\\' end
 
-du.check_min_api_version("5.0.0", "enfuseAdvanced") 
+du.check_min_api_version('5.0.0', 'enfuseAdvanced')
+
+-- Tell gettext where to find the .mo file translating messages for a particular domain
+local gettext = dt.gettext
+gettext.bindtextdomain('enfuseAdvanced',dt.configuration.config_dir..'/lua/locale/')
+local function _(msgid)
+    return gettext.dgettext('enfuseAdvanced', msgid)
+end
 
 -- INITS --
 local AIS = {
@@ -166,9 +173,9 @@ local GUI = {
 }
 
 local styles = dt.styles
-local styles_count = 1 -- "none" = 1
+local styles_count = 1 -- 'none' = 1
 for _,i in pairs(dt.styles) do
-    if type(i) == "userdata" then styles_count = styles_count + 1 end
+    if type(i) == 'userdata' then styles_count = styles_count + 1 end
 end
 
 -- FUNCTION --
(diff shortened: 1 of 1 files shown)


wpferguson:
Why not add a fourth container for configuration and put the executable_path_preference widgets there so the use can just configure the program on the fly and not have to restart darktable.

contrib/enfuseAdvanced.lua
```diff
@@ -45,7 +45,7 @@ If you have a specific set of parameters you frequently like to use, you can sav
 
 target file:
 Select your file destination path, or check the "save to source image location" option.
-Unless "Create unique filename" is check, it will overwrite existing files
+Unless "Create unique filename" is selected, it will overwrite existing files
```
wpferguson: Should the default be non destructive, i.e. create unique filename, just in case the user isn't paying attention?  That way the user would have to intentionally select the destructive behavior.

BzKevin: I can change the comment to reflect the behavior of the script. You will see at lines 835-850 I attempt to load the widget with the most recently used value. If there is no value found it defaults to create unique. Also on a reset event it sets itself to create unique. 

I could have the element always load a create unique instead of last used. Thoughts on that? That would differ from the rest of the elements as I made everything load up with last used.

contrib/enfuseAdvanced.lua
```diff
@@ -54,548 +54,146 @@ Same as other export modules
...
+		optimize_image_center 	= {text = '-i', 			style = 'bool'},
+		auto_crop				= {text = '-C', 			style = 'bool'},
+		distortion				= {text = '--distortion', 	style = 'bool'},
+		gpu						= {text = '--gpu',			style = 'bool'},
+		grid_size				= {text = '-g ', 			style = 'integer'},
+		control_points			= {text = '-c ', 			style = 'integer'},
+		control_points_remove 	= {text = '-t ', 			style = 'integer'},
+		correlation				= {text = '--corr=', 		style = 'float'}
+	}
 }
```
wpferguson: Is your editor set to convert tabs to spaces?  I'm guessing these all line up when you're editing, but when they go to a different system they look like the above.  If your editor converts tabs to spaces, then they'll always line up on any system.

BzKevin: it is not, I can do so. Thanks for the heads up and sorry for the jumbled mess. Yes, it looks much cleaner in my editor :)

BzKevin:
> 
> 
> Why not add a fourth container for configuration and put the executable_path_preference widgets there so the use can just configure the program on the fly and not have to restart darktable.

True. I originally avoided doing that as it would add even more clutter to the UI. But now that I have containers it wouldn't really make any difference at all. I also made a comment of in the HDRMerge PR that applies here:
I'm not sure when I need to ask the user for a bin path and when not to. Right now I always ask. I know Windows needs it. and my understanding is that linux doesn't need that at all? Do you know if macOS needs user-defined executable preferences or not? I could only display the file choosers on systems that need them.

contrib/enfuseAdvanced.lua
```diff
@@ -54,548 +54,146 @@ Same as other export modules
...
-	enf_chkbtn_cpytags,
-	enf_entry_tag,
-}
+end
 
--- FUNCTION -- 
-local function GetFileName(full_path)
-	--[[Parses a full path (path/filename_identifier.extension) into individual parts
-	Input: Folder1/Folder2/Folder3/Img_0001.CR2
+local function GetFileName(full_path) -- Parses a full path (path/filename_identifier.extension) into individual parts
```
wpferguson: There's already path parsing functions in the file library including split_filepath which returns path, filename, basename, and extension

BzKevin: I thought I looked at incorporating that the first time 'round and it didn't quite fit my needs. I could be wrong. I'll look again.

contrib/enfuseAdvanced.lua
```diff
@@ -612,434 +210,851 @@ local function GetFileName(full_path)
...
+
+local function PreCall(prog_tbls) --looks to see if this is the first call, if so checks to see if program is installed properly
+	for _,prog in pairs(prog_tbls) do
+		if prog.first_run then
+			prog.bin = df.check_if_bin_exists(prog.name)
+			if not prog.bin then 
+				prog.install_error = true
+			else
+				prog.bin = CleanSpaces(prog.bin)
+			end
```
wpferguson: Have you had a problem with executable paths having spaces?  They are stored in the darktablerc file without leading or trailing space, so I don't think you need this.

contrib/enfuseAdvanced.lua
```diff
@@ -612,434 +210,851 @@ local function GetFileName(full_path)
...
+	GUI.AIS.optimize_field,
+	GUI.AIS.optimize_image_center,
+	GUI.AIS.auto_crop,
+	GUI.AIS.distortion,
+	GUI.AIS.gpu,
+	GUI.AIS.grid_size,
+	GUI.AIS.control_points,
+	GUI.AIS.control_points_remove,
+	GUI.AIS.correlation
+}
```
wpferguson: Did you try doing table.unpack(GUI.AIS)?

contrib/enfuseAdvanced.lua
```diff
@@ -612,434 +210,851 @@ local function GetFileName(full_path)
...
+	GUI.Target.apply_style,
+	GUI.Target.copy_tags,
+	GUI.Target.add_tags
+}
+GUI.options_contain = dt.new_widget('stack'){
+	active = 2,
+	box_AIS,
+	box_ENF,
+	box_Target
+}
```
wpferguson: You could add a configuration stack for the executables, and if it's the first run then make that the default.

wpferguson:
I'm going to download the source file and go thru it and add comments.  It's really hard to get a good grip on what you've done in the diff view provided here.  What I can see looks good.  I think you might be able to take it a little farther and specify your gui elements in the tables and just call a routine to build them, but I need to see the script as a whole to figure it out.

wpferguson:
You might want to look at contrib/video_ffmpeg.lua.  @theres  had some preference read and write functions that looked pretty slick and might be useful.  There's also a placeholder replacement function so that you could replace {datatype} with bool, string, int, etc.

contrib/enfuseAdvanced.lua
BzKevin: That's a great idea. Then if it detects an install error I can make that container active in the stack.

contrib/enfuseAdvanced.lua
BzKevin: No. I don't know what that is. I'll have to look it up

contrib/enfuseAdvanced.lua
BzKevin: No I don't think I have. I did this just as a double check. The CleanSpaces() function is needed elsewhere, so since I already had it I figured I'd just be safe and run the binaries through it too.

wpferguson:
Linux can determine executable paths using which, and it works.  MacOS needs to specify the location to be safe, and windows needs to specify the location. So,

if dt.configuration.running_os ~= "linux" then
    -- create 4th stack for configuration
end

contrib/enfuseAdvanced.lua
BzKevin: I looked again and split_filepath  doesn't quite do the samething as what I do here. I could use it, but then I'd have to do further processing on the result, so I would still have to have a separate function to return exactly what I'm looking for (I return the strictly numerical part of the image name as the 'ID'. Here you can see what is returned from each and how they differ:

--Split Filename--
C:\Users\ERTELK\AppData\Local\Temp\
_MG_4839.tif
_MG_4839
tif

--Get Filename--
C:\Users\ERTELK\AppData\Local\Temp\
_MG_4839
4839
.tif

contrib/enfuseAdvanced.lua
BzKevin: I uploaded a new version converted tab to space

contrib/enfuseAdvanced.lua
BzKevin: I just made a commit that makes 'create unique' the default selection at startup

contrib/enfuseAdvanced.lua
BzKevin: I'll look into the unpack function. However, things are working as-is, and this is already a big step forward on refactoring this, so let's not let table.unpack implementation hold up this PR. If I can get around to implementing it in the future I can do so.

supertobi:
Is this version better then the old one and does it run on all OS? If yes, lets merge it. It can still be improved later.

BzKevin:
> 
> 
> Is this version better then the old one and does it run on all OS? If yes, lets merge it. It can still be improved later.

This does run on all OS, though the old one did too. However, I experienced some glitchy behavior with the old one that I have not had since refactoring. The main goals of refactoring were:
1) Make the code easier to understand, thus more maintainable by myself and others
2) Per @wpferguson it is a good idea to minimize usage of the preferences, so while the old script relied on that heavily, this does not at all
3) I had UX (user experience) problems with the old script as it generated one massive list of options, this has been optimized by grouping related options together and only displaying a single group at a time
4) This is now translatable, previously was not

PR #180, opened 2019-03-13
Add support for windows WHERE function to detect binaries referenced …

Varamil:
Add support for windows WHERE function to detect binaries referenced by Path environment variable.

Most of lua script calls check_if_bin_exists with only binary's name, which always fail on Windows as was coded the function. This patch use the Windows equivalent of unix which.

### DIFF
--- a/lib/dtutils/file.lua
+++ b/lib/dtutils/file.lua
@@ -45,9 +45,11 @@ dtutils_file.libdoc.functions["check_if_bin_exists"] = {
   Description = [[check_if_bin_exists checks to see if the specified binary exists.
     check_if_bin_exists first checks to see if a preference for the binary has been
     registered and uses that if found.  The presence of the file is verified, then 
-    quoted and returned.  If no preference is specified and the operating system is
-    linux then the which command is used to check for a binary in the path.  If found
-    that path is returned.  If no binary is found, false is returned.]],
+    quoted and returned.  If no preference is specified, if operating system is 
+    linux then the which command is used to check for a binary in the path, else if
+	operating system is windows the WHERE command is used to check if the binary is 
+	referenced in Path. If found that path is returned.  If no binary is found, 
+	false is returned.]],
   Return_Value = [[result - string - the path of the binary, false if not found]],
   Limitations = [[]],
   Example = [[]],
@@ -82,6 +84,13 @@ function dtutils_file.check_if_bin_exists(bin)
     if string.len(output) > 0 then
       result = output:sub(1,-2)
     end
+  elseif dt.configuration.running_os == "windows" then
+	local p = io.popen("WHERE " .. bin)
+    local output = p:read("*a")
+    p:close()
+	if string.match(output, "\\") then
+	  result = output
+	end
   end
   return result
 end


wpferguson:
On Linux binaries are installed in the path.  On macos binaries are sometimes installed in the path.  On windows binaries are never installed in the path.  Windows insists that binaries be installed in \Program Files\package\... or \Program Files(X86)\package\... which aren't in the path unless you add them.  So therefore using the where command on windows is a waste of time.  

check_if_bin_exists() first checks for an executable_path preference for the binary.  If that's found, that's what gets used.  If windows is the OS, then that's where the processing stops and we return not found (nil).  On macos and linux we try the where command.  It works on linux if the executable is installed and works on macos sometimes if the executable is installed in the path.  The only reliable way we've found for locating binaries on windows is using the executable_path preference and having the user set the location.  I thought about trying to use find and searching the disk, but then we run into the different possible names for an executable.  For instance searching for gimp.exe doesn't find Gimp-2.10.exe or gimp_2_8_2.exe.  We could try wildcards but then we might end up with gimp-batch.exe or gimp-debug-resume.exe, which isn't correct either.  The executable naming isn't consistent enough to search on.  I've tried a bunch of ways to make it work on windows, but so far the only method that works RELIABLY is the executable_path preferences.

Varamil:
Sure, this patch won't work in any case, but adding this modification will avoid preference management for all scripting language (like Python) for which Path is often set. By the way it doesn't replace executable_path preference, so if WHERE returns something you save time and configuration requirement. 

Sincerely, I'm new in darktable, and I've lost a lot of time to understand why official lua scripts don't work ! I haven't even found out where I can set this executable_path preference (if you can help ? :-) )! So I think this trick should help beginners.

wpferguson:
If you go to https://github.com/darktable-org/lua-scriptsand read the README.md file that displays on that page, you'll find detailed instructions on how to install the scripts and enable them.  There is also a statement that says read the start of each script for instructions on how to use it.  In there it tells you how to set the executable preference.  Usually it's a file selector widget in the GUI.  Sometimes it's in the global settings.  The only change I might suggest is just putting require "tools\\script_manager" in the luarc, then using the script manager module that shows up in the bottom left pane of the lighttable UI.  If you hover over an enable button, the comments on how to use the script are displayed.

If you had read farther down in file.lua, you would have seen the executable preference code and could have seen how it worked.  Then you could have searched to see where it was used and observe how it was used.

darktable is not just a piece of software.  It's a community of users, developers, educators, and others.  To participate in the community, you have to educate yourself.  The darktable website, https://darktable.org, has resources that will help.  I've already mentioned the lua-scripts readme. There are mailing lists, darktable-user and darktable-dev described on the darktable website.  There is a darktable web page.  There are multiple discussion threads on discuss.pixls.us, including many about how to use the lua scripts.

If, after you've read all the documentation, you still can't figure it out, then use the mailing lists and/or discuss.pixls.us.

The one thing you probably shouldn't do, ever, is submit code that you know doesn't work for inclusion in the project, especially if you don't understand the project.  Take the time to get to know the code and use it, before you try to change it.  And if you do make a change, make sure it improves/solves a problem and actually works.  Otherwise you're just wasting peoples time.  Please be considerate.

Varamil:
For your information I've tried to use script_manager, but if you check the code, oh surprise there is the function _check_if_bin_exists_ called on git, which fail on windows by default. Then I've tried again this script_manager after my modification and, oh miracle this script was executed well without specifying anything about git location, because... git is referenced in Path !

When you say that I should have read farther down file.lua script, you speak about _set_executable_path_preference_ function ? Seriously ? You have to read all scripts of a lambda sub library to be able to run a script that is designed for people doesn't code ? It's your plug'n play definition ? 

When I say I've lost a lot of time because of this function, it wasn't looking my nails waiting for illumination. I've read and read again the documentation, forums, etc. And it's starting from hello word script and adding debug trace to scripts that I've seen this function is the problem and I can improve it. 

Actually I don't understand your problem. If something can't be white so it has to be black ? Grey is
not a solution ? What's the problem using WHERE ? How adding this feature makes the function less reliable ?  The new code is executed only when result is _false_ AND only on windows, it's just a last try ! You say WHERE is a waste of time, for who ? 1ms shorter to say "bin not found" versus "don't care about, git (Python, ...)  is in Path, I manage that", hum... hard choice...

You say darktable is a community, I hope that all other aren't like you. For me, a community is share and mutual support, not "don't touch, it's my business". I've started to write several script for darktable, but I guess I'll let recode them, I don't want to fight each time. 
Well, after all I've shared my experience, so, now do what you want with this code, I don't care. It's not shared as I expected, but it's on Internet, so it can help some people, it will be just harder to find.

wpferguson:
The beauty of open source is that if you don't like the way it's done, fork it and do it your own way.  As for black or white, it has to work every time or fail in a predictable way.  Where works sometimes, and doesn't work sometimes.  Sometimes it returns the wrong executable.  Therefore it doesn't fail predictably.  I'm glad it works for you.  However, I support this code and work the issues and gray (sometimes it works, sometimes it doesn't) doesn't work for me.

wpferguson:
I tried to recreate the problem with script_manager.  I removed git from my windows system and script_manager still starts and manages the scripts.  It doesn't need git to run, just to download or update the scripts.

wpferguson:
@Varamil, you asked what my problem was.  My problem was that I got VERY confused with the whole interaction.

An example: You submitted a pull request to change the software.  To me, this meant you had tested the software and it worked.  Then you said "Sure, this patch won't work in any case", which means to me that it doesn't work, so I wondered why you submitted it.  Then you said "Grey is not a solution" which means, to me, that it works part of the time.  At this point I'm totally confused because my understanding of what you said, by actions and words, was the code did/did not/sometimes work.

This interaction didn't work for you and left you unsatisfield.  Would you tell me how you think it should have worked?

PR #181, opened 2019-03-16
Copy dtMediaWiki v44 from trougnouf/dtMediaWiki

trougnouf:
dtMediaWiki adds Wikimedia Commons as a target storage. Latest documentation on https://commons.wikimedia.org/wiki/Commons:DtMediaWiki


wpferguson:
@supertobi I think @trougnouf has just triggered the next step(s) in the evolution of the lua-scripts repository.  Here's what I think needs addressed:

1. luarocks dependency.  Currently we have one script that has a requirement for an external library and we've just kind of ignored it and left it up to the user to resolve. I've avoided using external lua dependencies because I could and I didn't want to figure it out at the time.  I think the time has come to address it.

2. script documentation.  Currently we don't really have a good method of addressing script documentation.  I rip the beginning comments from the scripts to use in the tooltips for script_manager, but that's the only script documentation that's available.  Perhaps markdown formatted files in a doc directory?

3. scripts in subdirectories.  I'm currently working on the plugin API, which envisions each plugin in it's own directory.  However, the plugin API is nowhere close to ready.  A script in a subdirectory will probably break script_manager at the worst, or at the best it will just get ignored.  One solution is to split this up and put the README.md in a doc directory, the MediaWikiAPI.lua in the lib directory, and dtMediaWiki.lua in the contrib directory.  Or, I can change script manager to correctly handle scripts in subdirectories.  I guess the real question is what do we want the repository tree to look like?

I propose opening issues for each of the above so that we can address them in their own thread.

@trougnouf could you point us at a page that has been populated using this script so that we can see what the output looks like? Thanks

trougnouf:
@wpferguson  All files uploaded with the plugin are added to [Uploaded_with_dtMediaWiki](https://commons.wikimedia.org/wiki/Category:Uploaded_with_dtMediaWiki). [File:SNCB_978_leaving_the_Ronet_classification_yard_(DSCF5536).jpg](https://commons.wikimedia.org/wiki/File:SNCB_978_leaving_the_Ronet_classification_yard_(DSCF5536).jpg) probably uses the latest revision (and makes uses the "other versions" feature).

supertobi:
1) I'm not a really big fan of all the script repositories (luarocks, CPAN, note.js, ...) not managed from the standard system way to install software. On the other hand it brings a lot of power and we should not block them.
2) Subfolder or document folder?
    I like subfolders more. Have you tested it with your script manager?

contrib/dtMediaWiki/dtMediaWiki.lua
```diff
@@ -0,0 +1,400 @@
...
+    Author: Trougnouf (Benoit Brummer) <trougnouf@gmail.com>
+    Contributor: Simon Legner (simon04)
+
+Dependencies:
+* lua-sec: Lua bindings for OpenSSL library to provide TLS/SSL communication
+* lua-luajson: JSON parser/encoder for Lua
+* lua-multipart-post: HTTP Multipart Post helper
+]]
+local dt = require "darktable"
+local MediaWikiApi = require "contrib/dtMediaWiki/mediawikiapi"
```
wpferguson, line 11: Please move mediawikiapi.lua into a lib subdirectory under the dtmediawiki directory.  script_manager ignores files in .../lib/... directories, figuring they are libraries and doesn't create a start button for them.

trougnouf: Done, thank you !

contrib/dtMediaWiki/dtMediaWiki.lua
```diff
@@ -0,0 +1,400 @@
...
+
+Dependencies:
+* lua-sec: Lua bindings for OpenSSL library to provide TLS/SSL communication
+* lua-luajson: JSON parser/encoder for Lua
+* lua-multipart-post: HTTP Multipart Post helper
+]]
+local dt = require "darktable"
+local MediaWikiApi = require "contrib/dtMediaWiki/lib/mediawikiapi"
+local version = 47
+
```
wpferguson, line 13: Translation would be nice.  examples/gettextExample.lua shows how

trougnouf: Will do, but it might take a little while because I currently spend my free time getting a boat/radio/motorcycle license.

trougnouf: @wpferguson done :-)

trougnouf: Actually this crashes when I use the "_" function inside one of the functions registered in the dt.register_storage function, namely register_storage_initialize and register_storage_finalize (although adding a print(_()) statement in  register_storage_supported is an easy way to test).
Any idea what went wrong?

trougnouf:
Thanks!

trougnouf:
Any feedback as to why this was rejected?

wpferguson:
> Any feedback as to why this was rejected?

It had the label waiting for feedback and it had been almost 4 years so I assumed that you had lost interest (and I finally got around to some house cleaning).  Now that I opened it, I see where you pushed some commits, so my bad for assuming.

I'll reopen it and we'll see if we can get it finished.

trougnouf:
> > Any feedback as to why this was rejected?
> 
> It had the label waiting for feedback and it had been almost 4 years so I assumed that you had lost interest (and I finally got around to some house cleaning). Now that I opened it, I see where you pushed some commits, so my bad for assuming.
> 
> I'll reopen it and we'll see if we can get it finished.

Indeed I think I addressed everything I can at the time.

I just pushed an update to the latest commit from my repo.

PR #182, opened 2019-03-21
Bz kevin open in prog 1

BzKevin:
Add OpenInProg script, modify system utility as needed to support "Open Each" behavior

Script adds a configurator that enables the user to configure multiple programs, and specify what controls should be generated for each program:
Open All - Attempts to open all selected Raw files in the program in a single call
Open Each - Attempts to open each of the selected Raw files in the program with a unique call per image
Export All - Exports images to selected folder then attempts to open them all in the program

All configuration data is saved real-time and loaded at startup. More programs can be added and controls for the new programs generated via the configurator, so a restart of dt is not necessary. Programs can be removed and they will not be auto-loaded on the next startup.

Open Each behavior required modifications to the system file. dsys.external_command() now takes an optional second parameter that enables 'no_wait' behavior. The batch file is modified to exit immediately after executing the specified command, thus allowing multiple back-to-back commands to fire. This "no_wait" behavior was only implemented for windows based systems, though it was done in such a way that all existing usages of dsys.external_command() are unaffected by this change.

### DIFF
--- a/contrib/OpenInProg.lua
+++ b/contrib/OpenInProg.lua
@@ -1,0 +1,456 @@
+--[[OpenInProg plugin for darktable
+
+  copyright (c) 2019  Kevin Ertel
+  
+  darktable is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+  
+  darktable is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


BzKevin:
Potential Issues:
1) "Open Each" behavior required modifications to system file. I did this only for Windows, so "Open Each" may not work as desired on MacOS and Linux, I don't know. 
2) Not translatable at the moment.
3) Currently does not support user-entered arguments

PR #184, opened 2019-04-15
On Windows we don't need anything to open a file

supertobi:
On Windows we don't need any command to open a file. 
And "start" has problems with spaces in the filename, even if we put quoter around them:
https://stackoverflow.com/questions/13691827/opening-file-with-spaces-in-windows-via-command-prompt


supertobi:
@wpferguson please review and push if OK.

wpferguson:
Works fine, merging...

PR #185, opened 2019-04-15
file.lua copy for windows

supertobi:
Could be improved, but at least it works for me.

### DIFF
--- a/lib/dtutils/file.lua
+++ b/lib/dtutils/file.lua
@@ -291,9 +291,13 @@ dtutils_file.libdoc.functions["file_copy"] = {
 function dtutils_file.file_copy(fromFile, toFile)
   local result = nil
   -- if cp exists, use it
-  if dtutils_file.check_if_bin_exists("cp") then
+  if dt.configuration.running_os == "windows" then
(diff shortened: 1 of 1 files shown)


supertobi:
@wpferguson Please test and merge if OK.

wpferguson:
I'll build a test suite for this and the other PR.  Should have it done tonight.

lib/dtutils/file.lua
```diff
@@ -291,9 +291,13 @@ dtutils_file.libdoc.functions["file_copy"] = {
 function dtutils_file.file_copy(fromFile, toFile)
   local result = nil
   -- if cp exists, use it
-  if dtutils_file.check_if_bin_exists("cp") then
+  if dt.configuration.running_os == "windows" then
+    os.execute('copy "' .. fromFile .. '" "' .. toFile .. '"')
+    result = true
```
wpferguson: this should be result = os.execute also.  It returns nil on failure and true on success, so we can use the windows copy command and if it fails for some reason then fall back to the pure lua copy solution.

supertobi: done

PR #190, opened 2019-06-23
mkdir/rmdir : handle path with spaces

AlicVB:
be careful, I've absolutely no idea if this break something on other platforms than linux

### DIFF
--- a/lib/dtutils/file.lua
+++ b/lib/dtutils/file.lua
@@ -579,7 +579,7 @@ dtutils_file.libdoc.functions["mkdir"] = {
 function dtutils_file.mkdir(path) 
   if not dtutils_file.check_if_file_exists(path) then
     local mkdir_cmd = dt.configuration.running_os == "windows" and "mkdir" or "mkdir -p"
-    return dsys.external_command(mkdir_cmd.." "..path)
+    return dsys.external_command(mkdir_cmd.." \""..path.."\"")
   else
     return 0
   end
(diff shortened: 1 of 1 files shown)


theres:
IIRC It was discussed before and you should pass sanitized paths instead. There is function in file utils called `sanitize_filename()` that is supposed to work on all systems. But bear in mind that it is not idempotent.

 Now it may break some scripts, for example `contrib/video_ffmpeg`, not sure if any other also use these methods.

AlicVB:
ohh, no pb. I don't want to break anything.

I'm not sure to fully understand your point as I can't see any use case where not adding quotes can be beneficial, and as other file operations which use system calls add quotes already.
But I'm far from lua specialist !

Just say if you want that I close this issue !

wpferguson:
There are several library functions that act on files and directories.  Windows and Linux/MacOS use different quoting rules.  It's safer, and easier :-), to provide a function, sanitize_filename(), that does the correct thing each time and pass the result of that to the functions.  This way the developer has control of where and when to sanitize.  Perhaps we should update the documentation of the file/directory functions to remind developers that they need to pass sanitized paths.

IIRC there is an executable that doesn't like quoted filenames, but I can't remember what it was. 

As I was writing this I came up with several possibilities.  We could...

1. Leave everything as is, and maybe update the documentation.

2.  Make sanitize_filename() smarter so that it doesn't sanitize a sanitized filename (thus unsanitizing it) and have every file/directory function call it to ensure the path is sanitized.

3. We could add code to check each path in each function and sanitize as necessary.

I'm good with 1, and I can do 2, but not 3.

@AlicVB  @theres  @BzKevin @supertobi I think we should close the PR and open an issue if we want to discuss this further to decide on a course of action

wpferguson:
Closing the PR and opening an issue

PR #192, opened 2019-07-06
String Library Unit Testing

wpferguson:
Made sanitize() function idempotent.  Created funtion is_not_sanitized() to check if a string is sanitized.  Fixed urlencode() to properly encode strings without spaces.  Created test directory and added unit testing.  LuaUnit is used as the test engine.  Unit tests are built for the string library.


wpferguson:
Pulled out the corrected libraries and submitted them as a separate PR.  I'll close this and maybe submit it as a test framework at a later date when it's more complete.

PR #193, opened 2019-10-09
README: list various third-party projects

simon04:


supertobi:
@BzKevin 
Are this two lines still different from the version here?
* [BzKevin/HDRMerge-dt-Plugin](https://github.com/BzKevin/HDRMerge-dt-Plugin) – HDRMerge
* [BzKevin/OpenInExplorer-darktable-plugin](https://github.com/BzKevin/OpenInExplorer-darktable-plugin) – open in file explorer

wpferguson:
A couple of thoughts.

1.  Is this the right place for this?  I can't think of another place that keeps track of lua scripts for darktable, so maybe it is.  The only other place I could think of would be pixls.us, but it would probably get lost in all the messages.

2.  Do we need a disclaimer statement saying that this is provided for information only, we haven't tested them, we aren't blessing them (in light of the OpenInPS script), use at your own risk, YMMV?

supertobi:
@wpferguson 
1. Yes, I think it's the right place. 
2. Yes, I think we need a disclaimer.

@simon04 
Please update your patch with the changes @wpferguson suggested.

simon04:
(Sorry for the long delay – I was travelling.)

@wpferguson, @supertobi, thanks for your review. I amended my commit with your comments and included a disclaimer line.

PR #197, opened 2019-12-09
Update face recognition

wpferguson:
Converted face_recognition to a module and updated it to use newer library functions.  Added tolerance adjustment to settings to fine tune the face detection.

### DIFF
--- a/contrib/face_recognition.lua
+++ b/contrib/face_recognition.lua
@@ -41,11 +41,21 @@ This plugin will add a new storage option and calls face_recognition after expor
 ]]
 
 local dt = require "darktable"
+local du = require "lib/dtutils"
 local df = require "lib/dtutils.file"
 local gettext = dt.gettext
 
--- works with darktable API version from 2.0.0 to 5.0.0
-dt.configuration.check_version(...,{2,0,0},{3,0,0},{4,0,0},{5,0,0})
+-- constants
+
+local MODULE = "face_recognition"
+local OUTPUT = dt.configuration.tmp_dir .. "/facerecognition.txt"
+
+-- namespace
+
+local fc = {}
+
+-- ensure we meet the minimum api
+du.check_min_api_version("5.0.0", "face_recognition")
 
 -- Tell gettext where to find the .mo file translating messages for a particular domain
 gettext.bindtextdomain("face_recognition", dt.configuration.config_dir.."/lua/locale/")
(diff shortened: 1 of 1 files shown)


contrib/face_recognition.lua
```diff
@@ -41,11 +41,21 @@ This plugin will add a new storage option and calls face_recognition after expor
...
+local du = require "lib/dtutils"
 local df = require "lib/dtutils.file"
 local gettext = dt.gettext
 
--- works with darktable API version from 2.0.0 to 5.0.0
-dt.configuration.check_version(...,{2,0,0},{3,0,0},{4,0,0},{5,0,0})
+-- constants
+
+local MODULE = "face_recognition"
+local OUTPUT = dt.configuration.tmp_dir .. "/facerecognition.txt"
```
supertobi: Is the tmp dir the right one, or is the facerecognition.txt used longer?
Do we need the system independent path separator here?:
local PS = dt.configuration.running_os == "windows" and "\\" or "/"

contrib/face_recognition.lua
```diff
@@ -54,39 +64,101 @@ local function _(msgid)
...
+  exporter.max_width = width
+
+  -- export the images
+  local job = dt.gui.create_job("export images", true, stop_job)
+  local exp_cnt = 0
+  local percent_step = 1.0 / images
+  job.percent = 0.0
+  for img,export in pairs(img_tbl) do
+    exp_cnt = exp_cnt + 1
+    dt.print(_("Exporting image") .. exp_cnt .. _(" of ") .. images)
```
supertobi: We should use string.format(_("Exporting image %i of %i images."), exp_cnt, images) here

contrib/face_recognition.lua
wpferguson: The facerecognition.txt file is used to output the results from face_recognition.  After it's populated, it's read to figure out what tags to apply.

When I read https://github.com/ageitgey/face_recognition/README.md it says linux and macos work and windows might, so I didn't work on adding windows compatibility.  I can go ahead and add it just in the face recognition software can be made to work on windows.  I'll add it and push it tonight.

contrib/face_recognition.lua
wpferguson: Yes, that's a much cleaner way to write that...

PR #204, opened 2019-12-20
Add ext_editor.lua to contrib

Mark-64:
The script contains utilities to export pictures to collection and launch external programs for further edits.
The scripts adds:
* a new target storage "collection". Images exported will be reimported to collection for		further edit with external programs
Cattura2
* a new lighttable module "external editors", to select a program from a list of up to 9 external editors and run it on one selected image
Cattura1
* a set of lua preferences in order to configure name and path for up to 9 external editors
Cattura3
* a set of lua shortcuts in order to quick launch the external editors
Cattura4

**Limitations: only tested on Windows, DT 3.0 RC2**

More info in the file header. 
Hope it can be useful, comments welcome.

Marco


Mark-64:
The script now work also without leaving darkroom mode, which is very handy.
Just configure export destination to "collection" before entering darkroom.
From there, CTRL+E will create a new image for external edit. 
To edit an image from darkroom, configure and use shortcuts. The image view will be updated when leaving the external program.
The script is now tested in Windows 10 and Ubuntu 19.10

Mark-64:
I made the requested changes for sanitize_filename and README, plus some changes in text strings to improve readability.
When coming to translations, I've tried, but I could't make it work, need more time to study.
Sorry if I messed up with github, I'm not very expert of it.
Marco

supertobi:
No problem. I'll merge this now and you can make more PRs later.

wpferguson:
Sorry I'm late to the party... :)  @Mark-64 is your workflow primarily jpg based, i.e. you shoot primarily jpgs?  This script makes sense (to me) in a jpg based workflow.  It doesn't make sense (to me) in a raw based workflow.

I loaded this up and tested it against my play raw (from pixls.us) folder.  I set up 4 executables as follows:

non existent
/path/to/something/that/doesnt/exist

Color Efex 4
~/.wine/drive_c/Program Files/Google/Nik Collection/Color Efex Pro 4/Color Efex Pro 4 (64-Bit)/Color Efex Pro 4.exe

GIMP
/usr/bin/gimp

krita
/usr/bin/krita

I used a DNG file (20161225_135912_239.DNG) for all the tests.  

Test 1: I ran non existent and the DNG file disappeared from the collection.  Fortunately darktable.image.delete() just removes the image instead of trashing it.  I got the image back by reimporting the folder.  This failed because there was no executable /path/to/something/that/doesnt/exist, so the script tried to run the executable, failed, then deleted the file expecting to reimport it, but wasn't able to.

Test 2 also failed because the executable needed to run under wine with the same result when deleting the file and trying to reimport it.

Tests 3 and 4 at least got the file into the editor because they can both handle raws.  However saving the file didn't save it to the collection.

The test results reinforce my thought that this will work well in a jpg workflow, but not in a raw workflow.  I think we need to say in the desciption and in the README.md that this is only recommended for a jpg workflow.

There also needs to be checks to see if the specified executable exists.  The results from the command also need checking to see if the command succeeded or not.  The file path in the preferences should be a file chooser button so the user can just click on the file they want to run instead of typing in the path which is error prone.

I'll go through the code and add some recommendations.

contrib/ext_editor.lua
```diff
@@ -0,0 +1,371 @@
...
+
+-- callback for button "edit"
+local function OpenWith(images, choice)
+		
+	if #images == 1 then
+		for _, image in pairs(images) do 
+			
+			-- image to be edited
+			name = image.path..PS..image.filename 					
+
```
wpferguson: you could add a check here to see if the image is raw or not and act accordingly.  You can check if image.is_ldr is true (jpg, tif, etc) or not (raw).

contrib/ext_editor.lua
```diff
@@ -0,0 +1,371 @@
...
+local function OpenWith(images, choice)
+		
+	if #images == 1 then
+		for _, image in pairs(images) do 
+			
+			-- image to be edited
+			name = image.path..PS..image.filename 					
+			
+			-- launch the external editor
+			local run_cmd = df.sanitize_filename(program_paths[choice]).." "..df.sanitize_filename(name)
```
wpferguson: You should check and see if program_paths[choice] is actually and executable or not, just in case someone made a typo when they typed in the path, or a file got moved, etc.  df.check_if_bin_exists() returns the sanitized executable command (i.e. prefixes wine when trying to run a .exe file on linux or macos).

contrib/ext_editor.lua
```diff
@@ -0,0 +1,371 @@
...
+		for _, image in pairs(images) do 
+			
+			-- image to be edited
+			name = image.path..PS..image.filename 					
+			
+			-- launch the external editor
+			local run_cmd = df.sanitize_filename(program_paths[choice]).." "..df.sanitize_filename(name) 
+			dt.print("Launching "..program_names[choice].."...")	
+			dtsys.external_command(run_cmd)
+
```
wpferguson: You can check the result of the dtsys.external_command to see if it succeeded or not and then act accordingly.

contrib/ext_editor.lua
```diff
@@ -0,0 +1,371 @@
...
+			local run_cmd = df.sanitize_filename(program_paths[choice]).." "..df.sanitize_filename(name) 
+			dt.print("Launching "..program_names[choice].."...")	
+			dtsys.external_command(run_cmd)
+			
+			-- refresh the image view
+			-- note that only image:drop_cache() is not enough to refresh view in darkroom mode
+			-- therefore image must be deleted and reimported to force refresh
+			
+			local image_leader = image.group_leader
+			image:delete ()
```
wpferguson: Make sure the edit was successful before doing the delete/import.

contrib/ext_editor.lua
```diff
@@ -0,0 +1,371 @@
...
+			if string.match(df.get_basename(new_name), "_(d-)$") == "99" then
+					break
+				end
+				
+			-- physical copy
+			local result = df.file_copy(name, new_name)
+			
+			if result then
+				-- launch the external editor
+				local run_cmd = df.sanitize_filename(program_paths[choice]).." "..df.sanitize_filename(new_name)
```
wpferguson: Same as earlier, make sure the executable exists

contrib/ext_editor.lua
```diff
@@ -0,0 +1,371 @@
...
+				end
+				
+			-- physical copy
+			local result = df.file_copy(name, new_name)
+			
+			if result then
+				-- launch the external editor
+				local run_cmd = df.sanitize_filename(program_paths[choice]).." "..df.sanitize_filename(new_name)  	
+				dt.print("Launching "..program_names[choice].."...")	
+				dtsys.external_command(run_cmd)
```
wpferguson: and check the result

contrib/ext_editor.lua
```diff
@@ -0,0 +1,371 @@
...
+
+
+-- callback function for shortcuts
+local function program_shortcut(event, shortcut)
+	local choice = tonumber(string.sub(shortcut, -1))
+	OpenWith(dt.gui.action_images, choice)
+	end
+
+
+-- export images and reimport in collection
```
wpferguson: Is this to create an export from a raw so that the editor can be invoked on the export?

contrib/ext_editor.lua
```diff
@@ -0,0 +1,371 @@
...
+	nil -- view_leave
+	)
+
+
+-- register new storage
+dt.register_storage("exp2coll", "collection", show_status, export2collection)
+
+
+-- register the new preferences
+dt.preferences.register("ext_editor","program_path_9", "string","command for external editor 9", "command for external editor, including full path if needed","")
```
wpferguson: All of the path entries should be file choosers to prevent typing errors

PR #207, opened 2019-12-30
contrib/ext_editor version 2

Mark-64:
Hi, I have done a lot of changes to ext_editor script, adding funcionality and fixing bugs:
- the combobox with the list of editor has no longer always 9 entries, but a variable number, according to how many editors are configured
- you can add or remove editors in the preferences and update the module without restarting DT. There is a new button in the module for this. It is not an ideal solution but unfortunately a change in the preferences triggers no event in lua (maybe an idea for new versions of the API ?).
Cattura12
- translation, I finally managed to make it work. I propose a sample translation in italian
- more robust algorithm which takes care of the grouping / tagging / colors properties of the edited images
- better error handling and related error messages
- do not allow external editing of raw files, which can't be changed anyway
- fixing for some libs issues who made the script fail when filepaths contain spaces (I also filed a bug for this)
- lots of code refactoring and cleaning


Mark-64:
added further refinements and fixes in the initialization procedure

contrib/ext_editor.lua
```diff
@@ -83,26 +83,29 @@ local PS = dt.configuration.running_os == "windows" and  "\\"  or  "/"
...
+	dt.preferences.write(MODULE_NAME,"lastchoice", "integer", 0)
 	dt.preferences.write(MODULE_NAME,"initialized", "bool", true)
 	end 
-local lastchoice = 1
+local lastchoice = 0
 
 
 -- update lists of program names and paths, as well as combobox ---------------
-local function UpdateProgramList(combobox, button_pressed) 
+local function UpdateProgramList(combobox, button1, button2, update_button_pressed)
```
supertobi: button1 and button2 are not the best name, perhaps rename them to button_edit and button_edit_copy

supertobi:
And please do not commit the .mo file.

Mark-64:
I did the requested changes, thanks for the suggestion

supertobi:
The changes look good, but:

"This branch has conflicts that must be resolved"
@Mark-64 could you look into this?

Mark-64:
conflict resolved

Issue #210, opened 2020-01-03
lua script face_recognition - small issue


Yvan-56:
Hi, I have just tested the contrib face_recognition lua script.
I encountered a small issue related to locate: when passing the tolerance argument, my locate converts "." to "," going to command error.
This fix has helped me:
      dt.print_log ("Face recognition: Path to unknown images: " .. path)
 --  BEG - prevent converting "." to "," depending on locales, forced to C locale
      os.setlocale("C")
      local tolerance = dt.preferences.read(MODULE, "tolerance", "float")
      local command = bin_path ..  " --cpus " .. nrCores .. " --tolerance " .. tolerance .. " " .. knownPath .. " " .. path .. " > " .. OUTPUT
      dt.print_log("Face recognition: Running command: " .. command)
      os.setlocale()
 --  END - prevent converting "." to "," depending on locales, back to system locales

Hope this will help others.
Regards,
Yvan

supertobi:
It would be cool if you could make a Pull request for the fix.

Yvan-56:
I am not familiar with Git. I hope the way I have proposed the pull request is correct.
And hope this will help at the end.
Regards,
Yvan

PR #211, opened 2020-01-04
Update face_recognition for tolerance and locales

Yvan-56:
When locales use comma for numeric decimal separator, tolerance argument is not passed correctly.
I propose to switch temporally to "C" locales before converting float tolerance to string to get a period decimal separator.
Related to issue "lua script face_recognition - small issue #210".
Many thanks,
Yvan


supertobi:
Thanks

Issue #216, opened 2020-01-19
'dtutils_string.sanitize' does not handle quote characters in input.


schwerdf:
The [`dtutils_string.sanitize` function](https://github.com/darktable-org/lua-scripts/blob/32d5240e5debec6cd9ccd0d361d9b061bd827e8f/lib/dtutils/string.lua#L195) does not produce a correct escaping of its input string in all cases. Specifically, on the POSIX side, trying to "sanitize" a filename with an apostrophe in it (such as I sometimes encounter when exporting by title) will result in a syntactically invalid command string. The Windows side seems similarly afflicted.

wpferguson:
Thanks for reporting it.  I'm already working on the sanitize function, so I'll add this to my list of tests.

wpferguson:
@schwerdf I rewrote sanitize and is_not_sanitized to make them more robust.  I tested it with a variety of strings, sanitized and unsanitized with single and multiple embedded quotes that were sanitized and not sanitized.  If you find a string that causes an error, please send it to me and reopen this issue.  Thanks

schwerdf:
Since I do not have commit access to this repository, I cannot reopen the issue, but see pull request #218.

PR #218, opened 2020-01-21
Correct argument escaping in 'dtutils_string.sanitize'.

schwerdf:
This pull request modifies the `dtutils_string.sanitize` and `dtutils_string.is_not_sanitized` functions to handle quote characters, per issue #216.

Currently, quote characters in these functions are escaped with backslashes. However, neither the Bourne shell (the destination for these strings on POSIX systems) nor batch files (the destination on Windows) support that kind of escaping. Consequently, passing a sanitized argument with a quote character does not work right:
```
> dtutils_system.external_command("echo " .. dtutils_string.sanitize("ab'cd"))
sh: -c: line 0: unexpected EOF while looking for matching `''
sh: -c: line 1: syntax error: unexpected end of file
```
This pull request should fully correct the handling on POSIX and at least prevent code-injection on Windows. Specifically:
* On POSIX, any single quote in an unsanitized string is replaced with `'\''`, which is necessary since _no_ escaping is recognized inside single-quoted strings.
* On Windows, similarly, any double quote in an unsanitized string is replaced with `"^""`. Windows batch file escaping rules being so byzantine, this is far from a perfect sanitization, but since double quotes are invalid in Windows pathnames, it should handle intended use cases barring a proper solution to this issue (_viz._, modifying the API to allow the use of `execvp()` rather than `system()` for running external programs).

### DIFF
--- a/lib/dtutils/string.lua
+++ b/lib/dtutils/string.lua
@@ -174,48 +174,6 @@ function dtutils_string.urlencode(str)
 end
 
 
-dtutils_string.libdoc.functions["sanitize"] = {
-  Name = [[sanitize]],
-  Synopsis = [[surround a string in quotes making it safe to pass as an argument]],
-  Usage = [[local ds = require "lib/dtutils.string"
-
-    local result = ds.sanitize(str)
-      str - string - the string that needs to be made safe]],
-  Description = [[sanitize converts a string into a version suitable for
-    use passing as an argument in a system command.]],
-  Return_Value = [[result - string - a websafe string]],
-  Limitations = [[]],
-  Example = [[]],
-  See_Also = [[]],
-  Reference = [[]],
-  License = [[]],
-  Copyright = [[]],
-}
-
-function dtutils_string.sanitize_posix(str)
-  if dtutils_string.is_not_sanitized(str) then
-      return "'" .. string.gsub(str, "'", "'\\''") .. "'"
-  else
-       return str
-  end
-end
-
-function dtutils_string.sanitize_windows(str)
-  if dtutils_string.is_not_sanitized(str) then
-      return "\"" .. string.gsub(str, "\"", "\"^\"\"") .. "\""
-  else
-      return str
-  end
-end
-
-function dtutils_string.sanitize(str)
-  if dt.configuration.running_os == "windows" then
-      return dtutils_string.sanitize_windows(str)
-  else
-      return dtutils_string.sanitize_posix(str)
-  end
-end
-
 dtutils_string.libdoc.functions["is_not_sanitized"] = {
   Name = [[is_not_sanitized]],
   Synopsis = [[Check if a string has been sanitized]],
@@ -233,7 +191,7 @@ dtutils_string.libdoc.functions["is_not_sanitized"] = {
   Copyright = [[]],
 }
 
-function dtutils_string.is_not_sanitized_posix(str)
+local function _is_not_sanitized_posix(str)
    -- A sanitized string must be quoted.
    if not string.match(str, "^'.*'$") then
        return true
@@ -253,7 +211,7 @@ function dtutils_string.is_not_sanitized_posix(str)
    end
 end
 
-function dtutils_string.is_not_sanitized_windows(str)
+local function _is_not_sanitized_windows(str)
    if not string.match(str, "^\".*\"$") then
       return true
    else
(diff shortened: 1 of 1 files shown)


wpferguson:
Ok, this makes sense.  I was use to the case "ab\"cd", which works, but I didn't think about single quotes being literal.  I want to run the string stuff through the test framework I have set up, and make sure something else doesn't break.  I also have a windows vm, so I'll run the test  framework on there.

And, I guess a final test should be the echo to make sure that the string I thought it should produce is actually a valid string :)

lib/dtutils/string.lua
```diff
@@ -192,33 +192,28 @@ dtutils_string.libdoc.functions["sanitize"] = {
   Copyright = [[]],
 }
 
-function dtutils_string.sanitize(str)
-  local result = str
-  local os_quote = dt.configuration.running_os == "windows" and '"' or "'"
-  local escaped_os_quote = "\\" .. os_quote
+function dtutils_string.sanitize_posix(str)
```
wpferguson: Why don't we declare this as 
`local function _sanitize_posix(str)` which will make this a private function for the library.

lib/dtutils/string.lua
```diff
@@ -192,33 +192,28 @@ dtutils_string.libdoc.functions["sanitize"] = {
...
-    if dtutils_string.is_not_sanitized(result) then --check for embedded os_quotes
-      pos = string.find(result, os_quote, 2)
-      while pos < string.len(result) do
-        if not string.find(result, escaped_os_quote, pos - 1) then
-          result = string.format("%s\\%s", string.sub(result, 1, pos -1), string.sub(result, pos))
-        end
-        pos = string.find(result, os_quote, pos+2)
-      end
-    end
+function dtutils_string.sanitize_windows(str)
```
wpferguson: `local function _sanitize_windows(str)`

lib/dtutils/string.lua
```diff
@@ -192,33 +192,28 @@ dtutils_string.libdoc.functions["sanitize"] = {
...
+  if dtutils_string.is_not_sanitized(str) then
+      return "\"" .. string.gsub(str, "\"", "\"^\"\"") .. "\""
+  else
+      return str
+  end
+end
+
+function dtutils_string.sanitize(str)
+  if dt.configuration.running_os == "windows" then
+      return dtutils_string.sanitize_windows(str)
```
wpferguson: `return _sanitize_windows(str)`

lib/dtutils/string.lua
```diff
@@ -192,33 +192,28 @@ dtutils_string.libdoc.functions["sanitize"] = {
...
+  else
+      return dtutils_string.sanitize_posix(str)
```
wpferguson: `return _sanitize_posix(str)`

lib/dtutils/string.lua
```diff
@@ -238,31 +233,40 @@ dtutils_string.libdoc.functions["is_not_sanitized"] = {
   Copyright = [[]],
 }
 
-function dtutils_string.is_not_sanitized(str)
-  local os_quote = dt.configuration.running_os == "windows" and '"' or "'"
-  local escaped_os_quote = "\\" .. os_quote
-  local length = string.len(str)
-  local not_sanitized = false
+function dtutils_string.is_not_sanitized_posix(str)
```
wpferguson: `local function _is_not_sanitized_posix(str)`

lib/dtutils/string.lua
```diff
@@ -238,31 +233,40 @@ dtutils_string.libdoc.functions["is_not_sanitized"] = {
...
-      while pos ~= length  do
-        if not  string.find(str, escaped_os_quote, pos - 1) then
-          not_sanitized = true
-        end
-        pos = string.find(str, os_quote, pos + 1)
-      end
-    else
-      not_sanitized = true
-    end
+function dtutils_string.is_not_sanitized_windows(str)
```
wpferguson: `local function _is_not_sanitized_windows(str)`

lib/dtutils/string.lua
```diff
@@ -238,31 +233,40 @@ dtutils_string.libdoc.functions["is_not_sanitized"] = {
...
+   if not string.match(str, "^\".*\"$") then
+      return true
+   else
+      return false
+   end
+end
+
+function dtutils_string.is_not_sanitized(str)
+  if dt.configuration.running_os == "windows" then
+      return dtutils_string.is_not_sanitized_windows(str)
```
wpferguson: `return _is_not_sanitized_windows(str)`

lib/dtutils/string.lua
```diff
@@ -238,31 +233,40 @@ dtutils_string.libdoc.functions["is_not_sanitized"] = {
...
   else
-    not_sanitized = true
+      return dtutils_string.is_not_sanitized_posix(str)
```
wpferguson: `return _is_not_sanitized_posix(str)`

wpferguson:
@schwerdf I think we should make the _posix and _windows routines private.  Thanks for the fixes that work correctlly.

And, thanks for reminding me that sometimes I'm not as clever as I think I am  :-D.

schwerdf:
Done in commit 51eda4e.

wpferguson:
Looks good to me.

PR #221, opened 2020-01-24
Add "no persons found" entry and other enhancements

piterdias:
1. Add "no persons found" entry for the situations that no persons are detected in the photos
2. Check if the tags to applied are non empty and skip otherwise. It applies to empty entries in the GUI ("unknown person tag" and "no persons found"), when the user types something and then deletes it
3. Change the string "togs of images to ignore" to "tags of images to ignore"

### DIFF
--- a/contrib/face_recognition.lua
+++ b/contrib/face_recognition.lua
@@ -69,16 +69,9 @@ end
 -- preferences
 
 if not dt.preferences.read(MODULE, "initialized", "bool") then
-  dt.preferences.write(MODULE, "unknown_tag", "string", "unknown_person")
-  dt.preferences.write(MODULE, "ignore_tags", "string", "")
-  dt.preferences.write(MODULE, "tolerance", "float", 0.6)
-  dt.preferences.write(MODULE, "num_cores", "integer", 0)
-  dt.preferences.write(MODULE, "known_image_path", "directory", dt.configuration.config_dir .. "/face_recognition")
-  dt.preferences.write(MODULE, "export_format", "integer", 1)
-  dt.preferences.write(MODULE, "max_width", "integer", 1000)
-  dt.preferences.write(MODULE, "max_height", "integer", 1000)
+  reset_preferences()
+  save_preferences()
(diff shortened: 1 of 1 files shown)


wpferguson:
I'm not sure I understand the purpose of tagging the image with "no persons found" if no persons are found.  Is this just a way to get feedback that the image was processed?  

For instance if I have a collection with a 1000 images and 200 of them have people in them, but they are scattered throughout the collection, I don't want to go through the whole collection selecting which images to process, so I just select them all and let the face detection software sort them out.  This way I would end up with 800 images tagged with "no person found". I don't see a benefit in this.  If we do this, I suggest we make it optional.

Can you please tell me your use case so I can better understand?

piterdias:
I thought about it when running face_detection against photos taken in a recent trip where several ones were tagged with "no_persons_found". Some of such photos had known persons (Python script failed to find) and the "no_persons_found" tag allowed picking up photos that I had fix manually.

The face_recognition (Python) script already writes "no_persons_found" to the facerecognition.txt file.
Currently face_recognition (Lua) script just use "no_persons_found", from facerecognition.txt, as a valid tag and creates it in darktable. My change allows the user
1) Change it to a meaningful name. I use "People|Face Recognition|No Persons Found", for example
2) Eliminating the tag by typing anything else and then deleting from the entry. The original Lua script behavior is just failing due to empty string during tag creation.

wpferguson:
This works fine, and since you can delete the tags and unknown and or no people found don't have to be tagged if you don't want them.  I do have one suggestion, add a function to reset the module back to the default tags and settings, then add a reset_callback = <name of reset function>, to the fc.widget, and make resetable true.

piterdias:
Sure, great idea. Let me work on this and let you know.

piterdias:
Hi. I added the reset feature and also showed the default values of the entries in the GUI to let people know how the script behaves when nothing is changed.

wpferguson:
Looks good to me.  Thanks

PR #222, opened 2020-01-28
allow use of 'which' command for macos in dtutils/file

matt-maguire:
In the lib/dtutils/file.lua script, if the path to an external command is not given in the darktablerc parameters, then it uses "which" command to try to locate the binary on the system. However, this is only done if the OS is linux. However, macos also supports commands like "which" and "test -e" just the same as in linux, so the script was modified to allow use of "which" command for macos in addition to linux.

### DIFF
--- a/lib/dtutils/file.lua
+++ b/lib/dtutils/file.lua
@@ -46,8 +46,8 @@ dtutils_file.libdoc.functions["check_if_bin_exists"] = {
     check_if_bin_exists first checks to see if a preference for the binary has been
     registered and uses that if found.  The presence of the file is verified, then 
     quoted and returned.  If no preference is specified and the operating system is
-    linux then the which command is used to check for a binary in the path.  If found
-    that path is returned.  If no binary is found, false is returned.]],
+    linux or macos then the which command is used to check for a binary in the path.
+    If found that path is returned.  If no binary is found, false is returned.]],
   Return_Value = [[result - string - the sanitized path of the binary, false if not found]],
   Limitations = [[]],
   Example = [[]],
@@ -75,7 +75,7 @@ function dtutils_file.check_if_bin_exists(bin)
         result = dtutils_file.sanitize_filename(path)
       end
     end
-  elseif dt.configuration.running_os == "linux" then
+  elseif dt.configuration.running_os == "linux" or dt.configuration.running_os == "macos"  then
     local p = io.popen("which " .. bin)
     local output = p:read("*a")
     p:close()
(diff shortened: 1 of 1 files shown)


supertobi:
I have no mac and can't test it. Have you tested it with some scripts? If yes, and if it works I'm fine with the change.

matt-maguire:
I tested it together with the autostyle.lua script. It can now pick up exiftool from my path even if I don't define   the pathname explicitly in darktablerc.

wpferguson:
When you install an executable on linux it's put in one of two directories, both of which are in your path.  So if I install darktable, then open a terminal and run `which darktable` I get /usr/bin/darktable.  MacOS uses a package based install paradigm, similar to windows.  You install a package and the operating system creates a finder entry for you to use the program you installed.  So, if we install darktable on MacOS, then open a terminal and run `which darktable`, we get a not found.  If we try and run darktable from the command line with just darktable, we get not found.  If you install a program using homebrew or macports, it might be found and it might not.  I've had mixed results.  Sometimes it's in your path, sometimes not.

In the original version of lib/dtutils/file.lua we tried to treat MacOS like Linux, including using the which command in check_if_bin_exists().  It did not work reliably and we (I) were constantly troubleshooting problems because of "I have XXXXX installed but I get `XXXXX not found`".  The only reliable way we've found to run executables on MacOS (and windows) is by specifying the path and so we modified a lot of scripts to include the file selector.  If you run contrib/gimp.lua or contrib/hugin.lua (and others), you see a file selector to specify the location of the required executables.  This was the only way we found that worked every time.  

I'm assuming this is in response to the fractured conversation (I'm only seeing random posts from it in 3 or 4 threads) about exiftool not found.  Now that I understand what it's about I'll contribute my 2 cents to the conversation.  

I'm not sure I want to go down this road again, using which with MacOS.  Unfortunately MacOS is not a command line system the way Linux is.

matt-maguire:
So I am using macports to install utilities like exiftool, and I’m building darktable from source and installing in /usr/local, which means on my system the which command will work well. But, this may not be typical of other Mac users, who may be installing application bundles in /Applications.

Since the code does already contain the logic to pick up the required pathname from darktablerc on macOS, and if “which” command is not reliable on macOS in the case where programs are installed as application bundles, then I agree that it makes sense to reject this pull request.

wpferguson:
I think I remembered why it was a problem.  When you install darktable as a package, it uses a "system" version of the path, and not your path, thus which doesn't look in the right places.

matt-maguire:
Closing this PR, as it seems it’s not so useful after all.

PR #223, opened 2020-01-28
Remove include_all.lua

wpferguson:
include_all.lua was written at a time when only Linux (and MacOS somewhat) was supported.  Now we are running cross platform and not all scripts run on all platforms. Therefore include_all.lua is likely to cause problems when trying to run all the scripts.  It's also a trap for new users that want "it" all without knowing what "it" is


supertobi:
Looks right, just do it.

PR #230, opened 2020-03-07
Sanitize filepath before checking for its existance in Windows (fix #229)

### LINKED ISSUE #229: Function dtutils_file.executable_path_widget() fails in Windows
The library function `dtutils_file.executable_path_widget()` registers a null string in preferences whenever the filepath contains spaces.
In turn, this causes the function `dtutils_file.check_if_bin_exists()` to return an incorrect value.
For example, this misbehaviour is visible in gimp.lua, which in fact doesn't work in Windows.

Tested with DT 3.1 git master on Windows 10

Mark-64:
See #229


wpferguson:
I'll test

PR #231, opened 2020-03-10
Add RL output sharpening script

Mark-64:
Added RL_out_sharp.lua
Richardson-Lucy output sharpening using GMic
This script provides a new target storage "RL output sharpen". 
Images exported will be sharpened using GMic (RL deblur algorithm)

Cattura15

### DIFF
--- a/contrib/RL_out_sharp.lua
+++ b/contrib/RL_out_sharp.lua
@@ -1,0 +1,208 @@
+--[[
+
+  DESCRIPTION
+    RL_out_sharp.lua - Richardson-Lucy output sharpening using GMic
+
+    This script provides a new target storage "RL output sharpen". 
+    Images exported will be sharpened using GMic (RL deblur algorithm)
+    
+  USAGE
+    * require this script from main lua file
+    * in lua preferences, select the GMic cli executable
+    * from "export selected", choose "RL output sharpen"
+    * configure output folder 
+    * configure RL parameters with sliders
+    * configure temp files format and quality, jpg 8bpp (good quality) 
+      and tif 16bpp (best quality) are supported
+    * configure other export options (size, etc.)
+    * export, images will be first exported in the temp format, then sharpened
+    * sharpened images will be stored in jpg format in the output folder
+
+  EXAMPLE
+    set sigma = 0.7, iterations = 10, jpeg output quality = 95,
+    that will be sufficient for most uses
+
+  CAVEATS
+    MAC compatibility not tested
+  
+  BUGS, COMMENTS, SUGGESTIONS
+    send to Marco Carrarini, marco.carrarini@gmail.com
+
+  CHANGES
+    * 20200308 - initial version
+  
+]]
+
+local dt = require "darktable"
+local du = require "lib/dtutils"
+local df = require "lib/dtutils.file"
+local dtsys = require "lib/dtutils.system"
+
+-- module name
+local MODULE_NAME = "RL_out_sharp"
+
+-- check API version
+du.check_min_api_version("5.0.2", MODULE_NAME)  -- darktable 3.x
+
+-- OS compatibility
+local PS = dt.configuration.running_os == "windows" and  "\\"  or  "/"
+
+-- translation
+local gettext = dt.gettext
+gettext.bindtextdomain(MODULE_NAME, dt.configuration.config_dir..PS.."lua"..PS.."locale"..PS)
+local function _(msgid)
+  return gettext.dgettext(MODULE_NAME, msgid)
+  end
+
+-- initialize module preferences
+if not dt.preferences.read(MODULE_NAME, "initialized", "bool") then
+  dt.preferences.write(MODULE_NAME, "sigma", "string", "0.7")
+  dt.preferences.write(MODULE_NAME, "iterations", "string", "10")
+  dt.preferences.write(MODULE_NAME, "jpg_quality", "string", "95")
+  dt.preferences.write(MODULE_NAME, "initialized", "bool", true)
+  end 
+
+
+-- setup export ---------------------------------------------------------------
+local function setup_export(storage, img_format, image_table, high_quality, extra_data)
+  -- set 16bpp if format is tif
+  if img_format.extension == "tif" then
+    img_format.bpp = 16
+    end
+  end
+
+
+-- temp export formats: jpg and tif are supported -----------------------------
+local function supported(storage, img_format)
+  return (img_format.extension == "jpg") or (img_format.extension == "tif")
+  end
+
+
+-- export and sharpen images --------------------------------------------------
+local function export2RL(storage, image_table, extra_data) 
+
+  local temp_name, new_name, run_cmd, result
+  local input_file, output_file, options
+
+  -- read parameters
+  local gmic = df.sanitize_filename(dt.preferences.read(MODULE_NAME, "gmic_exe", "string"))
+  local output_folder = output_folder_selector.value
+  local sigma_str = string.gsub(string.format("%.2f", sigma_slider.value), ",", ".")
+  local iterations_str = string.format("%.0f", iterations_slider.value)
+  local jpg_quality_str = string.format("%.0f", jpg_quality_slider.value)
+
+  -- save preferences
+  dt.preferences.write(MODULE_NAME, "sigma", "string", sigma_str)
+  dt.preferences.write(MODULE_NAME, "iterations", "string", iterations_str)
+  dt.preferences.write(MODULE_NAME, "jpg_quality", "string", jpg_quality_str)
+
+  local gmic_operation = " -deblur_richardsonlucy "..sigma_str..","..iterations_str..",1"
+ 
+  local i = 0
+  for image, temp_name in pairs(image_table) do
+
+    i = i + 1
+    dt.print(_("sharpening image ")..i.." ...")
+    -- create unique filename
+    new_name = output_folder..PS..df.get_basename(temp_name)..".jpg"
+    while df.check_if_file_exists(new_name) do
+      new_name = df.filename_increment(new_name)
+      -- limit to 50 more exports of the original export
+      if string.match(df.get_basename(new_name), "_%d%d$") == "_50" then break end
+      end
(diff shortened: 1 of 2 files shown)


contrib/RL_out_sharp.lua
```diff
@@ -0,0 +1,208 @@
...
+local dt = require "darktable"
+local du = require "lib/dtutils"
+local df = require "lib/dtutils.file"
+local dtsys = require "lib/dtutils.system"
+
+-- module name
+local MODULE_NAME = "RL_out_sharp"
+
+-- check API version
+du.check_min_api_version("5.0.2", MODULE_NAME)  -- darktable 3.x
```
wpferguson: This should be "5.0.0".  API versions 5.0.1 and 5.0.2 introduced some functions to control panel visibility, darkroom image selected, and collection manipulations.  Since you're not using any of those things, you the minimum API version is 5.0.0 since you are using dt.configuration.running_os.  This way the script will be compatible with darktable 2.6.x, for those who haven't upgraded yet.

Mark-64: done

contrib/RL_out_sharp.lua
```diff
@@ -0,0 +1,208 @@
...
+
+  DESCRIPTION
+    RL_out_sharp.lua - Richardson-Lucy output sharpening using GMic
+
+    This script provides a new target storage "RL output sharpen". 
+    Images exported will be sharpened using GMic (RL deblur algorithm)
+    
+  USAGE
+    * require this script from main lua file
+    * in lua preferences, select the GMic cli executable
```
wpferguson: After you select the executable, you have to restart darktable so that the darktablerc file, with the new preference gets read, otherwise the preference isn't available.

My preference is to do a df.check_if_bin_exists() then add the df.executable_path_widget() if it doesn't so the user can add the executable and then proceed to use the script without having to restart darktable.  If the executable does exist, then don't show the widget.

Mark-64: yes, I considered that option.
Personally, I prefer not to clutter the GUI with a widget that you use only once and forget
Moreover, df.check_if_bin_exists() adds a separator and a label that takes some space in the panel.
Maybe I can say to restart DT in the instructions, is that OK ?

contrib/RL_out_sharp.lua
```diff
@@ -0,0 +1,208 @@
...
+
+
+-- export and sharpen images --------------------------------------------------
+local function export2RL(storage, image_table, extra_data) 
+
+  local temp_name, new_name, run_cmd, result
+  local input_file, output_file, options
+
+  -- read parameters
+  local gmic = df.sanitize_filename(dt.preferences.read(MODULE_NAME, "gmic_exe", "string"))
```
wpferguson: If the user didn't go to the preferences and set this value, or if the user didn't restart after setting the preference, then this will return an empty string, IIRC, which will cause a crash when you try and execute the command.  You should add a check to make sure there is something.

Mark-64: will do

contrib/RL_out_sharp.lua
```diff
@@ -0,0 +1,208 @@
...
+  local sigma_str = string.gsub(string.format("%.2f", sigma_slider.value), ",", ".")
+  local iterations_str = string.format("%.0f", iterations_slider.value)
+  local jpg_quality_str = string.format("%.0f", jpg_quality_slider.value)
+
+  -- save preferences
+  dt.preferences.write(MODULE_NAME, "sigma", "string", sigma_str)
+  dt.preferences.write(MODULE_NAME, "iterations", "string", iterations_str)
+  dt.preferences.write(MODULE_NAME, "jpg_quality", "string", jpg_quality_str)
+
+  local gmic_operation = " -deblur_richardsonlucy "..sigma_str..","..iterations_str..",1"
```
wpferguson: I used fx_unsharp_richardsonlucy and built my widgets to resemble the gui in Gmic_qt in GIMP.  This adds the choice of exponential blur (0) or guassian blur (1) and a cut parameter that seems to preserve luminance (If I understand correctly it restricts the output betweeen 0 and 255 for each iteration).  fx_unsharp_richardsonlucy calls deblur_richardsonlucy and maps the arguments.

contrib/RL_out_sharp.lua
```diff
@@ -0,0 +1,208 @@
...
+
+    i = i + 1
+    dt.print(_("sharpening image ")..i.." ...")
+    -- create unique filename
+    new_name = output_folder..PS..df.get_basename(temp_name)..".jpg"
+    while df.check_if_file_exists(new_name) do
+      new_name = df.filename_increment(new_name)
+      -- limit to 50 more exports of the original export
+      if string.match(df.get_basename(new_name), "_%d%d$") == "_50" then break end
+      end
```
wpferguson: There is a function df.create_unique_filename() that does this but it goes to 99.

contrib/RL_out_sharp.lua
Mark-64: To my surprise, you don't need to restart DT when configuring GMic from lua preferences. It works just fine.
But it's true that if the user does not configure GMic there is an error, so I added a test anyway

contrib/RL_out_sharp.lua
Mark-64: I can use that, no pb, but the GMic functions of Gimp plugin are undocumented and I read somewher that they might change without notice. I thought using -deblur_richardsonlucy was safer.
When coming to preserve luminance, you might have a point. In fact, I have to force cut to 0,255 at the end of the pipeline (see GMic command line). Can't say which is better.
Is -fx_unsharp_richardsonlucy compatible with tiff input files ? I will have to test

contrib/RL_out_sharp.lua
Mark-64: Yes, this comes from my other script and when I tried to use df.create_unique_filename() it didn't work for some reason, so I implemented myself.
I will try again

contrib/RL_out_sharp.lua
```diff
@@ -0,0 +1,208 @@
...
+  value = dt.preferences.read(MODULE_NAME, "output_folder", "string"),
+  is_directory = true,
+  changed_callback = function(self)
+    dt.preferences.write(MODULE_NAME, "output_folder", "string", self.value)
+    end
+  }
+
+sigma_slider = dt.new_widget("slider"){
+  label = _("sigma"),
+  tooltip = _("sigma parameter in RL algorithm"),
```
wpferguson: Controls the width of the blur that's applied

contrib/RL_out_sharp.lua
```diff
@@ -0,0 +1,208 @@
...
+-- register the new preferences -----------------------------------------------
+dt.preferences.register(MODULE_NAME, "gmic_exe", "file", 
+_("executable for GMic CLI"), 
+_("select executable for GMic command line version")  , _("(None)"))
+
+-- set sliders to the last used value -----------------------------------------
+sigma_slider.value = dt.preferences.read(MODULE_NAME, "sigma", "float")
+iterations_slider.value = dt.preferences.read(MODULE_NAME, "iterations", "float")
+jpg_quality_slider.value = dt.preferences.read(MODULE_NAME, "jpg_quality", "float")
+
```
wpferguson: These statements will run when you first load the script, but not after so they will never update.  They need to be moved into the export2RL function if you want to save the values that were used.

contrib/RL_out_sharp.lua
wpferguson: You would use the widget the first time and then it wouldn't appear because the value is already set.  

But, adding restart to the instructions is fine too.

contrib/RL_out_sharp.lua
wpferguson: I wrote a postsharpen script that doesn't restrict the file types used.  I didn't try RL against all the file types so I'm not sure if there is a problem with them or not.  I guess I need to check.

It's [here](https://github.com/wpferguson/extra-dt-lua-scripts/blob/master/postsharpen.lua)

fx_unsharp_richardsonlucy just calls deblur_richardsonlucy to do all the work.

After thinking about it I would say that using deblur_richardsonlucy is probably a better choice since it is part of the api and not something built on top of it.

contrib/RL_out_sharp.lua
wpferguson: If there is a problem with create_unique_filename, let me know and I'll track it down.

Mark-64:
With the last commit, I think I've implemented the requested modifications.
Some comments:
1) I know and truly appreciate your script for output sharpening, but I wanted something very simple to routinely apply to my edits. For comparison, the super professional Capture One applies output sharpening by default and there is an option to disable, that's all. Not a single slider.
2) I used the df.create_unique_filename() as you requested, but yes there is an issue. The count doeasn't actually stop at 99. I guess it will be quite rare to go over 99, but anyway I will make a PR for the fix, since I have one.
3) About your last comment, those statements are there only to load the last used values of the sliders in the previous DT session at statup. I tried to put those within the widget declarations, but for some weird reason it doesn't work reliably that way. I made that more clear in the code comments.

wpferguson:
I exported a file just as a jpeg.  Then I tried various settings with RL_output_sharpen and loaded the sharpened file and the original export as layers in GIMP.  Then I set the blend mode to difference to see what effect the sharpening had.  A sigma of .7 with 10 iterations had almost no effect (I couldn't find anything).  I bumped the sigma to 2.0 (using my version) and could finally see a difference.  You might want to try various settings and comparing the outputs to figure out a good default setting.

Mark-64:
Well, it depends on the image, the size and the usage (archive, web, print).
I tipically use output sharpening when exporting images for web, in that case I resize from 6000x4000 native camera resolution to 3000x2000 and I use sigma=0.7.  See the effect below in a 1:1: crop:

Unsharpened
unsharpened

Sharpened
sharpened

Difference
difference

Anyway, in the last commit I increased the range of the sigma slider and put 1.0 as default, which by the way is the default in the Gimp plugin.

wpferguson:
Looks good to me

PR #232, opened 2020-03-11
Fix for df.create_unique_filename() not stopping at 99

Mark-64:
Fix for this library function not stopping the count as it should.
The issue is easily reproduced, just set the limit to something more manageable, like 10 ;-)


Mark-64:
messed up with commits

wpferguson:
Tried to make this bug happen on linux and windows and wasn't able to.  I've included a file, test_unique.lua that simply creates a file and then 99 copies of it.  Run it against the unpatched library file and see if it works for you.

[test_unique.zip](https://github.com/darktable-org/lua-scripts/files/4321784/test_unique.zip)

PR #233, opened 2020-03-11
fix for create_unique_filename not stopping at 99

Mark-64:
Fix for this library function not stopping the count as it should.
The issue is easily reproduced, just set the limit to something more manageable, like 10 ;-)

### DIFF
--- a/lib/dtutils/file.lua
+++ b/lib/dtutils/file.lua
@@ -451,8 +451,8 @@ function dtutils_file.create_unique_filename(filepath)
   while dtutils_file.check_if_file_exists(filepath) do
     filepath = dtutils_file.filename_increment(filepath)
     -- limit to 99 more exports of the original export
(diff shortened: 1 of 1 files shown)


wpferguson:
I tried to reproduce this on linux and windows and wasn't able to.  I wrote a test script, called test_unique.lua, that exercises the create_unique_filename function.  I'll include it here.  Please run it against your unpatched library and let me know what happens.
[test_unique.zip](https://github.com/darktable-org/lua-scripts/files/4322262/test_unique.zip)

Mark-64:
You are right, there is no malfunction now.
It turns out that I messed up with my fork of lua-scripts, so I was using a tweaked version of file.lua.
If there was an issue, it has been fixed since then.
I apologize for the inconvenience.
Closing

wpferguson:
No problem

PR #234, opened 2020-03-14
New script: 'transfer_hierarchy'.

schwerdf:
The recent debates over file management on the mailing list indicate that some other Darktable users would have a use for this script that I wrote for rapid archiving of photos.

Darktable's built-in batch move and copy functionality only allows one destination directory per batch (analogous to a `mv` or `cp` command applied to a list of files). This script moves or copies selected images to a new location in one batch while preserving their relative positions in the directory hierarchy (analogous to a `mv` or `cp` command applied to the directory that is the images' lowest-common-ancestor).

### DIFF
--- a/contrib/transfer_hierarchy.lua
+++ b/contrib/transfer_hierarchy.lua
@@ -1,0 +1,347 @@
+--[[
+  TRANSFER HIERARCHY
+  Allows the moving or copying of images from one directory
+  tree to another, while preserving the existing hierarchy.
+
+  AUTHOR
+  August Schwerdfeger (august@schwerdfeger.name)
+
+  ADDITIONAL SOFTWARE NEEDED FOR THIS SCRIPT
+  None.
+
+  USAGE
+  darktable's native operations for moving and copying images in
+  batches allow only one directory to be specified as the destination
+  for each batch. Those wanting to move or copy images from a _hierarchy_
+  of directories within darktable while preserving the directory structure,
+  must take the laborious step of performing the operation one individual
+  directory at a time.
+
+  This module allows the intact moving and copying of whole directory trees.
+  It was designed for the specific use case of rapidly transferring images
+  from a customary source (e.g., a staging directory on the local disk)
+  to a customary destination (e.g., a directory on a NAS device).
+
+  Instructions for operation:
+
+  1. Select the set of images you want to copy.
+
+  2. Click the "calculate" button. This will calculate the
+     lowest directory in the hierarchy that contains every selected
+     file (i.e., the common prefix of all the images' pathnames), and
+     write its path into the "existing root" text box.
+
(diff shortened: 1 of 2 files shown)


contrib/transfer_hierarchy.lua
```diff
@@ -0,0 +1,347 @@
...
+    for _,img in ipairs(darktable.gui.action_images) do
+       if job.valid then
+    	  destFilm = destFilms[img.film]
+	  transferFunc(img, destFilm)
+	  job.percent = job.percent + pctIncrement
+       end
+    end
+    job.valid = false
+    local filterRules = darktable.gui.libs.collect.filter()
+    darktable.gui.libs.collect.filter(filterRules)
```
wpferguson: I did a copy from a collection to a new collection.  After the copy I expected to end up at the destination, but the existing root just got refreshed and set to the top of the collection.  Would it be better to end up at the destination after a move or copy?

contrib/transfer_hierarchy.lua
```diff
@@ -0,0 +1,347 @@
...
+-- Preferences: END
+
+
+
+
+
+
+darktable.register_lib(LIB_ID,
+		       "transfer hierarchy", true, true, {
+			  [darktable.gui.views.lighttable] = { "DT_UI_CONTAINER_PANEL_RIGHT_CENTER", 1000 }
```
wpferguson: Setting the position to 700 puts transfer_hierarchy just below select, and selected_images.  1000 puts it at the top, above select, which might come as a surprise to users.

contrib/transfer_hierarchy.lua
```diff
@@ -0,0 +1,347 @@
...
+   end
+end
+
+-- Helper functions: END
+
+
+-- Widgets and business logic: BEGIN
+
+local sourceTextBox = darktable.new_widget("entry") {
+   tooltip = "Lowest directory containing all selected images",
```
wpferguson: Make strings translatable

schwerdf: Addressed in commit 0501758.

contrib/transfer_hierarchy.lua
schwerdf: Done in commit 0501758.

contrib/transfer_hierarchy.lua
schwerdf: Keeping the view on the source was done on account of there being no single "destination" directory for the images, so the only possible collection change would be to a collect-by-film-roll on the destination root, which in my case is an already-well-populated archive. Finding the freshly moved or copied images among all the images there is like finding a needle in a haystack.

wpferguson: That makes sense

PR #238, opened 2020-04-13
OpenInExplorer.lua

preklov:
This code adds the command to open the image file name in Finder, the file manager of Apple's macOS. In addition it uses only one function instead of three for the OSes in question.

The code is tested under macOS Catalina (10.15.4) with darktable 3.0.1. It is not tested under Linux or Windows.

PR #239, opened 2020-04-14
Enabled macOS Finder in OpenInExplorer.lua

preklov:
I rewrote the script to add mac support and by the way shortened the code considerably. I added gettext, too.

I tested the code under macOS Catalina (10.15.4) with Darktable 3.0.1. I hope there will be someone to test it under Windows and Linux.


preklov:
I have closed #238 (I'm slow to get comfortable with GitHub). I'm done with my changes and I'm looking forward to your tests.

contrib/OpenInExplorer.lua
```diff
@@ -32,82 +33,66 @@ Select the photo(s) you wish to find in explorer and press "Go to Folder".
...
-      resp = dsys.external_command(run_cmd)
-    end
-  else
-    dt.print('please select fewer images (max 15)')
-  end
+--Detect OS and modify accordingly--	
+local proper_install = true
+if act_os ~= "macos" and act_os ~= "windows" and act_os ~= "linux" then
+  proper_install = false
+  dt.print_error(_('OpenInExplorer plug-in only supports Windows and Linux at this time'))
```
wpferguson, line 61: Since you added mac support, you should add it to this message :)

contrib/OpenInExplorer.lua
```diff
@@ -32,82 +33,66 @@ Select the photo(s) you wish to find in explorer and press "Go to Folder".
...
+  return
 end
 
-local function open_in_nautilus() --Open in Nautilus
+--The commands to open the OSes' file managers
+local fmng_cmd = {}
+fmng_cmd.linux = [[busctl --user call org.freedesktop.FileManager1 /org/freedesktop/FileManager1 org.freedesktop.FileManager1 ShowItems ass 1 ]]
```
wpferguson, line 67: nautilus expects uri's as arguments, such as file:///some/path/to/an/image.cr2.  If you have multiple images then one one call should be made, with the number of images following ass and then the uri's separated by a space, such as busctl --user call org.freedesktop.FileManager1 /org/freedesktop/FileManager1 org.freedesktop.FileManager1 ShowItems ass 3 'file:///some/image1.cr2' 'file:///some/image2.cr2' 'file:///some/image3.cr2' ""

contrib/OpenInExplorer.lua
```diff
@@ -32,82 +33,66 @@ Select the photo(s) you wish to find in explorer and press "Go to Folder".
...
   if #images == 0 then
-    dt.print('please select an image')
+    dt.print(_('Please select an image'))
   elseif #images <= 15 then
     for _,image in pairs(images) do 
-      curr_image = image.path..PS..image.filename
-      local run_cmd = [[busctl --user call org.freedesktop.FileManager1 /org/freedesktop/FileManager1 org.freedesktop.FileManager1 ShowItems ass 1 ]] .. df.sanitize_filename("file://"..curr_image) .. [[ ""]]
+      curr_image = df.sanitize_filename(image.path..PS..image.filename)
+      local run_cmd = fmcmd..curr_image
+      if os == 'linux' then run_cmd = run_cmd .. [[ ""]] end
```
wpferguson, line 81: Linux should use one command, not up to 15 separate commands.  If you try multiple commands, then only the last one works.

wpferguson:
Linux is broken.  nautilus expects uri's as arguments, not paths.  I added code comments to that effect.  I'll test windows next.

contrib/OpenInExplorer.lua
```diff
@@ -32,82 +33,66 @@ Select the photo(s) you wish to find in explorer and press "Go to Folder".
...
-  end  
-end
 
 -- GUI --
 if proper_install then
   dt.gui.libs.image.register_action(
-    "show in file explorer",
-    function() open_in_filemanager() end,
-    "Opens File Explorer at the selected image's location"
+    _("Show in file explorer"),
```
wpferguson, line 94: Show should be lower case to fit in with all the other buttons in the selected image module

wpferguson:
Windows works fine

contrib/OpenInExplorer.lua
preklov: I did so only in the German translation :-(

contrib/OpenInExplorer.lua
preklov: OK.

preklov:
I made the changes and rewrote the function.

Please have another try. I'll reanimate my Linux notebook to test it on openSUSE.

contrib/OpenInExplorer.lua
```diff
@@ -32,82 +33,70 @@ Select the photo(s) you wish to find in explorer and press "Go to Folder".
...
   if #images == 0 then
-    dt.print('please select an image')
+    dt.print(_('Please select an image'))
   elseif #images <= 15 then
-    for _,image in pairs(images) do 
+    for _,image in pairs(images) do
       curr_image = image.path..PS..image.filename
-      local run_cmd = [[busctl --user call org.freedesktop.FileManager1 /org/freedesktop/FileManager1 org.freedesktop.FileManager1 ShowItems ass 1 ]] .. df.sanitize_filename("file://"..curr_image) .. [[ ""]]
+      if os == 'linux' then
+        run_cmd = fmcmd .. df.sanitize_filename("file://"..curr_image) .. [[ ""]]
```
wpferguson, line 82: This works if you open a single file.  Repeated calls to busctl only honors the last one.  My solution would be

```
local fmng_cmd = {}
fmng_cmd.linux = [[busctl --user call org.freedesktop.FileManager1 /org/freedesktop/FileManager1 org.freedesktop.FileManager1 ShowItems ass %d %s""]]
fmng_cmd.macos = 'open -R '
fmng_cmd.windows = 'explorer.exe /select, '
local file_uris = ""

--The working function that opens the file manager with the image file name selected.
local function open_in_fmanager(os, fmcmd)
  local images = dt.gui.selection()
  local curr_image = ""
  local run_cmd = ""
  if #images == 0 then
    dt.print(_('Please select an image'))
  elseif #images <= 15 then
    for _,image in pairs(images) do
      curr_image = image.path..PS..image.filename
      if os == 'linux' then
        --run_cmd = fmcmd .. df.sanitize_filename("file://"..curr_image) .. [[ ""]]
        file_uris = file_uris .. df.sanitize_filename("file://" .. curr_image) .. " "
        dt.print_log("file_uris is " .. file_uris)
      else
        run_cmd = fmcmd .. df.sanitize_filename(curr_image)
        dt.print_log("OpenInExplorer run_cmd = "..run_cmd)
        resp = dsys.external_command(run_cmd)
      end
      -- dt.print_log("OpenInExplorer run_cmd = "..run_cmd)
      -- resp = dsys.external_command(run_cmd)
    end
    if os == 'linux' then
      run_cmd = string.format(fmng_cmd.linux, #images, file_uris)
      dt.print_log("OpenInExplorer run_cmd = "..run_cmd)
      resp = dsys.external_command(run_cmd)
    end
  else
    dt.print(_('Please select fewer images (max 15)'))
  end
end

```

contrib/OpenInExplorer.lua
```diff
@@ -32,82 +33,70 @@ Select the photo(s) you wish to find in explorer and press "Go to Folder".
...
 
-local function open_in_nautilus() --Open in Nautilus
+--The commands to open the OSes' file managers
+local fmng_cmd = {}
+fmng_cmd.linux = [[busctl --user call org.freedesktop.FileManager1 /org/freedesktop/FileManager1 org.freedesktop.FileManager1 ShowItems ass 1 ]]
+fmng_cmd.macos = 'open -R '
+fmng_cmd.windows = 'explorer.exe /select, '
+
+--The working function that opens the file manager with the image file name selected.
+local function open_in_fmanager(os, fmcmd)
```
wpferguson, line 72: I'm not sure that passing os as a variable is a good idea, since I think that would override the lua supplied os.xxx() functions.

contrib/OpenInExplorer.lua
preklov: If this is ok for Windows and for Linux Nautilus, it is ok for macOS anyway. I thought this same procedure could work for Explorer and for Finder, because this would simplify the code, but it didn't work satisfactorily for Finder. So I'll leave it as it is. In the linux run_cmd line I changed "fmng_cmd.linux" for "fmcmd".

From this I conclude that the original script didn't work the way it should have on Linux. Is this right?

contrib/OpenInExplorer.lua
preklov: Are lua function parameter variables not local by default? But I won't take any risk, I'll name this parameter op_sys.

Now I want to commit my ultimate script. How should I proceed? I'm somewhat helpless at the sight of forks and branches and PRs. Many thanks for your help.

contrib/OpenInExplorer.lua
wpferguson: Yes, the original was supposed to open a window for each selected image, but it didn't work correctly.  When I tested this version I went back to the original and tested it to compare.  When it didn't open a window with all the files selected or a window for each file selected I queried the author and he confirmed that it was intended to open a window for each. So, once you added the uri instead of the file path, you reproduced the buggy behavior of the original exactly :).  In the end, you added a feature and fixed a bug.

contrib/OpenInExplorer.lua
wpferguson: You should just be able to push the changes to your branch on github and they should update the pull request

preklov: Just a few questions:

I'd like to add a few lines to the USAGE section about the differences concerning the actual OS. On the Mac there is one Finder window for each directory the image files belong to. In these windows the last corresponding selected image is highlighted. Is it different with Explorer and Nautilus?

Is there a way to translate this USAGE section?

Why isn't the variable "resp" local that holds the dsys.external_command return value? Is there a global variabe (which I couldn't find)? I think I can dispense with it completely and change the lines in question to simple statements as I don't need the value. To add a line with dt.print_error in case the command fails, is not really necessary, for you can conclude this fact from the missing output.

preklov: I updated my old Windows notebook and found that separate Explorer windows are opened for each selected image. My even older Linux notebook will possibly be a greater challange :-/

wpferguson: Explorer and Nautilus open one window for each image and the image is highlighted. 

There is no way to translate the usage section as it currently stands, since it's the leading comments to the file before any functions are defined.  script_manager displays the leading comments when you hover over the enable/disable button, and it gets them by grabbing the second block of comments from the file.  The libraries have built in documentation that can be accessed to build the library docs, but that's not set up for translation either.  Perhaps we need to open an issue about how to implement online help so that we could do usage translation.

The resp variable should be local.  It just returns success or failure from the command.

Linux will work the same as windows, i.e. 1 window per image.  I run linux as my base os, then run windows and macos in Virtualbox VMs.  I keep just enough images in each of the VMs to do testing.

preklov: Whoopee, my Linux notebook still obeys like an old dog, but leisurely as it is of a similar age. When I used to work on a Linux desktop, I also had a VB Windows 7 VM. But I don‘t need Windows any more, and my iMac lets me execute nearly all of the programs I was used to on Linux. As a benefit I get excellent communication with our family iPads and iPhones. So I‘ll save disk space and refrain from installing a second os.

My tests on openSUSE with KDE GUI featuring Dolphin as the default file manager showed one Dolphin window with one tab for each respective directory, highlighting all of the selected files within. In my opinion this is the optimal presentation.

I made a request on macuser, if someone knows of a similar behavior for macOS Finder.

Have a good time, don‘t let COVID-19 catch you. Volker

preklov: OT: Apropos whoopee: A wonderful definition of photography by an American photographer whose name I unfortunately forgot: "Photography is a way to say whoopee again." I think of it each time I work on a photo in dt.

preklov:
I think this is the finished version. What's next?

supertobi:
I expect @wpferguson will do a last test and then hits the merge button.

wpferguson:
@supertobi you read my mind.  @preklov looks good to me, thanks.

Everyone stay safe

PR #241, opened 2020-04-15
Solve stability and performance issues on face_recognition.lua

piterdias:
face_recognition.lua script face some stability and performance issues on an i5-6600, 32 GB and GTX 1060 6GB:

    1. Taking to much time – almost a day for 3000 images
    2. Crashing due to lines malformation in facerecognition.txt – some times face_recognition Python script concatenates two output lines in just once, what is not expected by face_recognition.lua script
    3. Crashing the whole system in Linux – I keep receiving “bash: fork: resource temporarily unavailable” trying to process a too big (starting from some hundreds) image list. Some other applications starts to crash at this point. It seems related to the attach tag call because commenting this instruction (for debugging purposes) eliminates the crashes.
After debugging the Lua script and checking Lua and SQL darktable out logs, I decided to make the following changes:
    1. Remove an inner loop in the processing results step – helps preventing issue 1
    2. Remove a loop in the export step – helps preventing issue 1
    3. Remove duplicate tags (when there are more more than a reference face for each person) while facerecognition.txt is read – helps preventing issue 1 and 3
    4. Make a sanity check with processing images, skip any malformation data – helps preventing issue 2
    5. Reduces the sqlite access by caching tags that where already created for a previous image – helps preventing issue 3

Those changes allowed may thousands (more than 30K) of images to be processed in half a day instead of several days.

### DIFF
--- a/contrib/face_recognition.lua
+++ b/contrib/face_recognition.lua
@@ -63,7 +63,7 @@ du.check_min_api_version("5.0.0", "face_recognition")
 gettext.bindtextdomain("face_recognition", dt.configuration.config_dir.."/lua/locale/")
 
 local function _(msgid)
-    return gettext.dgettext("face_recognition", msgid)
+  return gettext.dgettext("face_recognition", msgid)
 end
 
 -- preferences
@@ -92,8 +92,10 @@ local function build_image_table(images)
   end
 
   for _,img in ipairs(images) do
-    image_table[img] = tmp_dir .. df.get_basename(img.filename) .. file_extension
-    cnt = cnt + 1
+    if img ~= nil then
+      image_table[tmp_dir .. df.get_basename(img.filename) .. file_extension] = img
+      cnt = cnt + 1
+    end
   end
 
   return image_table, cnt
@@ -103,17 +105,12 @@ local function stop_job(job)
   job.valid = false
 end
 
-local function do_export(img_tbl)
+local function do_export(img_tbl, images)
   local exporter = nil
   local upsize = false
-  local upscale = false
   local ff = fc.export_format.value
   local height = dt.preferences.read(MODULE, "max_height", "integer")
   local width = dt.preferences.read(MODULE, "max_width", "integer")
-  local images = 0
-  for k,v in pairs(img_tbl) do
-    images = images + 1
-  end
 
   -- get the export format parameters
   if string.match(ff, "JPEG") then
@@ -134,7 +131,7 @@ local function do_export(img_tbl)
   local exp_cnt = 0
   local percent_step = 1.0 / images
   job.percent = 0.0
-  for img,export in pairs(img_tbl) do
+  for export,img in pairs(img_tbl) do
     exp_cnt = exp_cnt + 1
     dt.print(string.format(_("Exporting image %i of %i images"), exp_cnt, images))
     exporter:write_image(img, export, upsize)
@@ -188,7 +185,7 @@ local function ignoreByTag (image, ignoreTags)
       end
     end
   end
-  
+
   return ignoreImage
 end
 
@@ -221,25 +218,25 @@ local function face_recognition ()
   if nrCores < 1 then
     nrCores = -1
   end
-  
+
   -- Split ignore tags (if any)
   ignoreTags = {}
   for tag in string.gmatch(ignoreTagString, '([^,]+)') do
     table.insert (ignoreTags, tag)
     dt.print_log ("Face recognition: Ignore tag: " .. tag)
   end
-  
+
   -- list of exported images
 
   local image_table, cnt = build_image_table(dt.gui.action_images)
 
   if cnt > 0 then
-    local success = do_export(image_table)
+    local success = do_export(image_table, cnt)
     if success then
       -- do the face recognition
       local img_list = {}
 
-      for img,v in pairs(image_table) do
+      for v,_ in pairs(image_table) do
         table.insert (img_list, v)
       end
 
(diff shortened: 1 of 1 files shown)


contrib/face_recognition.lua
```diff
@@ -92,8 +92,10 @@ local function build_image_table(images)
   end
 
   for _,img in ipairs(images) do
-    image_table[img] = tmp_dir .. df.get_basename(img.filename) .. file_extension
-    cnt = cnt + 1
+    if img ~= nil then
```
wpferguson, line 95: Has you ever found an image that didn't exist?  Since images is dt.gui.action.images, this would be a bug in darktable, passing a non-existent image as part of the arguments.

contrib/face_recognition.lua
```diff
@@ -376,7 +387,7 @@ fc.known_image_path = dt.new_widget("file_chooser_button"){
   is_directory = true,
   changed_callback = function(this)
     dt.preferences.write(MODULE, "known_image_path", "directory", this.value)
-  end
+end
```
wpferguson, line 390: please restore the indent since this closes the changed_callback function

contrib/face_recognition.lua
```diff
@@ -403,9 +414,9 @@ fc.height = dt.new_widget("entry"){
 
 fc.execute = dt.new_widget("button"){
   label = "detect faces",
-  clicked_callback = function(this) 
+  clicked_callback = function(this)
     face_recognition()
-  end
+end
```
wpferguson, line 419: same as above

contrib/face_recognition.lua
```diff
@@ -422,14 +433,14 @@ local widgets = {
...
-  dt.new_widget("label"){ label = _("width  ")},
-  fc.width,
+  table.insert(widgets, dt.new_widget("section_label"){ label = _("processing options")})
+  table.insert(widgets, fc.tolerance)
+  table.insert(widgets, fc.num_cores)
+  table.insert(widgets, fc.export_format)
+  table.insert(widgets, dt.new_widget("box"){
+    orientation = "horizontal",
+    dt.new_widget("label"){ label = _("width  ")},
+    fc.width,
```
wpferguson, line 443: Did you mean to indent the above lines?  They shouldn't be.

contrib/face_recognition.lua
```diff
@@ -439,18 +450,13 @@ table.insert(widgets, dt.new_widget("box"){
 table.insert(widgets, fc.execute)
 
 fc.widget = dt.new_widget("box"){
-  orientation = vertical,
-  reset_callback = function(this)
+	orientation = vertical,
+	reset_callback = function(this)
```
wpferguson, line 454: Did you mean to add 2 extra indents?  Please fix.

contrib/face_recognition.lua
```diff
@@ -439,18 +450,13 @@ table.insert(widgets, dt.new_widget("box"){
...
     reset_preferences()
-  end,
-  table.unpack(widgets),
+	end,
+	table.unpack(widgets),
```
wpferguson, line 457: same as above

wpferguson:
I'll test and see if I can do some benchmarking to see what kind of improvement this yields.  I would guess most of the improvement came from 5, caching tags.  DB access is a killer.  I also found that I could save some time in the python face recognition script by optimizing my face library and not making the reference images full resolution.

contrib/face_recognition.lua
piterdias: I ran face_recognition.lua countless times in the last two weeks in order to realize how it was failing. Random (and very rare) times it failed trying to access image_table[img]. I cant tell why because most of times I was just running the same data set all over again without any error in this line. I decided doing a sanity check because I couldn’t reproduce the error in a deterministic way to raise an issue in darktable project.

contrib/face_recognition.lua
piterdias: I used an online beautifier to make sure all my changes were correctly indented. It also changed some of your code below “-- build the interface”. I will rollback it.

contrib/face_recognition.lua
piterdias: I used an online beautifier to make sure all my changes were correctly indented. It also changed some of your code below “-- build the interface”. I will rollback it.

contrib/face_recognition.lua
piterdias: I used an online beautifier to make sure all my changes were correctly indented. It also changed some of your code below “-- build the interface”. I will rollback it.

contrib/face_recognition.lua
piterdias: I used an online beautifier to make sure all my changes were correctly indented. It also changed some of your code below “-- build the interface”. I will rollback it.

contrib/face_recognition.lua
piterdias: I used an online beautifier to make sure all my changes were correctly indented. It also changed some of your code below “-- build the interface”. I will rollback it.

contrib/face_recognition.lua
wpferguson: I saw where you did another check later and wondered if that was the case.

wpferguson:
Did you find an optimum size for image exports when you were doing your testing?  I've been using 1000x1000.  During my profiling, export took the most time, followed by the face_recognition script, then tagging.  The tag caching did result in a significant time savings.  And, the line checking in the face_recognition.txt file is good.  I ran into the blank and concatenated line problems during testing with the current contrib version.

piterdias:
I didn't touch the size of scaling because my library has a lot of photos where people are much smaller than the landscape behind them (trekking). I was afraid that the faces would become too small and reduce accuracy of detection. Also, the export is a mix of modules pipeline and final image export, so I am not sure if changing the export size really helps.
So you got blank and concatenated on “face_recognition.txt” but my changes were able to skip them?
I am glad that you saw significant time savings.

wpferguson:
Did a quick test with resolution - 462 images

1000x1000 - 271 seconds export, 82 seconds face recognition, 3 seconds tagging, 261 detected
500x500 - 268 seconds export, 48 seconds face_recognition, 3 seconds tagging, 251 detected
2000x2000 - 294 seconds export, 162 seconds face_recognition, 3 seconds tagging, 306 detected
3000x3000 got me 3 more faces just processing the untagged.

Or perhaps a 2 pass approach 500x500 then 2000x2000. Tried it but the export time outweighed the detection time.

PR #243, opened 2020-04-26
Fixes face_recognition.lua failing to run in a fresh installation of darktable and adds some more logs

piterdias:
1 - face_recognition.lua fails to run in a fresh installation of darktable. It is due to pull request #221 where  reset_preferences() and save_preferences() are called before being defined. This is actually a expected behavior because compiles script is one pass instead of two pass (like some other languages, like C). Moving a small piece of preferences setting code to the end of file solves the issue, because all required objects and functions are defined at that point.

2 - Add some more logs

### DIFF
--- a/contrib/face_recognition.lua
+++ b/contrib/face_recognition.lua
@@ -66,14 +66,6 @@ local function _(msgid)
   return gettext.dgettext("face_recognition", msgid)
 end
 
--- preferences
-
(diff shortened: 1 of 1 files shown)


wpferguson:
This has bitten me a couple of times in the past.  I've added a check to my testing of all scripts to try them on a clean system just to make sure they initialize properly the first time.

PR #247, opened 2020-05-04
Add an exporthaldclut lighttable module

noahwc:
This script creates a lighttable module which exports all user styles in the database as haldcluts. 

**Usage**
- Select the identity haldCLUT.
- Select a folder to output the generated haldCLUTs to
- Press export
- This will generate haldCLUTs for all user styles currently in the database.

Unfortunately due to limitations within the lua API I was unable to find a way to verify the a generated haldCLUTs is valid. Currently the script has only been tested on Linux (Ubuntu) but I would expect it to work on all OSs.

### DIFF
--- a/contrib/exportLUT.lua
+++ b/contrib/exportLUT.lua
@@ -1,6 +1,6 @@
 --[[
     This file is part of darktable,
-    copyright (c) 2016 Tobias Jakobs
+    copyright (c) 2020 Noah Clarke
 
     darktable is free software: you can redistribute it and/or modify
     it under the terms of the GNU General Public License as published by
@@ -16,18 +16,14 @@
     along with darktable.  If not, see <http://www.gnu.org/licenses/>.
 ]]
 --[[
+Add the following line to .config/darktable/luarc to enable this lightable module: 
+  require "exportLut"
 
-USAGE
-* require this script from your main lua file
-  To do this add this line to the file .config/darktable/luarc: 
-require "moduleExample"
+Given a haldCLUT identity file this script generates haldCLUTS from all the user's
+styles and exports them to a location of their choosing.
 
-* it creates a new example lighttable module
-
-More informations about building user interface elements:
-https://www.darktable.org/usermanual/ch09.html.php#lua_gui_example
-And about new_widget here:
-https://www.darktable.org/lua-api/index.html.php#darktable_new_widget
+Warning: during export if a naming collision occurs the older file is automatically
+overwritten silently.
 ]]
 
 local dt = require "darktable"
@@ -35,24 +31,20 @@ local du = require "lib/dtutils"
 
 du.check_min_api_version("3.0.0", "exportLUT") 
 
--- Thanks Kevin Ertel for this trick
+-- Thanks Kevin Ertel for this bit
 local os_path_seperator = '/'
 if dt.configuration.running_os == 'windows' then os_path_seperator = '\\' end
 
-local combobox = dt.new_widget("combobox"){label = "on conflict", value = 1, "skip", "overwrite"}
-
-local file_chooser_button = dt.new_widget("file_chooser_button")
-{
-    title = "Identity_file_chooser",  -- The title of the window when choosing a file
-    value = "",                       -- The currently selected file
-    is_directory = false              -- True if the file chooser button only allows directories to be selecte
+local file_chooser_button = dt.new_widget("file_chooser_button"){
+    title = "Identity_file_chooser",
+    value = "",
+    is_directory = false
 }
 
-local export_chooser_button = dt.new_widget("file_chooser_button")
-{
-    title = "Export_location_chooser",  -- The title of the window when choosing a file
-    value = "",                       -- The currently selected file
-    is_directory = true              -- True if the file chooser button only allows directories to be selecte
+local export_chooser_button = dt.new_widget("file_chooser_button"){
+    title = "Export_location_chooser",
+    value = "",
+    is_directory = true
 }
 
 local identity_label = dt.new_widget("label"){
@@ -63,7 +55,9 @@ local output_label = dt.new_widget("label"){
   label = "choose the output location"
 }
 
-local separator = dt.new_widget("separator"){
+local export_button = dt.new_widget("button"){
+  label = "export",
+  clicked_callback = export_luts
 }
 
 local function export_luts()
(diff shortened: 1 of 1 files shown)


contrib/exportLUT.lua
```diff
@@ -0,0 +1,126 @@
...
+
+    You should have received a copy of the GNU General Public License
+    along with darktable.  If not, see <http://www.gnu.org/licenses/>.
+]]
+--[[
+Add the following line to .config/darktable/luarc to enable this lightable module: 
+  require "exportLut"
+
+Given a haldCLUT identity file this script generates haldCLUTS from all the user's
+styles and exports them to a location of their choosing.
```
wpferguson, line 23: Could you provide a link to an explanation of a haldCLUT identity file?  I didn't know what one was but I found https://rawpedia.rawtherapee.com/Film_Simulation, which provided an explanation and a way to create one.

contrib/exportLUT.lua
```diff
@@ -0,0 +1,126 @@
...
+
+Warning: during export if a naming collision occurs the older file is automatically
+overwritten silently.
+]]
+
+local dt = require "darktable"
+local du = require "lib/dtutils"
+
+du.check_min_api_version("3.0.0", "exportLUT")
```
wpferguson, line 32: API version should be 5.0.0 since you are checking dt.configuration.running_os.

contrib/exportLUT.lua
```diff
@@ -0,0 +1,126 @@
...
+local output_label = dt.new_widget("label"){
+  label = "choose the output location"
+}
+
+local warning_label = dt.new_widget("label"){
+  label = "WARNING: files may be silently overwritten"
+}
+
+local function export_luts()
+  identity = dt.database.import(file_chooser_button.value)
```
wpferguson, line 63: Should you create a separate film roll for this file? if you just do an import, then the file is imported into the active film_roll at time of invocation.

contrib/exportLUT.lua
```diff
@@ -0,0 +1,126 @@
...
+    dt.print("Invalid identity lut file")
+  else
+    for style_num, style in ipairs(dt.styles) do
+      
+      identity:reset()
+      dt.styles.apply(style, identity)
+      
+      io_lut = dt.new_format("png")
+      dt.print("Exporting: " .. export_chooser_button.value .. os_path_seperator .. style.name .. ".png")
+      io_lut:write_image(identity, export_chooser_button.value .. os_path_seperator .. style.name .. ".png")
```
wpferguson, line 74: PNG files can be 8 or 16 bit.  Do you just want the default, should they always be 16 bit, or should the user decide?

contrib/exportLUT.lua
```diff
@@ -0,0 +1,126 @@
...
+    end
+    identity:reset()
+  end
+end
```
wpferguson, line 78: Feedback to the user would be nice so that they would know that the export completed successfully.

contrib/exportLUT.lua
```diff
@@ -0,0 +1,126 @@
...
+local warning_label = dt.new_widget("label"){
+  label = "WARNING: files may be silently overwritten"
+}
+
+local function export_luts()
+  identity = dt.database.import(file_chooser_button.value)
+  if(type(identity) ~= "userdata") then
+    dt.print("Invalid identity lut file")
+  else
+    for style_num, style in ipairs(dt.styles) do
```
wpferguson, line 67: Depending on the number of styles, this should probably be a job.  That would also give feedback to the user on the progress.

contrib/exportLUT.lua
```diff
@@ -0,0 +1,126 @@
...
+    identity:reset()
+  end
+end
+
+local export_button = dt.new_widget("button"){
+  label = "export",
+  clicked_callback = export_luts
+}
+
+if (dt.configuration.api_version_major >= 6) then
```
wpferguson, line 85: The current API version is 5.0.2.  I'm not aware of any major API changes on the horizon.  I think you don't need this piece of code.

wpferguson:
This is a neat idea.  I'm curious to see how it will work in practice, meaning that the applied LUT may not meet people's expectations depending on what the style contains.  I'll test it with some of the film styles and see what I get.

contrib/exportLUT.lua
noahwc: Here's a direct link to the identity haldCLUT: https://rawpedia.rawtherapee.com/images/8/83/Hald_CLUT_Identity.tif

Ideally I feel like it would be better from a usability standpoint to include it with the script. I wasn't able to find a clearly defined way to include extra resources with a script, however, so I left it up to the user to provide their own identity haldCLUT.

noahwc:
Thanks for the feedback, looks like I have some digging to do into the lua API :smile:. Hopefully in the next week or so I can push up some changes.

I think you're correct in stating that getting the plugin to be intuitive for users may be difficult. I used the script to create LUTs for the t3mujin film simulation pack and while the results were good I did run into some trouble with  the pixel pipe order ([here's a discussion on that](https://github.com/t3mujin/t3mujinpack/issues/23)). Beyond operations like cropping and adding frames will break the outputted LUT. Unfortunately I don't see a way around this (beyond documentation) since as far as I can tell the lua API lacks the ability to inspect the edits being performed by styles.

contrib/exportLUT.lua
noahwc: Added `io_lut.bpp = 16`. The LUTs should always be outputted as 16 bit images and I believe this accomplishes that.

contrib/exportLUT.lua
noahwc: Film rolls have always been a little mysterious to me so it's entirely possible I'm missing something here but I don't seem to encounter this behaviour.

contrib/exportLUT.lua
phweyland: > should they always be 16 bit, or should the user decide?

> The LUTs should always be outputted as 16 bit images

Are you sure ? Do you see any difference ? Let the user choose may be a good idea...

noahwc: After [doing some reading](https://rawpedia.rawtherapee.com/Film_Simulation#How_It_Works) it a appears at least in theory an 8-bit LUT should be sufficient. Initially I was apprehensive about relying on interpolation but it seems to work consistently after testing. While I'm generally in favour of giving users lots of options I don't see any particular advantage to adding a user configurable bit depth. Always open to discussion though.

contrib/exportLUT.lua
```diff
@@ -0,0 +1,121 @@
...
+     but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with darktable.  If not, see <http://www.gnu.org/licenses/>.
+]]
+--[[
+Add the following line to .config/darktable/luarc to enable this lightable module: 
+  require "exportLut"
```
wpferguson, line 20: should be 
`require "contrib/exportLut"`

wpferguson:
I found where you got the api_major_version > 6.  I guess I need to fix the moduleExample.lua file.  

You might also consider enabling translation.  You can use the gettextExample.lua file to see how it's done.

Tested it and it works fine.

wpferguson:
I had one more thought.  I tested using the film emulation styles, which I have set up in a style hierarchy.  When a style gets converted to a LUT, the name is Film|Color|Agfa|Agfaxxxxxxx.png.  Would it be better to put these styles in a folder hierarchy in the LUT folder similar to the style hierarchy?

noahwc:
Both those (the translation and folder structure) seem like good ideas. I'll do my best to push out an update in a relatively timely manner (uni is kicking back in so unfortunately I have some time constraints now 😞)

supertobi:
@noahwc What is the status of this PR?

noahwc:
@supertobi hopefully I'll have everything resolved within the next week. Definitely still committed to getting this done/merged.

noahwc:
@wpferguson I've enabled translation (I think) and luts are now exported to a folder hierarchy.

contrib/exportLUT.lua
```diff
@@ -0,0 +1,139 @@
...
+overwritten silently.
+]]
+
+local dt = require "darktable"
+local du = require "lib/dtutils"
+local df = require("lib/dtutils.file")
+local ds = require("lib/dtutils.system")
+
+local gettext = dt.gettext
+
```
wpferguson, line 35: Add this
  ```
gettext.bindtextdomain("exportLUT",dt.configuration.config_dir.."/lua/locale/")

  local function _(msgid)
      return gettext.dgettext("exportLUT", msgid)
  end
```
then you can replace your gettext.gettext with "_", which matches the convention used in the darktable source as well as the other scripts.

noahwc: Should be fixed now, thanks for the heads up!

PR #250, opened 2020-05-08
face_recognition.lua: Added "People" category to tag routine

sbolten:
Added field in GUI to provide "People" category

### DIFF
--- a/contrib/face_recognition.lua
+++ b/contrib/face_recognition.lua
@@ -154,7 +154,7 @@ local function reset_preferences()
   fc.unknown_tag.text = "unknown_person"
   fc.no_persons_found_tag.text = "no_persons_found"
   fc.ignore_tags.text = ""
-  fc.category_tags.text = "People"
+  fc.category_tags.text = ""
   fc.known_image_path.value = dt.configuration.config_dir .. "/face_recognition"
   fc.tolerance.value = 0.6
   fc.num_cores.value = -1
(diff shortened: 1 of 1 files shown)


contrib/face_recognition.lua
```diff
@@ -153,6 +154,7 @@ local function reset_preferences()
   fc.unknown_tag.text = "unknown_person"
   fc.no_persons_found_tag.text = "no_persons_found"
   fc.ignore_tags.text = ""
+  fc.category_tags.text = "People"
```
wpferguson, line 157: I think this should be an empty string so that it acts just like the current version when reset.  This takes into account the people who have been using this and have set up their reference face files with a tag hierarchy.

contrib/face_recognition.lua
```diff
@@ -303,6 +306,9 @@ local function face_recognition ()
                 t = nonpersonsfoundTag
               end
               if t ~= "" and t ~= nil then
+                if categoryTagString ~= "" then
+                  t = categoryTagString .. "|" .. t
```
wpferguson, line 310: This prepends the category string to all tags including no_persons_found and unknown_person.  While People|unknown_person makes a kind of sense, People:no_persons_found doesn't.  I suggest just adding the category tag before the recognized persons only.

sbolten:
I am actually not that familiar with github stuff - this is my first pull-from-fork-change-commit-submit-thing :-) Used GitKraken to do the changes.

How is this supposed to work now? I am doing the changes you suggested and then pushing it again?

Sorry for the maybe stupid question?

wpferguson:
I went to reply and saw that you had already figured it out.

sbolten:
Thx

Issue #245, opened 2020-04-28
Windows 10: problem with dt window after opening script manager


preklov:
Windows 10, dt 3.0.1, ASUS notebook 1920 x 1080

When I open the script manager in dt, the dt window becomes higher than the screen height, so the lower part of dt is hidden. In my experience it is **impossible** to scroll it up permanently to make it accessible. And I cannot resize the dt window’s height. Even if I switch from the long list of contrib entries to a shorter one, e.g. tools, the window doesn't shrink to its normal size.

Script manager closed: normal size:
![Screenshot (1)](https://user-images.githubusercontent.com/63585095/80487598-62c0fc80-895d-11ea-9684-9fe8a429527e.png)

Script manager opened with contrib list:
![Screenshot (2)](https://user-images.githubusercontent.com/63585095/80487718-8ab06000-895d-11ea-9b48-00a8a8248f73.png)

Script manager opened with tools list:
![Screenshot (3)](https://user-images.githubusercontent.com/63585095/80487788-a4ea3e00-895d-11ea-9b1f-ec67498a7565.png)

I don't know who the culprit is, Windows, dt or the lua script. Can anyone confirm this behavior?

wpferguson:
I have seen this behavior on Linux also.  I'm not sure who the culprit is.  I'll try playing with the script and see if there is something I can do with it.

preklov:
On a 1440 x 900 px notebook the dt window fits on the screen even with the full list of contrib scripts (OpenSUSE Leap 15.0, KDE 5.12.8). It seems Windows has a problem to adjust the font size of the list box.

wpferguson:
I experimented a little bit.  If I open all the modules in the left panel, then open script_manager, a scrollbar appears that scrolls the center and top portions of the panel.  If I set script_manager to use the center panel, then open all the modules then open script_manager, script_manager is included in the scrollable part of the panel.  Does this mean that the bottom part of the panel is not supposed to scroll, or is this a bug?  As an enhancement, we could do an api change and make the stack items scrollable.

In the meantime, a workaround is to set DT_UI_CONTAINER_PANEL_LEFT_BOTTOM to DT_UI_CONTAINER_PANEL_LEFT_CENTER which will prevent the issue from happening.

preklov:
The workaround works. Scrollable stack items would be most elegant.

wpferguson:
[...]

PR #251, opened 2020-05-23
Make script manager scrollable.

wpferguson:
Changed position of script manager to the center section of the left panel at the bottom of the section.  This takes care of the problem of darktable exceeding the vertical screen size and not being
scrollable.


wpferguson:
Fixes #245

PR #256, opened 2020-06-06
add photils plugin

scheckmedia:
A darktable plugin that tries to predict keywords based on the selected image. This plugin uses photils-cli to handle this task. Photils-cli is an application that passes the image through a neural network, classifies it, and extracts the suggested tags. Everything happens offline without the need that your data are sent over the internet.

[Further details at discuss.pixls.us](https://discuss.pixls.us/t/plugin-for-creating-automatic-tags/18167)

### DIFF
--- a/contrib/photils.lua
+++ b/contrib/photils.lua
@@ -16,13 +16,11 @@
 --]]
 
 --[[
-    A darktable plugin which tries to predict keywords based on the selected image.
-    This plugin uses photils-cli to handle this task. Photils-cli is an application
-    which passes the image through a neural network and generates a feature vector.
-    This feature vector will be used to find similar images in a self-hosted online
-    database and the corresponding tags. The usage of the feature vector has this advantage
-    that at no time your selected image will leave the PC. Rather a representation
-    of the image is sent.
+   A darktable plugin that tries to predict keywords based on the selected image.
+   This plugin uses photils-cli to handle this task. Photils-cli is an application
+   that passes the image through a neural network, classifies it, and extracts the
+   suggested tags. Everything happens offline without the need that your data are
+   sent over the internet.
 
     ADDITIONAL SOFTWARE NEEDED FOR THIS SCRIPT
     * photils-cli - https://github.com/scheckmedia/photils-cli
--- a/contrib/photils.lua
+++ b/contrib/photils.lua
@@ -160,8 +160,7 @@ function PHOTILS.image_changed()
 end
 
 function PHOTILS.tagged_image_has_changed()
-    GUI.warning.label = _(
-                            "The suggested tags were not generated\n for the currently selected image!")
+    GUI.warning.label = _("The suggested tags were not generated\n for the currently selected image!")
 end
 
 function PHOTILS.paginate()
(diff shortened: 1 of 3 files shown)


contrib/photils.lua
```diff
@@ -0,0 +1,414 @@
...
+            PHOTILS.tagged_image_has_changed()
+        end
+
+        PHOTILS.tagged_image = tostring(current_image)
+    end
+end
+
+function PHOTILS.tagged_image_has_changed()
+    GUI.warning.label = _(
+                            "The suggested tags were not generated\n for the currently selected image!")
```
supertobi, line 164: Why is here a line brake.

contrib/photils.lua
```diff
@@ -0,0 +1,414 @@
...
+    end
+
+    if PHOTILS.page > num_pages - 1 then
+        PHOTILS.page = num_pages
+        GUI.next_button.sensitive = false
+    else
+        GUI.next_button.sensitive = true
+    end
+
+    local offset = ((PHOTILS.page - 1) * PHOTILS.per_page) + 1
```
supertobi, line 187: A comment why you are doing this offset calculation would not hurt.

contrib/photils.lua
```diff
@@ -0,0 +1,414 @@
...
+    local image = dt.gui.selection()[1]
+    for tag, _ in pairs(PHOTILS.selected_tags) do
+        local dt_tag = dt.tags.create(tag)
+        dt.tags.attach(dt_tag, image)
+    end
+
+    dt.print(_("Tags successfully attached to image"))
+end
+
+function PHOTILS.get_tmp_file()
```
supertobi, line 216: @wpferguson I think is some thing we should add to lib/dtutils.file. What do you think?

contrib/photils.lua
```diff
@@ -0,0 +1,414 @@
...
+end
+
+function PHOTILS.get_tags(image, with_export)
+    local tmp_file = PHOTILS.get_tmp_file()
+    local in_arg = df.sanitize_filename(tostring(image))
+    local out_arg = df.sanitize_filename(tmp_file)
+    local executable = photils_installed
+
+    if dt.configuration.running_os == "macos" then
+        executable =  executable .. "/Contents/MacOS/photils-cli"
```
supertobi, line 239: I have no clue about macOS. Is it save to use a fixed path?

supertobi:
Works faster then expected, better results then expected and the photils-cli is smaller then expected. Very nice.

One small blocker, please write all labels in lower case to integrate it better in darktable. 

And I have three ideas of improvements. But nothing that blocks from merging.
1. Add the info, that photils-cli does not work on windows.
2. Perhaps an option to auto update the tags, when selecting a new image.
3. Perhaps show the confidence value in brackets behind the tags to get a feeling for this number.

scheckmedia:
I've changed to labels to lower case as suggested, looks better :)

1. I've added a small comment in the source code
2. I've implemented this function but the function slows down darktable a lot, so I will keep it out. A nice idea but unfortunately not possible at the moment.
3. it's implemented now

contrib/photils.lua
wpferguson: @supertobi I agree.  I think I would name it create_tmp_file, since it creates a tmp file.

contrib/photils.lua
wpferguson: @supertobi I think this is safe since @scheckmedia controls the executable and where it's installed.

supertobi:
Thanks for the nice script. I've merged it. Feel free to send new pull requests for further improvements.

contrib/photils.lua
supertobi: @scheckmedia Then please rename get_tmp_file() to create_tmp_file() and move it to lib/dtutils.file and then use it here, thanks.

PR #258, opened 2020-06-11
Update OpenInExplorer.lua

GLLM:
1. Add the possibility to create a shortcut for this LUA script
2. Capitalize the first letter of "Show in file explorer" which is the button created by the script

### DIFF
--- a/contrib/OpenInExplorer.lua
+++ b/contrib/OpenInExplorer.lua
@@ -104,8 +104,13 @@ end
 -- GUI --
 if proper_install then
   dt.gui.libs.image.register_action(
-    _("show in file explorer"),
+    _("Show in file explorer"),
     function() open_in_fmanager(act_os, fmng_cmd[act_os]) end,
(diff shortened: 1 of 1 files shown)


contrib/OpenInExplorer.lua
```diff
@@ -104,8 +104,13 @@ end
 -- GUI --
 if proper_install then
   dt.gui.libs.image.register_action(
-    _("show in file explorer"),
+    _("Show in file explorer"),
```
wpferguson, line 107: The darktable standard is that all actions start with lower case letters.  Look at the selected images module and notice all the actions are named that way.

GLLM: You are right. I've never managed to integrate this .... it's such a weird convention.
I'll correct this ...
Thanks @wpferguson

GLLM: OK, I'm lost .... I don't know how to change this.
I'll start from scratch :facepalm:

PR #262, opened 2020-06-15
Links option in OpenInExplorer.lua

preklov:
I added an option to contrib/OpenInExplorer.lua to use soft links in an arbitrary directory instead of opening one extra window for each selected image file in the OS's default file manager.

This option originates from the bug in the newer versions of macOS that the file manager Finder does not open a new window for each selected file, but only for each new directory, so the file name selections can only be recognized for the last file in the respective directory.

As Windows allows creating soft links in the terminal only to administrators, this option is not available to Windows users of dt.

In addition I provide a German translation for locate/de_DE/LC_MESSAGES/.


wpferguson:
In order to fix the problems (conflicts and file delete/replace history) I pulled this into a branch made the necessary changes.  I'm going to close this PR and replace it with #263.

PR #264, opened 2020-06-21
Fix OpenInExplorer shortcut function call

wpferguson:
Removed arguments from shortcut function call.  Removed DOS line endings.


wpferguson:
Removed DOS line endings which caused every line in the file to be marked as changed.  Will close this and submit another PR with the DOS line endings still in place.

PR #266, opened 2020-06-21
Reset GPS information

wpferguson:
Created reset_GPS.lua to reset the GPS information to what is in the image file or clear it if none is available.


supertobi:
@wpferguson What do you think about adding this to the geoToolbox.lua script?

wpferguson:
I was looking at geoToolbox.lua last night for another reason and as I was reading the code the thought crossed my mind that resetGPS would fit in there nicely.  I'll go ahead and do that.

PR #267, opened 2020-06-21
Image time

wpferguson:
Created image_time.lua to non-destructively adjust image time.

### DIFF
--- a/contrib/image_time.lua
+++ b/contrib/image_time.lua
@@ -109,7 +109,7 @@ local df = require "lib/dtutils.file"
 local ds = require "lib/dtutils.string"
 local gettext = dt.gettext
 
-local adj_time = {}
+local img_time = {}
 
 du.check_min_api_version("3.0.0", "image_time") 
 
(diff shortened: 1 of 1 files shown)


spaceChRiS:
Oh wow, this is incredible, big thank you again :smiley:! It works very well already, I did test a bit (darktable 3.0.0) and only found two little issues so far, but nothing mission critical:

1. In set time, you can also apply a date that misses some figures in the middle. This is e.g. convenient if you know the day and month, but not the exact year (e.g. ????-12-31). Darktable does the sorting correct in that case, at least to my taste. However, once you enter a year, you can only get the field blank again if you scroll up to the topmost entry above the year 1900. It would be great if the module reset button would reset all fields to blank.
2. You have to change the collection to apply the new date for the current lighttable sorting. It would be convenient if the sorting would be applied immediately and the lighttable would still focus on the selected image(s), but with a changed surrounding.

Thank you again, it seems, all my darktable dreams are becoming reality :grin:.

wpferguson:
The thought of a reset crossed my mind, but I wasn't sure it was needed.  Sounds like you found a good use case.

I'll think about how to do 2.  I think all of the pieces are there, so...

spaceChRiS:
Hm, after some more real life testing (I was able to finally set the dates of the scans of the ultrasonic images of my kids) I suggest to skip 2. or at least only make it optional. All the images without date tag are probably in a sequence at the beginning of the lighttable (sorted by date) and therefore it is convenient if the view does not change when adding time to several images in a sequence.

Issue #274, opened 2020-07-18
executable_manager doesn't appear to clear setting when clear button pushed


wpferguson:
If an executable path is set to a user's home directory in error, clearing it with executable manager appears to have no effect.  The file chooser button widget displays the directory that darktable was started in, usually the home directory if started from the gui, so even though the value was cleared no change was observed in executable manager.

wpferguson:
I propose adding an entry widget with the executable path displayed.  If no executable path is specified, then a placeholder of "search the path for the executable" will be displayed on linux and macos and a placeholder of "please select the executable with file chooser" will be displayed on windows.

When a user uses the clear button, then the path

PR #276, opened 2020-07-19
executable_manager provide feedback

wpferguson:
Added entry widget to display the current setting, or a placeholder specifying what would or should happen depending on operating system to provide visual feedback.

This change provides visual feedback when the value is set and when it is cleared.


wpferguson:
Fixes #274

Issue #281, opened 2020-08-11
Update README.md installation instructions


wpferguson:
Update the README.md to reflect the current best practices for installing and enabling the lua scripts.

PR #284, opened 2020-08-18
Update README.md

wpferguson:
Updated install instructions adding flatpak information and updating preferred way to enable and disable scripts.


wpferguson:
Fixes #281

Issue #213, opened 2020-01-05
executable_path gets set to darktable configuration directory


wpferguson:
The executable_path preference can get set to the darktable configuration directory, which should never happen.

wpferguson:
The work around is start tools/executable_manager and clear the setting for the executable.

wpferguson:
This is mitigated by the new check_if_bin_exists, and so ceases to be a problem.

PR #299, opened 2020-12-01
Fix check if bin exists and executable paths

### LINKED ISSUE #212: check_if_bin_exists returns a directory as the executable
If the executable path in the darktablerc file gets set to a directory path, check_if_bin_exists returns it as an executable.

wpferguson:
Rewrote check_if_bin_exists() so that it doesn't return directories as executables, or store them that way, thus preventing errors on *nix and macos systems.  Fixes #212.

Added a test_file function that can do cross platform tests to see if a path is a directory, or a file, executable, or exists.

check_if_bin_exists() can now do an operating system specific searches for executables that works well in well behaved systems and pretty good on windows :).  This means that we can get rid of the executable_path_preference widgets in the modules and rely on executable_manager if we need to override what the search auto detected.  The one problem executable I found was gimp on windows.  I have 2 different version of gimp on my windows test bed.  One works well, but the executable is gimp-2.10.exe not gimp.exe so it doesn't get found.  The other package is 2.10.10 and provides a gimp.exe which is detected, but doesn't start gimp and wait for it to finish.  If it do a wildcard search, i.e. gimp*exe, then lots of results are returned such as gimptool.exe, gimpdebug.exe, etc., which doesn't solve the problem either. So, on windows systems users will almost certainly have to specify the location of the gimp executable that should be run.

A partial copy of the old check_if_bin_exists() is added along with a preference to use it on windows systems.  The new check_if_bin_exists() does lots of system calls to check the file and to search which can result in "flickering windows" where a window is created to run the command then destroyed when it returns.  Since the result of check_if_bin_exists() gets saved when its found, this shouldn't be a problem, but it might be so I gave the windows users an out.

@supertobi sanity check please.  For the code, not for me :-D.


wpferguson:
I'll fix the conflicts...

wpferguson:
Used the web editor to fix the conflicts, but you have to merge once you've resolved the conflicts in order to save it.  So, when it comes time to merge I'll do the resolve.

wpferguson:
This also provides a permanent workaround for #213.  While the value can still get set to the directory that darktable starts in, check_if_bin_exists will fix it the first time the bin is checked.

PR #301, opened 2020-12-14
Sanitize filename in fujifilm_ratings.lua to handle spaces.

### LINKED ISSUE #300: Fujifilm ratings script doesn't support directories with spaces
In the Spanish Darktable Telegram group, one user reported problems that we reduced to the `contrib/fujifilm_ratings.lua` script. After some digging into the script, it is clear that is not designed with filenames/directories that support spaces. I would happily provide a PR, but I can't test the script.

pitbuster:
This should fix #300


wpferguson:
Looks good to me, thanks

PR #305, opened 2020-12-26
Update clear_GPS.lua

chrisaga:
Changed first line of header which was referring to another plugin (gimp.lua)


wpferguson:
That's what happens when you don't pay  attention when you copy and paste. :rofl:. Thanks

PR #306, opened 2020-12-26
Localize copy paste metadata.lua

chrisaga:
Use gettext in copy_paste_metadata.lua and add a po-file for French.


wpferguson:
Thanks

PR #309, opened 2020-12-28
script_manager.lua function clear_button must reset tooltip too

chrisaga:


wpferguson:
Actually, you don't need to clear the tooltip.  You need to clear the label because it's showing.  The tooltip is a popup that won't fire because the button is set to sensitive = false.  The next time the button gets used, the tooltip is populated from the script using the button.

You can test it by paging through the scripts until you reach the last page, then hover over an inactive button and you wont see a tooltip.

chrisaga:
Well … I did see tooltips on blank buttons (darktable 3.4.0, Kubuntu 20.04).
It's the reason why I looked at the code in the first place.
This might not be the right fix, but it does the job.
Do you think I should open an issue so somebody could investigate ?

wpferguson:
> I did see tooltips on blank buttons (darktable 3.4.0, Kubuntu 20.04).

When you submitted the PR this would have been good information to have and then I would have understood the reason rather than guessing.  

If you find a problem and don't have a solution, then open an issue.  If you find a problem and have a fix, do a PR but state the problem, and how to recreate it in the PR description.

And the problem was present on my system too, I just didn't hover long enough to make it happen.

Thanks

chrisaga:
OK ! It felt straightforward that I didn't want to bother you with long explanations. My mistake ;-)

PR #313, opened 2021-01-03
copy_attach_detachtags.lua localization proposal

chrisaga:


contrib/copy_attach_detach_tags.lua
```diff
@@ -163,7 +163,7 @@ end
 
 local function install_module()
   if not cadt.module_installed then
-    dt.register_lib("tagging_addon","Tagging addon",true,true,{
+    dt.register_lib("tagging_addon","tagging addon",true,true,{
```
wpferguson, line 166: You could also mark the change to be translated.  The second string is the label visible on the module.

chrisaga: Sure ! I wondered what the policy is regarding module (addon) names: 
1. **Should they be translated ?** After all darktable is still darktable in French. Not "table sombre" ;-) On the other hand, darktable's native modules do have their names translated.
2. Here is not a really good translation in French for addon. It could be "module additionnel" so **Tagging addon** could be translated as **module mots-clés**.
3. Finally, we should translate "geo toolbox" too. "geo boîte à outils" would look strange, but "gestionnaire de script" would be good for "script manager"

**Well the main question for the .lua file is point number 1. I'll fix this and do the same for the script manager since I translated it.**

wpferguson: I did a quick check and it doesn't seem that there is a standard.  I'm a native English speaker, so I don't think I should have a vote in the translate or not discussion.  I know that we try and translate the widget labels, tooltips and screen messages so it would seem logical to translate the module titles too, but I'm open to suggestion.  Perhaps a topic on pixls.us?

wpferguson:
Looks good to me, thanks

PR #314, opened 2021-01-06
Fujifilm dynamic range

dtorop:
This adds a script to compensate for Fujifilm's "dynamic range" setting. It reads the `RawExposureBias` tag on import and use this to appropriately offset the "EXIF exposure bias" for the image.

See https://github.com/darktable-org/darktable/issues/7705.

### DIFF
--- a/contrib/fujifilm_dynamic_range.lua
+++ b/contrib/fujifilm_dynamic_range.lua
@@ -2,7 +2,7 @@
 
 Compensate for Fujifilm raw files made using "dynamic range".
 
-Copyright (C) 2020 Dan Torop <dant@pnym.net>
+Copyright (C) 2021 Dan Torop <dant@pnym.net>
 
 This program is free software; you can redistribute it and/or modify
 it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


contrib/fujifilm_dynamic_range.lua
```diff
@@ -0,0 +1,115 @@
...
+
+gettext.bindtextdomain("fujifilm_dynamic_range", dt.configuration.config_dir.."/lua/locale/")
+
+local function _(msgid)
+	return gettext.dgettext("fujifilm_dynamic_range", msgid)
+end
+
+local function detect_dynamic_range(event, image)
+	if image.exif_maker ~= "FUJIFILM" then
+		dt.print_log(_("[fujifilm_dynamic_range] ignoring non-Fujifilm image"))
```
wpferguson, line 75: We came to a decision a while ago to not translate debug messages(i.e. print_log() and print_error()) because the author might not be able to read the output if it was a language they weren't familiar with. So the only thing we translate are labels, tooltips, and dt.print() messages.

contrib/fujifilm_dynamic_range.lua
```diff
@@ -0,0 +1,115 @@
...
+		dt.print_log(_("[fujifilm_dynamic_range] ignoring non-raw image"))
+		return
+	end
+	if not df.check_if_bin_exists("exiftool") then
+		dt.print_error(_("[fujifilm_dynamic_range] exiftool not found"))
+		return
+	end
+	local RAF_filename = df.sanitize_filename(tostring(image))
+	-- without -n flag, exiftool will round to the nearest tenth
+	local command = "exiftool -RawExposureBias -n -t " .. RAF_filename
```
wpferguson, line 89: df.check_if_bin_exists() returns the full path to the executable.  Using the returned value would let the script run on windows and macos too.

contrib/fujifilm_dynamic_range.lua
dtorop: Sounds good. Removed all translations, which also allowed for removing use of gettext.

contrib/fujifilm_dynamic_range.lua
dtorop: Done! (Though I can only test on Linux...)

dtorop:
@wpferguson: Thank you for reviewing. A new commit is uploaded for the suggested changes.

wpferguson:
Looks good to me, thanks

dtorop:
Thank you! The script makes a big difference here -- and way better to do this in lua than in darktable core. I hope it's useful to others.

PR #315, opened 2021-01-19
New utility function for random image access by ID.

schwerdf:
Going by recent correspondence on the mailing list, I am not the only Darktable user who has had to manipulate my database directly due to the Lua API's near-total lack of support for random access to image objects.

This pull request introduces a new utility function, `dtutils.find_image_by_id`, providing log-time random access by ID to any image in the database.

### DIFF
--- a/lib/dtutils.lua
+++ b/lib/dtutils.lua
@@ -283,4 +283,47 @@ function dtutils.check_os(operating_systems)
   return false
 end
 
+dtutils.libdoc.functions["find_image_by_id"] = {
+  Name = [[find_image_by_id]],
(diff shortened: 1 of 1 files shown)


wpferguson:
This looks good to me.

I naively assumed that image table row id and image id were the same so I checked my 5 year old database and found a 60000+ discrepancy between the two.

My plan is to merge this then submit an API change to add `darktable.database.get_image(imgid)`.  Once that gets merged, either you or I can modify this function to do an API check and use the direct access if it's available and the search if it isn't.  That way any one using dt 3.4 will have a way to directly access images and anyone using 3.5 or later will get the image quicker.

PR #327, opened 2021-02-21
External editors: make gui visible in darkroom

Mark-64:
With the recent introduction of export module in darkroom it comes in handy to make ext_editor.lua GUI visible in darkroom as well.
With this, we can export to collection and make round trips to our favorite program, all without leaving darkroom and not being forved to rely on keyboard shortcut only.

### DIFF
--- a/contrib/ext_editor.lua
+++ b/contrib/ext_editor.lua
@@ -337,23 +337,40 @@ local function export2collection(storage, image_table, extra_data)
 
 
 -- install the module in the UI -----------------------------------------------
-local function install_module()
+local function install_module(dr)
   if not ee.module_installed then
-    -- register new module "external editors" in lighttable and darkroom ------
-    dt.register_lib(
-      MODULE_NAME,          
-      _("external editors"),  
-      true, -- expandable
-      false,  -- resetable
-      {[dt.gui.views.lighttable] = {"DT_UI_CONTAINER_PANEL_RIGHT_CENTER", 100},
-       [dt.gui.views.darkroom] = {"DT_UI_CONTAINER_PANEL_LEFT_CENTER", 100}},  
-      dt.new_widget("box") {
-        orientation = "vertical",
-        table.unpack(ee.widgets),
-        },
-      nil,  -- view_enter
-      nil   -- view_leave
-      )
+    if dr then
+      -- register new module "external editors" in lighttable and darkroom ----
+      dt.register_lib(
+        MODULE_NAME,          
+        _("external editors"),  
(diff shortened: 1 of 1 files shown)


contrib/ext_editor.lua
```diff
@@ -409,27 +410,11 @@ local box1 = dt.new_widget("box") {
...
   }
 
+
+-- install module in lighttable and darkroom ----------------------------------
 table.insert(ee.widgets, combobox)
 table.insert(ee.widgets, box1)
-
--- register new module "external editors" in lighttable ------------------------
-if dt.gui.current_view().id == "lighttable" then
-  install_module()
```
wpferguson, line 417: Have you tried starting darktable and having it open an image directly, I.e.darktable /path/to/an/image.xxx ?  The reason this is in there is that adding a module to lighttable, when you're not in llighttable view, crashes darktable.

Mark-64: Indeed! 
I misunderstood the purpose of that part of code, sorry about that. 
Reverted now

wpferguson:
Should displaying the lib be an option, or make it the default but give the user a preference they can set to not display in darkroom?  Just a thought, it's up to you.

Mark-64:
good idea, will do

contrib/ext_editor.lua
```diff
@@ -334,30 +332,50 @@ local function export2collection(storage, image_table, extra_data)
...
-      {[dt.gui.views.lighttable] = {"DT_UI_CONTAINER_PANEL_RIGHT_CENTER", 100}},  
-      dt.new_widget("box") {
-        orientation = "vertical",
-        table.unpack(ee.widgets),
-        },
-      nil,  -- view_enter
-      nil   -- view_leave
-      )
+    if dr then
+      -- register new module "external editors" in lighttable and darkroom ----
```
wpferguson, line 343: you could do 

```
local views = {[dt.gui.views.lighttable] = {"DT_UI_CONTAINER_PANEL_RIGHT_CENTER", 100}}
if dr then 
views = {[dt.gui.views.lighttable] = {"DT_UI_CONTAINER_PANEL_RIGHT_CENTER", 100},
         [dt.gui.views.darkroom] = {"DT_UI_CONTAINER_PANEL_LEFT_CENTER", 100}}
end
```

then just replace the view part with the variable views and so only have the one code block for installing the module

Mark-64: implemented

wpferguson:
Thanks
\

PR #363, opened 2021-09-13
Fix URL of Finnish police document

lbschenkel:
The old URL returns a 404, this is the current replacement.


supertobi:
Thanks for the fix

PR #370, opened 2021-12-22
rename-tags: use darktable.tags.rename API

phweyland:
Solves partially the darktable issue [#10368](https://github.com/darktable-org/darktable/issues/10368) along with PR [#10669](https://github.com/darktable-org/darktable/pull/10669).


phweyland:
For the moment I don't see any benefit to use this script instead of the core tagging renaming.
So I prefer cancel this PR.
If the above assumption is true, this script should be set as deprecated.

PR #375, opened 2022-01-09
Added DNG to list of allowed file types for external editor

ChristianBirzer:
I tried to preprocess my images with Topaz DeNoise but was not allowed to open the DNG files with an external editor. 
I think, DNG is a common file type for external image processing programs so it should be allowed globally.
At least in this usecase, it works perfectly. DeNoise will write its output to the DNG and it is updated in dt.

### DIFF
--- a/contrib/ext_editor.lua
+++ b/contrib/ext_editor.lua
@@ -108,7 +108,7 @@ local n_entries
 
 
 -- allowed file extensions for external editors
-local allowed_file_types = {"JPG", "jpg", "JPEG", "jpeg", "TIF", "tif", "TIFF", "tiff", "EXR", "exr", "PNG", "png"}
+local allowed_file_types = {"JPG", "jpg", "JPEG", "jpeg", "TIF", "tif", "TIFF", "tiff", "EXR", "exr", "PNG", "png", "DNG", "dng"}
 
 
 -- last used editor initialization


wpferguson:
The allowed file types are all non-raw.  A DNG can be either raw or non-raw.  If we allow DNG's to be sent to external editors then there should be a check to make sure it's not an actual raw file.

ChristianBirzer:
Hmm. ok, I see the logic behind that. 
On the other hand, the DNG I sent to DeNoise was in fact the raw file from the camera. My approach is to first let DeNoise handle noise and sharpness in the raw and than use the generated DNG to do other things like exposure and crop and so on.
Otherwise I'd need a way to either export the image that was preprocessed in dt as DNG (would be cool...). Or export it as TIF, re-import it as a new file and process that one in DeNoise. In that case, the easy one-click copy and grouping is missing at least.
Then, the missing part is "export image according to some settings, reimport and group the exported and open that one in the external editor". Puh...
Aaah. stop. Now that I write this, I read the manual 🤣 There is the automatic reimport with the collection target storage. So to fit my needs, I have to export the image to that target and then edit it externally. Right?
That works fine. But: Is there a way to make this as a one-click action? Instead (or additionally) to edit a copy of the image in the external editor it would be great to "edit a developed copy" which will make the export (according to some setting of course).

So long story short: Forget about the DNG export. My wish than should be the single-click export-reimport-external-edit function 😉

wpferguson:
> My wish than should be the single-click export-reimport-external-edit function 

contrib/gimp.lua does a single export, edit, import, group.  You could use that as a template to write the function you desire.

ChristianBirzer:
Thanks a lot!
I'll take a look at the gimp.lua as soon as I have some more time and will try to add such a function 😀

PR #376, opened 2022-01-09
Update face_recognition.lua

Huipie1:
Add support for docker

python and docker-gpu have not been tested but should work if you have the right prerequisites


wpferguson:
If I understand correctly you only tested the docker-cpu branch and none of the others?

Huipie1:
Exactly, I only tested it with the cpu docker branch, as I don't have those available currently (also the reason why I added support for docker).

-edited the opening post to clarify a bit more

PR #377, opened 2022-01-09
Added tooltip to filter manager to better match new import scheme.

spaceChRiS:
The new import dialog does not immediately reflect the settings of the filter manager lua script, which made me wonder if it still works after upgrading from dt 3.2 to dt 3.8. It works, but does not fit very well to the new layout. A tooltip could prevent others to do the test themselves.

### DIFF
--- a/official/import_filter_manager.lua
+++ b/official/import_filter_manager.lua
@@ -45,6 +45,7 @@ local filter_dropdown = dt.new_widget("combobox")
 {
   label = "import filter",
   editable = false,
+  tooltip = "import filters are applied after completion of the import dialog",
 
   changed_callback = function(widget)
     dt.preferences.write("import_filter_manager", "active_filter", "string", widget.value)


official/import_filter_manager.lua
```diff
@@ -45,6 +45,7 @@ local filter_dropdown = dt.new_widget("combobox")
 {
   label = "import filter",
   editable = false,
+  tooltip = "import filters are applied after completion of the import dialog",
```
wpferguson, line 48: How about `import filters are applied during the import process`?

The import data is gathered (i.e. list of files), then the filter is applied, then the import is actually performed.  

The difference in the statements would be "we import all the data then delete what we don't want" versus "the filter discards the data we don't want to import, and imports the rest".  If you can think of a better way to word that, that would be great

spaceChRiS: Hm, the root issue is that the current implementation of input filters is not working well with the new import dialogs. With the old implementation, the folder import did not list any selected files. Therefore it was clear that the filtering happens on all files in the folder. Also, the filter settings appeared in the dialog, not in the module.

Now, the filter appears in the module, and there is only one unified dialog for file and folder import, which is great, but does not integrate well (UX wise) with the filtering, in particular because there are native filters in the dialog (ignore jpeg, only new images) which interfere with the concept of the lua script (a fix of the root cause would move the lua filtering here, but that is beyond my skills).

The tooltip should imho reflect that filtering is handled **after** the dialog and **on** the list of images that is returned by the dialog, and not “live” on the image selection in the dialog. But mentioning the two filter stages in the tooltip may be even more confusing. How about “``import filters are applied on the selected images after completion of the import dialog``”? “Completion” to me implies clicking OK and not cancel …

I am sorry that I bother you with this minor issue and steal your time. I am also fine with a rejection, as there may be not so many users of this feature (likely I am the only one) and therefore it's probably not worth the effort.

wpferguson: You're suggested tooltip works for me.  You're explanation made it clearer for me so now I understand the difference.

Issue #381, opened 2022-02-07
help wanted: strange / unknown behaviour of gui.create_job


ChristianBirzer:
I'm currently writing (or, trying to write 🙄) a lua script that will generate the preview cache for selected images (without leaving dt to start the external program...). Basically, it seems to work but I see two strange things that I don't understand right now.

First, the essential parts of my script:
```
local function stop_job( job )
  job.valid = false
end

local function create_previews()
  job = dt.gui.create_job( "generating previews", true, stop_job )
  local images = dt.gui.action_images
  for i,image in pairs(images) do
    if job.valid then
      image:generate_cache( true, 0, 8 )
      job.percent = i / #images
      dt.control.sleep( 1 ) -- without this, the job cannot be aborted
    else
      break
    end
  end
  if job.valid then -- without this, dt crashes with an exception in dt_pthread_mutex_destroy() after aborting the job
    job.valid = false
  end
end
```

the create_previews function will be called from a registered action.

My two questions:

1. Without the sleep() I cannot stop the running job. After starting the job, the progress bar is shown and when I hit the 'X' nothing happens until the for loop processed all images. Then the stop_job() is called. With the sleep the stop_job is called after processing the current image and the job terminates as desired.

2. When I set the job.valid to false after the loop it only works as long as the job is not terminated via the stop_job function. As this already sets the job to invalid I guess it is already destroyed and in the second job.valid=false dt crashes with an exception in dt_pthread_mutex_destroy.

Both phenomena are completely understandable, but: I see a lot of scripts that execute job.valid=false twice (when aborted) and seem to rely on stopping without a call to sleep(). Hmmm. What I'm missing here?

I hope you can clarify this,
thanks a lot!
Christian

PR #384, opened 2022-02-17
Add script to create thumbnails/previews of selected images

mfg92:
This script adds a button 'create thumbnails' to 'selected image[s]' module of darktable's lighttable view. When this button is pressed  full sized previews of all selected images are created.

Btw.: I noticed a discrepancy between the lua API description and the implementation of `dt_lua_image_t.generate_cache`.
In the [API documentation](https://docs.darktable.org/lua/stable/lua.api.manual/types/dt_lua_image_t/#dt_lua_image_tgenerate_cache) the last two parameters are called _start_size_ and _end_size_.
In the [code ](https://github.com/darktable-org/darktable/blob/master/src/lua/image.c#L102) they are called _min_ and _max_.
But not only their names are different but also the behaviour is not like described in the API documentation. When _max_/_start_size_ is greater than _min_/_end_size_ no image is added to the cache. But the API states _"When generating multiple sizes put the largest size as the start size and the smallest as the end"_.

### DIFF
--- a/contrib/create_thumbnails.lua
+++ b/contrib/create_thumbnails.lua
@@ -1,0 +1,86 @@
+--[[
+Create thumbnails plugin for darktable
+
+  darktable is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+  
+  darktable is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+  
+  You should have received a copy of the GNU General Public License
+  along with darktable.  If not, see <http://www.gnu.org/licenses/>.
+]]
+
+--[[About this Plugin
+This plugin adds the button 'create thumbnails' to 'selected iamge[s]' module of darktable's lighttable view
+
+
+----USAGE----
+Click the 'create thumbnails' button to let the script create full sized previews of all selected images.
+
+To create previews of all images of a collection:
+Use CTRL+A to select all images of current collection and then press the button.
+]]
+
+local dt = require "darktable"
+local du = require "lib/dtutils"
+
+du.check_min_api_version("8.0.0", "create_thumbnails_button")
+
+-- stop running thumbnail creation
+local function stop_job(job)
+    job.valid = false
+end
+
+-- add button to 'selected images' module
+dt.gui.libs.image.register_action(
+    "create_thumbnails_button",
+    "create thumbnails",
+    function(event, images)
+        dt.print_log("creating thumbnails for " .. #images .. " images...")
+
+        -- create a new progress_bar displayed in darktable.gui.libs.backgroundjobs
+        job = dt.gui.create_job("creating thumbnails...", true, stop_job)
+
+        for i, image in pairs(images) do
+            -- generate all thumbnails, a max value of 8 means that also a full size preview image is created
+            image:generate_cache(true, 0, 8)
+            
+            -- update progress_bar
+            job.percent = i / #images
+
+            -- sleep for a short moment to give stop_job callback function a chance to run
+            dt.control.sleep(10)
+
+            -- stop early if darktable is shutdown or the cancle button of the progress bar is pressed
+            if dt.control.ending or not job.valid then
+                dt.print_log("creating thumbnails canceled!")
+                break
+            end
+        end
+
+        dt.print_log("create thumbnails done!")
+
+        -- stop job and remove progress_bar from ui, but only if not alreay canceled
+        if(job.valid) then
+            job.valid = false
+        end
+    end,
+    "create full sized previews of all selected images"
+)
+
+-- clean up function that is called when this module is getting disabled
+local function destroy()
+    dt.gui.libs.image.destroy_action("create_thumbnails_button")
+end
+
+
+local script_data = {}
+script_data.destroy = destroy -- function to destory the script
+script_data.destroy_method = nil -- set to hide for libs since we can't destroy them commpletely yet
+script_data.restart = nil -- how to restart the (lib) script after it's been hidden - i.e. make it visible again
+return script_data
--- a/contrib/create_thumbnails.lua
+++ b/contrib/create_thumbnails.lua
@@ -16,11 +16,11 @@ Create thumbnails plugin for darktable
 ]]
 
 --[[About this Plugin
-This plugin adds the button 'create thumbnails' to 'selected iamge[s]' module of darktable's lighttable view
+This plugin adds the button 'create thumbnails' to 'selected image[s]' module of darktable's lighttable view.
 
 
 ----USAGE----
-Click the 'create thumbnails' button to let the script create full sized previews of all selected images.
+Click the 'create thumbnails' button in the 'selected image[s]' module to let the script create full sized previews of all selected images.
 
 To create previews of all images of a collection:
 Use CTRL+A to select all images of current collection and then press the button.


wpferguson:
You are correct about the documentation.  IIRC I wrote the documentation while the code was still a work in progress.  I finally set the code up to use min/max and handle it correctly internally.  But, it seems I forgot to go back and update the documentation.

wpferguson:
Documentation is updated to reflect the way the code works.

wpferguson:
After looking through the code and reading everything, I'm somewhat confused.  

Are you trying to generate a full set of cache images, a full resolution preview, or a full size preview, or something else that I didn't name?

Generating caches for sizes 0-8 is a waste of time and a waste of disk space.  You only need 3 sizes of cache images (or 2 if you don't use the filmstrip).  The sizes are:
* 0 for the filmstrip
* full size preview - 4 if your screen width is 1920 (HD), 5 if your screen width is 2560, etc...
* lighttable thumbnail size - the width of your screen / number of images in a row.  Match that to the size chart in the docs.  I have a 1920, and a 2560 and the thumbnail cache is size 2 (5 images per row).

Size 8 is the full resolution preview.  To produce it requires the pixelpipe and therefore takes a second or 2 on an unprocessed image.  I don't bother pre-generating them because the cost is not worth the time and storage.  If I'm doing a full preview culling and want to see the full resolution image I just generate it then, with a middle click on the image.

mfg92:
@wpferguson Ty for your response and correcting the API documentation.

The main purpose of this script for me is to generate the full size 100% previews. Why I need this? I shoot mostly wildlife and macro where you can never be sure if the focus is on the right spot or if there is motion blur in the image. That's why I shoot bursts, so I end up with many images with more or less the same content. Back at home I only want to keep one of the images of a burst. To be able to decide which image to keep I use the preview at 100% (middle mouse button). Generating this image (45MP) takes 3 sec on my PC. When I'm comming home from a shooting with 1000 images is is quite annoying to wait a thousand time 3 seconds.

Maybe I should not generate all thumbnail sizes < 8. But I can not hardcode the sizes which are needed as this script shall be general purpose and not tailored to my system. Perhaps it is possible to determine the width of the darktable window and the amount of rows, then the script could compute which thumbnail sizes are needed.

Another topic: Is it possible to get the information which modifier key was hold when the 'create thumbnails' button was pressed? This would allow me to implement another behaviour in those cases, like only generate the 100% previews when SHIFT is holed or delete all cached images if CTRL is holed.

ChristianBirzer:
Hi, and thanks a lot for doing my work 🤣.
This is exactly the use case I had in mind when starting with almost the same lua script (I mentioned it in issue #381).

Some ideas and suggestions:

- Instead of only the the `print_log` I added `print_toast` when generation was finished or aborted.
- I added an additional button to delete the thumbnails (cached images) with the `image:drop_cache()`. The main purpose for that was to make testing easier because I can delete and create the previews by a click an see how everything works, but it may be a good idea to have this button to explicitly clear the cache to free disc space. (you suggested deleting with a key modifier, but I'm sure, I won't remember it when I need it -> at least mention it in the buttons tool tip then for me 😊 )
- One thing I thought about but did not implement yet: Would it make sense to add a couple of checkboxes or a min/max dropdown in the preferences to select the cache sizes that shall be generated. In my case I really want the 100% and the lighttable size to allow fast scrolling the first time and instant 100% zoom to check focus and sharpness for a series of almost same pictures (doing wildlife and macro but with only 36MP 🙄). Other sized don't matter for me that much but I don't care about the disc space either.

mfg92:
@ChristianBirzer: I have implemented the first and scond point of your list of suggestions. I have to think a bit more about the third point.

mfg92:
The correct tooltip text of the script's buttons do not appear, instead it is always 'images', it looks like a bug to me:
grafik

The corresponding documentation can be found here: https://darktable-org.github.io/luadocs/lua.api.manual/darktable/gui/libs/image/#darktableguilibsimageregister_action

@wpferguson Can you have a look at this?

ChristianBirzer:
> I have implemented the first and scond point of your list of suggestions. I have to think a bit more about the third point.

Thanks a lot. A few days ago I played a bit with the preferences (just to learn...) and also thought about my third suggestion. 
I think, a couple of checkboxes is not that nice. In the preferences dialog, it won't look very nice when adding 8 checkboxes and the script would have to create each size with a call for its own.
It's easier to have a min/max setting in the prefs but I don't know if this helps that much (maybe just to omit the largest previews to save disc space...?)
If you want to add the min/max setting (with enum type/combo box), you might have a look at https://github.com/darktable-org/darktable/pull/11175

wpferguson:
@mfg92 https://github.com/darktable-org/darktable/pull/11216

AlicVB:
Sorry to chime in lately here, but maybe this points can help (or not... in this case just ignore my comment !) : 
- it's not mandatory to create thumbnails for all size , as small thumb may be created directly by reducing the size of larger one internally (and it needs very few resources if the sizes are not too different, esp. for small sizes) That can be taken in account to mitigate the disk cache size (and reduce the thumb creation time)
- Saying this, I really thing that a "min-max" widget is needed. Maybe we even can have a simple entry with a syntax like what exists in print dialog : (1;2;5 or 1-7 ...) it will need some extra code to decode the values, but it may be worth the pain...
- I wonder if a specific lib shouldn't be better in this case, as we may have 3 lines (min-max + create + delete) for something not used on daily basis... Using a specific lib allow to fold it, in order to gain some precious vertical space.

ChristianBirzer:
While doing my tests about preferences, I tried to add such a min-max widget to the lua options page. Using enums it could look like this:
grafik

I don't like (sorry for that wording 😉) such a entry field with a special syntax. From a developers view this is cool but from a useres view I don't know what to enter here. 
If we really need more than min/max maybe an additional lib would be the best because here we can use a couple of check boxes for the specific sizes.

BTW: Is it possible to add more complex widget layouts to the lua options in preferences? Such as tabbed pages or group boxes? This could make the lua options better readable. (I guess the answer is no because all pages are simple a list of labels and type dependent widgets 🤔)

wpferguson:
Here's a snapshot of how I'm doing it

pref

The options are 
* generate cache on import - automatically generate all the thumbnails I need as soon as import is complete.  The only drawback to this is that no other lua script can run until it finishes.
* other cache sizes to generate - my laptop is my main computer with an HD screen (1920 px).  My external monitor is 2560 px wide.  So full screen preview size is 4 for my laptop and 5 for my external monitor.  I list both sizes, separated by a comma, so that whichever size doesn't that get automatically generated, gets generated specifically.  
* generate full size preview for culling - Whether or not to generate the full size preview (not full resolution, so in my case 4,5) cache.
* generate culling cache first - I generate the lighttable thumbnails first by default.  If you want to start culling with full size previews immediately after export, then they can be generated first and then the lighttable thumbnails.
* screen width in pixels - you already know what this is.  I'm going to change it to an enum.

It addition to the above
* I've turned off filmstrip thumbnail generation in mine because I don't use the filmstrip.  If I do turn it on I don't notice any lag in the thumbnails appearing.  This is probably because my lighttable thumbnail size is 2, so it just grabs those and down sizes them.
* I also automatically regenerate the lighttable thumbnail after an edit, so that if you move from image to image during an editiing session you don't have to wait for all the thumbnails to regenerate when you return to lighttable view.

I see caching as a "set it and forget it" type of application, so I think putting the settings in the lua options is the best place for them rather than taking up space in the lighttable.

> BTW: Is it possible to add more complex widget layouts to the lua options in preferences? Such as tabbed pages or group boxes? This could make the lua options better readable. (I guess the answer is no because all pages are simple a list of labels and type dependent widgets thinking)

It could use some love :)

mfg92:
Sorry for the long delay, but I am unsure about how to proceed.

About @wpferguson implementation: I like the automatic generation of thumbnails on import. But I think additionally to that a button to trigger this manually is needed as well, for example when a user opens an already imported film roll that has no thumbnails yet (e.g. because the thumbnails were deleted in the past).

Furthermore I think we dont have to give the user the option to specify directly which thumbnail size they want, instead we could given them these three options and determin the corresponding thumbnail size (0-8) by using the display resolution:
- generate thumbnails ☑
   - size: 2? 3?
- generate fullscreen previews ☑
   - size: 4-7, depending on display resolution
- generate fullresolution previews ☑
   - size: 8

ChristianBirzer:
In my opinion (and my workflow) both is very useful (not to say 'needed'). The automatic generation on import and the manual generation.
My workflow in lightroom is like this: Import images from camera and automatically generate the full set of previews. During this time I do something else anyway so I don't care how long this takes. 
This allows me to sort out images very quickly including 100% view after I come back.
Very often I also work on past sessions. Due to the long time, the previews were already deleted, so before I start I'll regenerate them.

I think the preconfigured selection you suggest is very handy. Even if I don't know which sizes shall be in which selection. I'd like either have the checkboxes like you suggest where I can select all or have a drop down selection with an additional entry for 'all sizes'. I'm not sure which one is better.

wpferguson:
This kind of got overtaken by events.  The cache building crawler got implemented in darktable, so this kind of went by the wayside.

If you just want to regenerate thumbnails, there is official/regenerate_thumbmails

wpferguson:
I'm getting ready to do a merge fest with all the stuff I've been building, but not distributing.  responsive_cache wont be part of it, but I'll post a link.

PR #385, opened 2022-03-05
RL_out_sharp: copy EXIF, add variable output path (copy from rename_i…

hqhoang:
RL_out_sharp: support variable path (copy from rename_images.lua), copy EXIF to final file, work around GMIC problem with long filename by moving the exported file afterward.

Also update the end indentation and trim trailing spaces to match the other scripts's convention.

### DIFF
--- a/contrib/RL_out_sharp.lua
+++ b/contrib/RL_out_sharp.lua
@@ -19,19 +19,20 @@
   DESCRIPTION
     RL_out_sharp.lua - Richardson-Lucy output sharpening using GMic
 
-    This script provides a new target storage "RL output sharpen". 
+    This script provides a new target storage "RL output sharpen".
     Images exported will be sharpened using GMic (RL deblur algorithm)
 
   REQUIRED SOFTWARE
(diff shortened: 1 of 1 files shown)


wpferguson:
2 days ago I started looking at not waiting until the end to sharpen for postsharpen.lua.  Let me see what I come up with and we'll compare the 2.

The variable substitution code needs to be moved to a library.  This makes (at least) 3 scripts that are using it.  I'll try and get that done tonight for tomorrow.

hqhoang:
LOL, I was also about to request if you could move that variable substitution code into a library, many scripts definitely will benefit from that. Having the combo entry/button like the "file on disk" would be nice. No hurry, though. I got it to work for me at the moment. 

It seems that the old code was able to modify img_format.bpp (forcing TIF 16bit) but that doesn't seem to work anymore. When I set the return empty table {} from `initialize()` to stop the export, it would crash darktable if I click on the export buttons repeatedly. Not sure if I should submit a bug report.

PR #387, opened 2022-03-13
added script to integrate workflow for Helicon Focus stacking software

ChristianBirzer:
I've created a script to integrate Helicon Focus stacking software into darktable.

I'm not sure if it's ok to submit this in the dt contrib repo since it's about integrating a commercial application. On the other hand, there might be other users that use this software too.
So you decide. If you think it doesn't belong there you still can reject it 😉

Since it's my first 'complete' script there might also be some errors or missing things. 
I already planned to support translations but didn't implement it so far. German translation will come for sure tough.

grafik

BTW: Using the external editor script wasn't an option for me because Helicon Focus has very sophisticated command line parameters to pass input images and get results back what I wanted to make use of.

### DIFF
--- a/contrib/HeliconFocus.lua
+++ b/contrib/HeliconFocus.lua
@@ -50,7 +50,7 @@ local du = require "lib/dtutils"
 local df = require 'lib/dtutils.file'
 local dsys = require 'lib/dtutils.system'
 
-du.check_min_api_version("7.0.0", "HeliconFocus") 
+du.check_min_api_version("7.0.0", "HeliconFocus")
 
 local script_data = {}
 local temp
@@ -70,7 +70,7 @@ local GUI = { --GUI Elements Table
     add_tags_label       = {},
     add_tags             = {},
   },
-  options = {},  
+  options = {},
   run = {},
 }
 
@@ -103,7 +103,7 @@ local function create_temp_directory()
 end
 
 
---[[ export the given single image to tiff and write the file path of the tiff file to the 
+--[[ export the given single image to tiff and write the file path of the tiff file to the
       file that is given in inputfilelist
       param:  image - source image to export
               tempdirname - path to the temporary directory where to write the exported files to
@@ -115,14 +115,14 @@ local function export_image( image, tempdirname, inputfilelist )
 
   local exporter = dt.new_format("tiff")
   exporter.bpp = 16
-  
+
   local export_file = tempdirname..os_path_seperator..image.filename..".tif"
   exporter:write_image(image, export_file)
   dt.print_log( "exported file: "..export_file )
   inputfilelist:write( export_file..'\n' )
 end
 
---[[ this is the companion function for export_image that does not export the 
+--[[ this is the companion function for export_image that does not export the
     given image but just writes the path of the original to the list of files
     param:  image - source image to use
             tempdirname - path to the temporary directory, not used here
@@ -190,7 +190,7 @@ local function add_additional_tags( image )
     for tag in string.gmatch(set_tag, '[^,]+') do
       tag = clean_spaces(tag)
       tag = dt.tags.create(tag)
-      dt.tags.attach(tag, image) 
+      dt.tags.attach(tag, image)
     end
   end
 end
@@ -211,7 +211,7 @@ local function load_preferences()
   --   temp = nil
   -- end
   GUI.optionwidgets.add_tags.text = dt.preferences.read(mod, 'add_tags', 'string') -- temp
-  
+
 end
 
 -- stop running export
@@ -276,9 +276,9 @@ local function start_stacking()
     job.percent = i / #images
 
     -- sleep for a short moment to give stop_job callback function a chance to run
-    dt.control.sleep(10)    
+    dt.control.sleep(10)
   end
-  
+
   -- stop job and remove progress_bar from ui, but only if not alreay canceled
   if(job.valid) then
     job.valid = false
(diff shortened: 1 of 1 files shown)


mfg92:
Thank you @ChristianBirzer for this handy script, it makes working with Darktable together with Helicon Focus so much more fun.

I played around a bit with it and found a bug in the group feature: Only the first selected, the last selected and the stacked image are part of the newly created group but not the other selected images that were used to create the focus stack.

I have a feature request: In probably >95% of the times the location of the executable of Helicon Focus is "C:\Program Files\Helicon Software\Helicon Focus 8\HeliconFocus.exe" (were the 8 might change over time) on Windows. The script could check on startup/install/stackbuttonclick if HF is installed on that default location and if that's the case, set the property _Helicon Focus executable_ to that value (but only if the property is not already set).

And finally: When _Helicon Focus executable_ is not set but the users clicks "stack" a unspecific error message appears. It would be great if in that case the error message would tell the user that they have to go to the preferences and set a path for _Helicon Focus executable_.

Again, thank you very much for this great script.

contrib/HeliconFocus.lua
```diff
@@ -0,0 +1,491 @@
...
+
+  -- cleanup:
+  df.rmdir( tempdirname )
+
+end
+
+GUI.optionwidgets.use_original = dt.new_widget('check_button' ) {
+  label = _('use original RAW'),
+  value = false,
+  tooltip = _('stack original RAW images instead of exported'),
```
mfg92, line 353: perpahs something like "stack original RAW images instead of exported TIFF"

contrib/HeliconFocus.lua
```diff
@@ -0,0 +1,491 @@
...
+  end,
+  reset_callback = function(self)
+    self.value = false
+  end
+}
+
+GUI.optionwidgets.group = dt.new_widget('check_button') {
+  label = _('group'),
+  value = false,
+  tooltip = _('group stacked images'),
```
mfg92, line 365: This desciption is missing the information that the result  image is also part of the created group.
Suggestion: "group selected source images and imported results together"

contrib/HeliconFocus.lua
```diff
@@ -0,0 +1,491 @@
...
+  end,  
+  reset_callback = function(self) 
+    self.value = false 
+  end  
+}  
+
+GUI.optionwidgets.copy_tags = dt.new_widget('check_button') {
+  label = _('copy tags'),
+  value = false,
+  tooltip = _('copy tags'),
```
mfg92, line 377: Suggestion: "copy all tags from all source images to the imported result image(s)"

ChristianBirzer:
@mfg92 thanks a lot for your feedback and testing! I'll work through your suggestions the next days.

wpferguson:
> I have a feature request: In probably >95% of the times the location of the executable of Helicon Focus is "C:\Program Files\Helicon Software\Helicon Focus 8\HeliconFocus.exe" (were the 8 might change over time) on Windows. The script could check on startup/install/stackbuttonclick if HF is installed on that default location and if that's the case, set the property Helicon Focus executable to that value (but only if the property is not already set).

lib/dtutils.file has a function, `check_if_bin_exists`,  which will search the windows program directories for executable.  You just need to supply the name without the `.exe `extension.  If/when it's found the location will be saved to the darktablerc file so that the search doesn't have to be performed in the future.

If the file is not found, out can add a widget to your UI using the lib/dtutils.file function executable_path_widget which adds file selector widgets for each executable you specify.  The user selects the file, which is stored in the darktablerc, so the next check to check_if_bin_exists() will return the path.

wpferguson:
@ChristianBirzer 

> I'm not sure if it's ok to submit this in the dt contrib repo since it's about integrating a commercial application.

Scripts interfacing to commercial applications wont be merged.  However...

* you can maintain your own repository of scripts that won't/can't be accepted into the lua-scripts repository and we can add a link in the README.md on this repository.  In fact, I maintain just such a repository of scripts that didn't get merged, or wont be merged, or are a work in progress but useful  (https://github.com/wpferguson/extra-dt-lua-scripts)
* we can leave this open for awhile if you want to gather comments and feedback.

ChristianBirzer:
> Scripts interfacing to commercial applications wont be merged. However...

I expected this answer... But I love the idea of having a own (mini) repo with this script and get a link to it. Thanks!
I'll prepare this as soon as I added the suggestions above. 

Afterwards I'll close this PR.

ChristianBirzer:
Just found another bug, @mfg92 maybe you could verify this:
The call to Helicon Focus gets the preferred output path set to the path of the first input image. But this does not work when the path contains a german umlaut. 
On the console I see that path seems to have a wrong encoding.
grafik
However, the same image paths written to a text file (for the input images for Helicon) works fine.

The same problem occurs with other scripts like the "OpenInExplorer": Images in directories containing Umlauts are not working. I guess this is a more general problem (or only on my setup?)

mfg92:
For me folders with umlauts are working correctly in "OpenInExplorer" as well as in your script.

ChristianBirzer:
> If I start the offical darktable installation both do NOT work as expected.

Thanks a lot for testing. I guess this is worth a new issue or at least a note in the manual... (but I'm not sure if its lua or dt related).
I found out that it works fine when I activate the Unicode support in the Region settings (Control Panel -> Language -> Administrative language settings -> Administrative -> Change system locale
grafik
I was very surprised to see that UTF-8 support is marked as "Beta" in Windows 10 "Pro" from the year 2021 😱

ChristianBirzer:
Hi,
I started to host this script in my own repository. Not completely finished, but should be working so far.

https://github.com/ChristianBirzer/darktable_extra_scripts

wpferguson:
Added link to your scripts

PR #412, opened 2023-03-26
Add substitution to string lib

### LINKED ISSUE #386: Move variable substitution functions to a library
The variable substitution code used in rename_images.lua, is also used in postsharpen.lua and is now being added to RL_out_sharp.lua.  It is a sizeable piece of code (~130 lines) and should be in a library instead of being repeated in several places.

wpferguson:
Added the variable substitution functions used in other scripts to the string library.

Enhanced the substitution functions to be equivalent to the darktable variable substitution functions, including the string manipulations.  The darktable variable documentation, https://docs.darktable.org/usermanual/4.2/en/special-topics/variables/, can be used as the reference.

Fixes #386.

### DIFF
--- a/lib/dtutils/string.lua
+++ b/lib/dtutils/string.lua
@@ -706,7 +706,7 @@ function dtutils_string.build_substition_list(image, sequence, variable_string, 
                         image.version_name,                    -- VERSION.NAME
                         dt.configuration.version,              -- DARKTABLE.VERSION
                         "",                                    -- DARKTABLE.NAME
-                        sequence,                              -- SEQUENCE
+                        string.format("%04d", sequence),       -- SEQUENCE
                         image.width,                           -- WIDTH.SENSOR
                         image.height,                          -- HEIGHT.SENSOR
                         is_api_9_1 and image.p_width or "",    -- WIDTH.RAW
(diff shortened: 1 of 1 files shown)


wpferguson:
Needs https://github.com/darktable-org/darktable/pull/14032 and https://github.com/darktable-org/darktable/pull/13965 to supply data for the variables.

wpferguson:
For those wishing to test/play with substitution I've attached a test script and the image I used to test with is located here, https://drive.google.com/file/d/1xb7AcG_5NSxs4W679K3w61iPqigB5Fts/view?usp=share_link.  It's a play raw from pixls.

Put it in a folder and import it into darktable.  Get the image id (hover over it and look in the image information panel) and replace the 1 in the dt.database.get_image(1) statement.  The script prints to the console with the results.

[test_substitution.zip](https://github.com/darktable-org/lua-scripts/files/11073322/test_substitution.zip)

wpferguson:
The string substitution is feature complete.  Some of the variables aren't implemented since the main purpose of this is for use in generating path and filenames in the exporter.

However, you can use it to create formatted strings for other uses, such as GUI labels, messages, etc.  If enough use is made for this purpose, then the rest of the variables can be implemented.

The variables not implemented are:
```                     
DARKTABLE.NAME
WIDTH.MAX - Available from the exporter
HEIGHT.MAX - Available from the exporter
EXIF.DATE.REGIONAL 
EXIF.TIME.REGIONAL
IMAGE.EXIF
GPS.LOCATION
RATING.ICONS
LABELS.ICONS
TAGS
SIDECAR.TXT
OPENCL.ACTIVATED
JOBCODE
```
Some of these are comma separated lists, others are images, and some have no meaning such as jobcode which is used internally.

Issue #419, opened 2023-05-24
Update script_data blocks in scripts with script_data.show


wpferguson:
#418 is caused by the lack of a `show()` function in the returned `script_data` block.  When I copied a `script_data` block as an example I noticed that is didn't include a `.show` entry in the initialization, but it did return one.  This could lead to a misunderstanding, so the `script_data` initialization blocks should be updated to show all the elements.

wpferguson:
Fixed by #422.

PR #422, opened 2023-07-16
Fix script data show

wpferguson:
Added script_data.show to script_data blocks with an explanation for it's use.


wpferguson:
Fixes #419

PR #435, opened 2023-10-28
select non-existing images

ddittmar:
Enable selection of non-existing images in the the currently worked on images

This script helps (at least me) to select images that don't exist anymore. This way there is no need to call `purge_non_existing_images.sh`.


wpferguson:
Looks good to me.  Thanks

Issue #438, opened 2023-12-20
Some GUI proposal for the script manager


TurboGit:
The current script manager looks like:

![image](https://github.com/darktable-org/lua-scripts/assets/467069/d2fc6f30-3725-4071-b2da-c86abc265f02)

I find it hard to read each line and do proper separation between the script name and the status.

A proposal would be to left align the script name and right align the status and possibly use some symbol for started / stopped.

wpferguson:
Let's work on status first...

The buttons can be styled using CSS.  

Add this to your user.css to control the color of the labels (red for stopped, green for running):

```
button#sm_started label { color: rgb(50,210,50); }
button#sm_stopped label { color: rgb(210,50,50); }
```
![colorlabels](https://github.com/darktable-org/lua-scripts/assets/1634891/a5c21180-2c04-445e-86f3-ae3bfc0e1c84)

or if you'd rather have the button background turn colors:

```
button#sm_started label { background: rgb(0,128,0);
			 font-weight: bold; }

button#sm_stopped label { color: rgb(128,0,0);
			 font-weight: bold; }
```
![colorbkgd](https://github.com/darktable-org/lua-scripts/assets/1634891/56504ac8-bb2b-4b44-ada6-dd271b374de7)

You also need to drop down the `start/stop scripts` combobox and select configure, then click the `use color interface?` checkbox.
![colorinf](https://github.com/darktable-org/lua-scripts/assets/1634891/10cf410c-375c-4853-8022-950e711830fc)

The buttons in the lua API currently only support ellipsize as an argument.  I'd have to change the API and add `halign`.  I agree that they are hard to read with each line starting in a different place.

Probably the best solution to the just text configuration would be a table on each button with the name left aligned and the status right aligned.  I'll play with it and see if I can pass a widget to the button creation.

wpferguson:
Added left align

![leftalign](https://github.com/darktable-org/lua-scripts/assets/1634891/42eabdbe-0f7b-4609-ab8c-dd0f200a7ae8)


with color

![leftaligncolor](https://github.com/darktable-org/lua-scripts/assets/1634891/c5f478df-862f-4a51-bc35-98a7208f89a8)

TurboGit:
I'm not convinced with the color :)

What about just making the background as "selected" button on dt interface?

TurboGit:
I was initially thinking about something like this:

[...]

PR #450, opened 2024-01-17
POC: translation done by darktable.

TurboGit:
This is a POC on how Lua script translation could be handled directly in darktable.

### The darktable part

- Add lua-scripts has a submodule in src/external

- Add all lua files into `po/POTFILES.in`

This will make it possible with the current main darktable translation workflow to also translate the scripts.

 ### The Lua script part

- As proposed in this PR remove the gettext domain, use a common one to share as much as possible the strings across darktable and all lua scripts.

- Use common casing for all Lua strings. Indeed there is some string identical except for the casing. Again this is to share a maximum of string.

- Remove any spaces in start or end of strings, the layout must not be done in strings. I see for example `"of "` and some other cases.

### Redesign

- As discussed in issue #438 we also want to redesign the Lua script description to be more uniform and structured. Those description must be done using `_('...")` to allow for translation. Some description are quite big, I'm wondering if we want this in the dt tooltip or maybe a lite description in dt tooltip and a full description in main dt documentation. To be discussed.

- We probably want to have a string for the name of the module provided by the script. As I understand we use the script filename for the name today. This is not really user friendly, so better have a string for the name of the script which is displayed on the UI (as done for dt modules).

- The redesigned of the UI with standard button as already discussed.

@wpferguson : This is my current idea to make Lua scripts a first class citizen. Let me know how it sounds to you.

I hope I have not missed something, but at least with this POC I was able to translate Lua scripts in French.

### DIFF
--- a/official/copy_paste_metadata.lua
+++ b/official/copy_paste_metadata.lua
@@ -58,11 +58,8 @@ local publisher = ""
 local rights = ""
 local tags = {}
 
--- Tell gettext where to find the .mo file translating messages for a particular domain
-gettext.bindtextdomain("copy_paste_metadata",dt.configuration.config_dir.."/lua/locale/")
-
 local function _(msgid)
-    return gettext.dgettext("copy_paste_metadata", msgid)
+    return gettext.gettext(msgid)
 end
 
 local function copy(image)
--- a/tools/script_manager.lua
+++ b/tools/script_manager.lua
@@ -59,11 +59,8 @@ local debug = require "darktable.debug"
 local gettext = dt.gettext
 
 
--- Tell gettext where to find the .mo file translating messages for a particular domain
-gettext.bindtextdomain("script_manager",dt.configuration.config_dir.."/lua/locale/")
-
 local function _(msgid)
-    return gettext.dgettext("script_manager", msgid)
+    return gettext.gettext(msgid)
 end
 
 -- api check


TurboGit:
@wpferguson : All this is quite some work, so if you agree with the general direction I'd propose to concentrate on the two scripts in this PR. The main script manager and the copy_paste_metadata plug-in. When (if) we can get this working properly then we'll have a clear idea of the transition for all others plug-ins.

tools/script_manager.lua
```diff
@@ -59,11 +59,8 @@ local debug = require "darktable.debug"
 local gettext = dt.gettext
 
 
--- Tell gettext where to find the .mo file translating messages for a particular domain
-gettext.bindtextdomain("script_manager",dt.configuration.config_dir.."/lua/locale/")
```
wpferguson, line 63: same as above

official/copy_paste_metadata.lua
```diff
@@ -58,11 +58,8 @@ local publisher = ""
 local rights = ""
 local tags = {}
 
--- Tell gettext where to find the .mo file translating messages for a particular domain
-gettext.bindtextdomain("copy_paste_metadata",dt.configuration.config_dir.."/lua/locale/")
```
wpferguson, line 62: change this line to `gettext.bindtextdomain("darktable", dt.configuration.locale_dir)`

TurboGit: I may be missing something, but why is it better than removing the line? What did I broke?

wpferguson:
> Remove any spaces in start or end of strings, the layout must not be done in strings. I see for example "of " and some other cases.

The problem is the way we format output, i.e.`dt.print(_("export ")..tostring(number).." / "..tostring(total))`.  We could either .`dt.print(_("export") .. " " ..tostring(number).." / "..tostring(total))`or `dt.print(string.format(_("export %d/%d", number, total))`

TurboGit:
> dt.print(string.format(_("export %d/%d", number, total))

Yes, seems close to what is done on darktable core.

wpferguson:
Does this match what you translated?

script_manager
script_manager2
copypaste

official/copy_paste_metadata.lua
wpferguson: Actually nothing.  I just wasn't sure what Lua would inherit from darktable.  I tried it with and without the line and it works the same either way.  So, we can get rid of it.

TurboGit:
> Does this match what you translated?

Yes I recognize the few strings I have translated there. So it's working.

wpferguson:
I think the next step should be to clean up what we have, before we start adding new things.  

* remove bindtextdomain statements
* change gettext.dgettext to gettext.gettext
* change translatable statements with variables to string.format()

wpferguson:
@TurboGit how do want to proceed on this?  Do you want me to push what I have to your branch do you want to look at #455 and see if I've covered everything?

wpferguson:
@TurboGit I think this is how we should proceed on this:

* I'll add bindtextdomain back into #455 and merge it so we'll have better strings but it will still work.
* It's late in the cycle to potentially drop a new set of strings on the translators so I think we should aim for 5.0 and start early in the cycle so that the translators have time to translate.

wpferguson:
If I put bindtextdomain back, then the current translations we have will still work, minus the strings I changed.  

When we start 4.9, I'll bump the API then remove the bindtextdomain again and merge that into the development branch.  Anyone using master will get the development branch to develop and test against.

wpferguson:
@TurboGit are we ready to start on this again?

wpferguson:
> as in vacation :)

I hope you have a great time!

> I'm very motivated to see this project integrated in 5.0

I am too.  Once this is done I think we can bundle the lua-scripts as part of the build, which removes the git barrier.  I'll give the ones that want to download an override so they still can and stay up to date.

And I have some other goodies for when you get back :rofl:

wpferguson:
@TurboGit when we stopped I think we were close, but we were too close to release to drop the additional strings on the translators.  This is how I think we should resume:
* I'll pull all the translatable strings and give them another going over
* When you're ready, add the scripts back as a submodule
* After that I'll remove the bindtextdomain block again
* Generate the PO files with the scripts included

That should be all

wpferguson:
@TurboGit I've gone over the strings again and did some more cleaning and some refactoring.  I think we're ready to move forward.

wpferguson:
> Which means adding the sub-module, right?

Yes

wpferguson:
I looked at it last week and had the same thought :smile:

Done

PR #454, opened 2024-01-27
tools/script_manager - fixed crash

mime01:
The script manager crashes if a LUA script was not copied to a subfolder but directly into the LUA directory. The reason is that the regular expression, which among other things extracts the category of the script from its file path, does not match and thus the category is null. To prevent this, a null check has been added.

### DIFF
--- a/tools/script_manager.lua
+++ b/tools/script_manager.lua
@@ -497,14 +497,15 @@ local function process_script_data(script_file)
 
   -- add the script data
   local category,path,name,filename,filetype = string.match(script_file, pattern)
-  log.msg(log.debug, "category is " .. category)
-  log.msg(log.debug, "name is " .. name)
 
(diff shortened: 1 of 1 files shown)


wpferguson:
The fix works and solves the crashing problem.

What I'm wondering is what the "correct" way to handle this is.

Do we
* ignore the script because it's not in folder (what the fix does)
* should we assign it a category ("local" or "unbundled"?) and then ignore the category when we start the script
* if we just ignore it, should we pop a toast message to let the use know we saw the script but didn't include it in script_manager
* if we decide to include it, then we should probably add a toast message saying what we did.

Thoughts?

wpferguson:
I think I'll go with just the patch.  The only way a script can can be in the root folder is if the user puts it there. so they should know.

PR #464, opened 2024-05-16
Implement script metadata

wpferguson:
Added metadata fields to scripts that script_manager can access and display as a tooltip.

Added script_manager integration to those scripts that didn't have it.


wpferguson:
Partial fix for #438

PR #467, opened 2024-05-27
Support selecting multiple files in MacOS in OpenInExplorer.

koto:
In OpenInExplorer, use AppleScript instead of `open -R` for MacOS.

That way multiple files can be selected.


wpferguson:
Thanks

PR #476, opened 2024-06-09
Add contrib/fujifilm_auto_settings script

charludo:
This PR adds a script for autmatically applying fujifilm film emulations during import, based on the exif data contained within fuji raw images.

**Note that I am not the author of this script**. The actual author talks about the script [in this post](https://bastibe.de/2022-05-04-customizing-darktable-for-fujifilm-cameras.html), where they also link the required LUTs and styles. 

I'm opening this PR because I believe this script can be a very useful addition for many (fuji-) users. I don't know why the original author did not add the script to this collection, but the license appears to permit adding it.


bastibe:
Author of the script here.

I'm afraid the script is rather useless without the accompanying styles. All the script does is apply one or more styles based on the exif metadata. But without the styles, it does nothing at all. Is that even useful?

Also, the script version in this pull request is outdated. Please use the more recent version from https://github.com/bastibe/Fujifilm-Auto-Settings-for-Darktable. 

I have better versions of the styles in this repo: https://github.com/bastibe/Darktable-Film-Simulation-Panel

I'd be fine with shipping both the styles and the script in darktable, if that's desirable.

charludo:
> Author of the script here.
> 
> I'm afraid the script is rather useless without the accompanying styles. All the script does is apply one or more styles based on the exif metadata. But without the styles, it does nothing at all. Is that even useful?

Oh wow, thank you for jumping in on this!!

I would not call the script useless. Sure, it needs some additional styles, but "just" having the styles still leaves you with manual work which the script nicely automates away.

> Also, the script version in this pull request is outdated. Please use the more recent version from https://github.com/bastibe/Fujifilm-Auto-Settings-for-Darktable.
>
> I have better versions of the styles in this repo: https://github.com/bastibe/Darktable-Film-Simulation-Panel

I do not know why I didn't check for a repo containing the script 🤦🏼‍♀️ Thanks for letting me know. 

> I'd be fine with shipping both the styles and the script in darktable, if that's desirable.

Maybe the maintainers of this repo can jump in on this? For the time being I'll add the newer version of your script, but since you do have a repo with detailed instructions, I guess this PR isn't really that necessary any longer 😅 

> I have better versions of the styles in this repo: https://github.com/bastibe/Darktable-Film-Simulation-Panel

Btw, do these styles no longer need LUTs...?

bastibe:
Gee, thank you for your kind words!

I suppose we should add entries for the newer film simulations, Classic Neg, Nostalgic Neg, Eterna Bleach Bypass, and Reala Ace. We'd need to look up how they are represented in the exif data.

I can create new styles for the former three as well, and for Reala Ace as soon as they release the firmware update for the X-T5 that adds it. Or just use a Capture One or Lightroom render in the meantime.

PR #489, opened 2024-06-22
Rework executable path handling

wpferguson:
Updated executable path preferences so that executable_manager doesn't have to read the darktablerc file to get the executables and paths.

Changed check_if_bin_exists to use preferences first.  This should cut down on "flickering windows" on windows


wpferguson:
Depends on https://github.com/darktable-org/darktable/pull/17300

PR #495, opened 2024-07-13
fix(fujifilm_ratings.lua): improve the script and make it work again

LucasGGamerM:
I don't know what changed between 9.2.0 and 9.3.0, but it broke the previous iteration of the script. So I fixed it and improved it a bit. 

The main improvement here is that it also checks for JPGs, which I think is necessary. And it works again.

### DIFF
--- a/contrib/fujifilm_ratings.lua
+++ b/contrib/fujifilm_ratings.lua
@@ -44,7 +44,7 @@ local script_data = {}
 script_data.metadata = {
   name = "fujifilm_ratings",
   purpose = _("import Fujifilm in-camera ratings"),
-  author = "Ben Mendis <ben.mendis@gmail.com>",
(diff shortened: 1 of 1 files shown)


wpferguson:
> I don't know what changed between 9.2.0 and 9.3.0

There was an error in the libraries.  It was fixed around the time you submitted the PR, so the original code was working again.

> The main improvement here is that it also checks for JPGs

The improvement checks all JPGs, not just Fuji jpgs.  Since the name of the script is fujifilm_ratings users would get a rude surprise if all their JPGs had their ratings changed by the script.

If you want to write a script that corrects ratings for JPGs that darktable doesn't automatically detect then it should be as a separate script.  If you decide to do that, you probably want to check _image_.exif_maker to make sure that you are only changing ratings for maker images that are not understood by darktable.

LucasGGamerM:
Thank you got the feedback, I will see what I can do about it (for some reason my fuji rafs don't have a ratings, while the JPGs do), so maybe another script shall be written

PR #498, opened 2024-08-08
fixed translatable strings for inclusion into darktable

wpferguson:


wpferguson:
latest round of fixes for #450

PR #510, opened 2024-09-09
Fix string errors preventing translation

wpferguson:


wpferguson:
@TurboGit this fixes the list with the exception of EXIF stuff.  I'll do that separately and replace the variable functions with the string library versions.

PR #526, opened 2024-10-16
Substitution library updates

### LINKED ISSUE #525: avoid new translations and ensure proper language support
https://github.com/darktable-org/lua-scripts/blob/e3595676202b9c1bd57b494e0c895a673f848233/lib/dtutils/string.lua#L868

All those strings should add EXIF into the translation. Indeed in some language like French "EXIF year" will be translated to "Année EXIF".

I propose to keep the whole string as done in current dt code as this will avoid new translations.

`"$(EXIF.YEAR) - EXIF year"` is already translated in all languages, better keep this form.

This must be done for all similar strings.

wpferguson:
Fixes #525.

### DIFF
--- a/lib/dtutils/string.lua
+++ b/lib/dtutils/string.lua
@@ -557,6 +557,7 @@ local PLACEHOLDERS = {"ROLL.NAME",
                       "MONTH.SHORT",
                       "DAY", 
                       "HOUR", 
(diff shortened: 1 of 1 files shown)


wpferguson:
@TurboGit the strings match (were copied from) gui/gtkentry.c so there shouldn't be any new translations related to variables.

PR #536, opened 2024-11-14
Add gen_uuid() function to dtutils library

wpferguson:
Add a function to return a UUID string for use as a unique identifier

Partial fix for https://github.com/darktable-org/darktable/issues/16827


wpferguson:
This replaces the darktable.util.gen_uuid() function of https://github.com/darktable-org/darktable/pull/17795

PR #541, opened 2024-12-02
Fixed variable substitutions

koto:
According to comments, https://docs.darktable.org/usermanual/4.6/en/special-topics/variables/ and/or DT source code: `$(FILE.NAME)` and `$(ROLL.NAME)` should both be basenames.

Fixed error in `build_substitute_list` for images with duplicates, as the code referred to an undefined `version` property.


wpferguson:
Looks good to me, thanks

PR #544, opened 2024-12-15
Enhanced x-touch.lua for colorequalizer

MStraeten:
just adapted the colorzone stuff for color equalizer. requires explicitly set the focus via a further switch


wpferguson:
Thanks

PR #553, opened 2025-01-18
Update x-touch.lua switch colorequal tab

MStraeten:
added recommended additional shortcuts to switch between hue, saturation and brightness tabs: midi:D0=iop/colorequal/page;previous
midi:D#0=iop/colorequal/page;next


wpferguson:
Thanks

Issue #559, opened 2025-02-15
script_manager - don't start scripts until after all directories are scanned


wpferguson:
If you have libraries in multiple directories then a script in a early directory (starting with a, b, c, ...) may require a library in a later directory (x, y, z...) and not start because the library path hasn't been added yet.

wpferguson:
Fixed by #560

PR #560, opened 2025-02-16
script_manager - delay script start until all libraries are scanned

wpferguson:
added a start queue for scripts that start at start up to allow for including and using libraries in other than the default path.  Scripts are now started after all directories are scanned instead of  being started as they were found.


wpferguson:
Fixes #559

PR #563, opened 2025-02-23
Fix version comparison

sercxanto:
Currently this is a textual string comparison which will break with 10.0.0 API version.


wpferguson:
Could we change the name to `compare_API_versions`, so that people understand versions of what we are comparing.

Other than that looks good to me.  Thanks for catching it.

sercxanto:
> Could we change the name to `compare_API_versions`, so that people understand versions of what we are comparing.
> 
> Other than that looks good to me. Thanks for catching it.

Sure. I renamed the function.

PR #566, opened 2025-03-01
Fix dbmaint.lua

sercxanto:
"log.log_msg" triggers a lua error, because that function does not exist.

Additionally add add set_log_level / restore_log_level to improve compatibility with other files.


wpferguson:
Thanks

Issue #571, opened 2025-03-15
Bug in contrib/geoToolbox.lua - use of  for _,image in ipairs(sel_images) do  conflicts with function _(msgid)


fjb2020:
I wanted to use the contrib/geoToolbox script to tweak some GPS settings on some images, but it wasn't working.

Inspecting via ./darktable -d lua showed this error
    11.0105 LUA ERROR : /Volumes/EXT-Data/Data/darktable/lua/contrib/geoToolbox.lua:310: attempt to call a number value (local '_')

It seems that the use of 

    for _,image in ipairs(sel_images) do 

causes a conflict with the function _(msgid) 

I changed the 8 instances of 

    for _,image in ipairs(sel_images) do 

to 

    for jj,image in ipairs(sel_images) do

This fixed the error I was seeing and let me use the script as required.

System: 5.1.0+473~g51e22e438a (nightly) on macOS Sequoia 15.3.2 on M1 Max Studio

wpferguson:
If you want to do a PR to fix it that would be fine, otherwise I can do it.

wpferguson:
Fixed by #576.

PR #572, opened 2025-03-15
Fix issue #571, Bug in contrib/geoToolbox.lua - use of for _,image in ipairs(sel_images) do conflicts with function _(msgid)

fjb2020:
The use of

`for _,image in ipairs(sel_images) do `

causes a conflict with the function _(msgid)

All instances of

`for _,image in ipairs(sel_images) do `

changed to

`for jj,image in ipairs(sel_images) do`


wpferguson:
I'm not quite sure what went wrong when you created the PR, but the only file that should have changed was `contrib/geoToolbox.lua`.  

Deleting the icons will break `tools/script_manager.lua` and that wouldn't be a good thing.

fjb2020:
Oh sorry - no, that certainly wasn't deliberate - I'm not super-experienced with git so it will be user error somewhere - I'll try and find out what went wrong

wpferguson:
My guess is that you did a `git add `. and it picked up other things.  If you just `git add contrib/geoToolbox.lua`, then commit it should be fine.

PR #576, opened 2025-03-19
Fix issue #571, use of for _,image in ipairs(sel_images) do conflicts with function _(msgid)

fjb2020:
Change instances of 

`  for _,image in ipairs(sel_images) do`

to

`  for jj,image in ipairs(sel_images) do`


wpferguson:
Looks good to me, thanks

PR #577, opened 2025-04-01
Fix de_DE translation charset

kmilos:
This is currently showing up as an issue in Debian CI: https://i18n.debian.org/l10n-pkg-status/d/darktable.html


wpferguson:
I guess I'm not sure why Debian CI is even scanning the lua-scripts directory.  It's not part of the build.  The only reason it exists, for now, is to populate the translation files.

However, it's not a big deal and I'll merge and push a submodule update.

Thanks.

PR #578, opened 2025-04-04
adds fix_sigma_foveon_dngs script

steph72:
Adds a lua script as a workaround for https://github.com/darktable-org/darktable/issues/16586
cf. https://schallundstille.de/2025/04/04/fixing-foveon-dngs-in-darktable/

### DIFF
--- a/contrib/fix_sigma_foveon_dngs.lua
+++ b/contrib/fix_sigma_foveon_dngs.lua
@@ -1,0 +1,110 @@
+--[[ fix-sigma-foveon-dngs-v0.1
+
+Fix Sigma Foveon DNGs white balance handling
+
+Copyright (C) 2025 Stephan Kleinert <stephan@schallundstille.de>
+
+This program is free software; you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation; either version 2 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along
+with this program; if not, write to the Free Software Foundation, Inc.,
+51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+]]
+
+
+--[[About this Plugin
+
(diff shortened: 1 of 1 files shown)


contrib/fix_sigma_foveon_dngs.lua
```diff
@@ -0,0 +1,110 @@
...
+    purpose = _("Workaround for Sigma Foveon DNG white balance problems"),
+    author = "Stephan Kleinert <stephan@schallundstille.de>",
+    help = "https://schallundstille.de/"
+}
+
+script_data.destroy = nil -- function to destory the script
+script_data.destroy_method = nil -- set to hide for libs since we can't destroy them commpletely yet, otherwise leave as nil
+script_data.restart = nil -- how to restart the (lib) script after it's been hidden - i.e. make it visible again
+script_data.show = nil -- only required for libs since the destroy_method only hides them
+
```
wpferguson, line 65: create a namespace to contain the global variables and functions and then prefix sad_cameras, dt_fixed_tag and containse with it.  

for example...

fsfd = {}

fsfd.sad_cameras = 
fsfd.dt_fixed_tag =
function fsfd.contains(

contrib/fix_sigma_foveon_dngs.lua
```diff
@@ -0,0 +1,110 @@
...
+        return
+    end
+    if contains(dt.tags.get_tags(image), dt_fixed_tag) then
+        dt.print_log("image already foveon_fixed")
+        return
+    end
+    fix_image()
+    dt.tags.attach(dt_fixed_tag, image)
+end
+
```
wpferguson, line 107: create a destroy function that destroys the darkroom-image-loaded event

local function destroy()
  dt.destory_event("fix_sigma_foveon_dngs", "darkroom-image-loaded")
end

and then

scrpt_data.destroy = destroy

This lets the script be stopped and restarted without causing a problem because the event already exists.

contrib/fix_sigma_foveon_dngs.lua
```diff
@@ -0,0 +1,110 @@
...
+dt.register_event("fix_sigma_foveon_dngs", "darkroom-image-loaded", fix_on_load)
+
+dt.print_log("[fix_sigma_foveon_dngs] loaded")
```
wpferguson, line 110: return script_data at the end of the script so that script_manager has access to the metadata and the destroy function.

wpferguson (changes requested):

PR #579, opened 2025-04-09
Align sequence tooltip with code.

tabac:
I believe the tooltip is not aligned with the code. 

Based on [this](https://github.com/darktable-org/lua-scripts/blob/master/lib/dtutils/string.lua#L997) we expect the start number first and the width second. I didn't change the string match pattern, it's not clear to me if we want it the other way around of not.


tabac:
Oh, I didn't notice that, I can update the PR if you want.

wpferguson:
If you want to update the PR that would be great.  Thanks

wpferguson:
Looks good to me, thanks.

PR #584, opened 2025-05-04
Add manual coordinate input option to geoToolbox

balping:
Previously you could only copy/paste GPS data from one picture to another. This PR makes it possible to manually enter coordinates and apply them to a picture.

### DIFF
--- a/contrib/geoToolbox.lua
+++ b/contrib/geoToolbox.lua
@@ -1,6 +1,7 @@
 --[[
     This file is part of darktable,
     copyright (c) 2016 Tobias Jakobs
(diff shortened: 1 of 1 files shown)


supertobi:
@wpferguson The code looks good and I loke the idea, but I don't have a dt here to test. Please merge if it works.

wpferguson:
Should the entry fields have a tooltip to let the user know they can manually enter the values?

supertobi:
Sounds like a good idea.

balping:
@wpferguson I've added tooltips and I've also renamed "paste GPS data" to "apply GPS data" to avoid confusion.

PR #585, opened 2025-05-24
Add a script to automatically tag manually modified images

mreiger:
This PR adds a script to automatically tag images which have been manually modified in the darkroom.

The reason for this script: I wanted to easily distinguish between edited and un-edited images; and with the apply_camera_style script enabled I can not use darktable's history filter because all images show up as changed.

Based on the suggestion from @wpferguson in this pxls.us thread: https://discuss.pixls.us/t/filter-for-images-with-auto-applied-camera-styles-but-no-manual-processing/50260/1 and I also used his auto_snapshot script as template.


mreiger:
Documentation: https://github.com/darktable-org/luadocs/pull/17

PR #610, opened 2025-11-24
official/selection_to_pdf

### LINKED ISSUE #609: selection_to_pdf.lua open pdf after generating
The script uses xdg-open, which is a Linux thing to open the pdf after running.  On macos it probably needs to be just open.

wpferguson:
added default opens for Linux, MacOS, and Windows

Fixes #609


blabberlicious:
Thank you for the speedy fix!

PR #631, opened 2026-02-03
Updated script_manager for bundled scripts

wpferguson:
Made the necessary changes to run the lua-scripts when bundled with the darktable distribution.

User installed scripts will always take priority over the bundled scripts.

This version of script_manager is targeted at Lua API 9.7.0, but will work with 9.6.0 and earlier


scorpi11:
Tested with the darktable changes from https://github.com/darktable-org/darktable/pull/19899, works fine with system-wide lua scripts.

PR #645, opened 2026-03-25
use_paired_jpg_as_mipmap

### LINKED ISSUE #19470: [FR] Option to force embedded JPEG in full-preview layout
**Is your feature request related to a problem? Please describe.**

When culling RAW files, lighttable initially shows the embedded JPEG preview for thumbnails but switches to the unprocessed RAW data in full-preview layout (F) after a short delay. This introduces:
- Unnecessary latency (RAW decoding is slower than displaying the embedded JPEG).
- Poor visual quality for culling: Unprocessed RAW files appear flat, dark, or noisy compared to the camera-generated JPEG, which has in-camera contrast/sharpening applied.

**Describe the solution you'd like**

Add a preference setting (in Preferences → Lighttable) to:
Use the embedded JPEG preview not only for thumbnails but also for full-screen mode (F) until the first darkroom edit is applied to the file.

**Alternatives**

- Import RAW+JPEG pairs, use Lua script to set JPEG as group leader, cull, then switch back to RAW as leader.
Downside: Requires duplicate files (see #17149) 

**Additional context**
Consistent with existing thumbnail logic, where users can already choose to prefer embedded JPEGs until darkroom edits are applied.
No overhead: The JPEG is already extracted for thumbnails.

wpferguson:
Canon R series cameras don't include a full size embedded jpeg so generating mipmap cache for anything larger than mipmap 3 requires the raw to be loaded and processed.  This script copies the paired jpeg from the RAW+JPEG pair to the full resolution mipmap.  This can then be used to generate the remaining mipmap cache.  The jpeg image can be discarded after being copied to the cache or retained based on a preference in Lua options.

Fixes https://github.com/darktable-org/darktable/issues/19470

### DIFF
--- a/official/use_paired_jpg_as_mipmap.lua
+++ b/official/use_paired_jpg_as_mipmap.lua
@@ -277,7 +277,7 @@ local function process_images(images, count)
   if not pmj.keep_jpgs then
     refresh_collection()
   end
-  dt.util.message(MODULE, "responsive_cache", "build cache")
+  dt.util.message(MODULE, "broadcast", "finished")
 end
 
 -- - - - - - - - - - - - - - - - - - - - - - - - 
--- a/official/use_paired_jpg_as_mipmap.lua
+++ b/official/use_paired_jpg_as_mipmap.lua
@@ -201,7 +201,7 @@ local function refresh_collection()
 end
 
 local function get_mipmap_dir()
-  local old_log_level = set_log_level(pmg.log_level)
+  local old_log_level = set_log_level(pmj.log_level)
   local mipmap_dir = nil
   local cachedir = dt.configuration.cache_dir
 
@@ -231,7 +231,7 @@ local function stop_job(job)
 end
 
 local function process_images(images, count)
-  local old_log_level = set_log_level(pmg.log_level)
+  local old_log_level = set_log_level(pmj.log_level)
   
   -- if dt.collection is equal to count there are no
   -- RAW+JPEG pairs
--- a/official/use_paired_jpg_as_mipmap.lua
+++ b/official/use_paired_jpg_as_mipmap.lua
@@ -175,11 +175,11 @@ end
 local function pref_read(name, pref_type)
   local old_log_level = set_log_level(namespace.log_level)
 
-  log.msg(log.debug, "name is " .. name .. " and type is " .. pref_type)
+  log.msg(log.debug, MODULE .. ": name is " .. name .. " and type is " .. pref_type)
 
   local val = dt.preferences.read(MODULE, name, pref_type)
 
-  log.msg(log.debug, "read value " .. tostring(val))
+  log.msg(log.debug, MODULE .. ": read value " .. tostring(val))
 
   restore_log_level(old_log_level)
   return val
@@ -188,7 +188,7 @@ end
 local function pref_write(name, pref_type, value)
   local old_log_level = set_log_level(namespace.log_level)
 
-  log.msg(log.debug, "writing value " .. tostring(value) .. " for name " .. name)
+  log.msg(log.debug, MODULE .. ": writing value " .. tostring(value) .. " for name " .. name)
 
   dt.preferences.write(MODULE, name, pref_type, value)
 
@@ -201,6 +201,7 @@ local function refresh_collection()
 end
 
 local function get_mipmap_dir()
+  local old_log_level = set_log_level(pmg.log_level)
   local mipmap_dir = nil
   local cachedir = dt.configuration.cache_dir
 
@@ -215,11 +216,13 @@ local function get_mipmap_dir()
     for line in p:lines() do
       if line:len() > 4 then
         mipmap_dir = cachedir .. PS .. line
+        log.msg(log.info, MODULE .. ": mipmap_dir is " .. mipmap_dir)
       end
     end
     p:close()
   end
 
+  restore_log_level(old_log_level)
   return mipmap_dir
 end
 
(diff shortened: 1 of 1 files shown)


jdchristensen:
responsive_cache is working for me, but there are a couple of minor issues:

1. If you unload the script, you can't load it again in the same darktable session.  If you try, you get `duplicate index name responsive_cache for event type post-import-film`.  So some kind of releasing of resources needs to happen.
2. After it runs, it says it created the mipmaps (2 and 5 in my case), but they are not written to disk.  However, if I do something to use up memory, like previewing many images at 100%, then the 2 and 5 level caches get written to disk.  But if I exit darktable before doing so, they are never written to disk.  Is there a way to flush these to disk right after creating them?

jdchristensen:
@wpferguson Not sure if you are planning a PR for response_cache.lua at some point.  I have attached an updated version.  Changes compared to the version I attached on March 30:
- Adds things to the destroy method so you can unload and reload this module.
- Fixes a typo:  "reciever" -> "receiver".
- When you exit the darkroom, it also regenerates the full size mipmap, if that preference is set.
- It gets rid of some debug code.
```diff
--- /tmp/chrome/responsive_cache2.lua   2026-03-30 19:20:58.000000000 -0400
+++ responsive_cache_jdc.lua    2026-04-01 10:04:34.990204638 -0400
@@ -179,14 +179,6 @@
   return 0
 end
 
--- Code for testing the above on import
-local test_widths = {179, 180, 181, 359, 360, 361, 1919, 1920, 1921, 5121, 7681}
-for i = 1, #test_widths do
-    local w = test_widths[i]
-    local m = cache_size(w)
-    dt.print_log("For width " .. w .. " got mip " .. m .. " with width " .. CACHE_SIZES[m+1])
-end
-
 local function update_settings()
 
   log.log_level(rc.log_level)
@@ -416,8 +408,13 @@
   end
 --]]
   log.msg(log.info, "updating just developed image " .. image.filename)
-  local size = cache_size(rc.width / rc.lt_zoom)
-  log.msg(log.debug, "generating updated thumbnail size " .. size)
+  if rc.full_size then
+    local size = rc.full_preview_size
+    log.msg(log.info, "generating updated thumbnail size " .. size)
+    image:generate_cache(true, size, size)
+  end
+  local size = rc.thumbtable_size
+  log.msg(log.info, "generating updated thumbnail size " .. size)
   image:generate_cache(true, size, size)
   log.msg(log.debug, "generating new filmstrip thumbnail")
   image:generate_cache(true, 0, 0)
@@ -442,7 +439,7 @@
         generate_cache(tonumber(start), tonumber(stop), tonumber(mipmap_size))
       end
     end
-  elseif sender == "log_tool" and reciever == MODULE then
+  elseif sender == "log_tool" and receiver == MODULE then
     if message == "default" then
       rc.log_level = DEFAULT_LOG_LEVEL
     elseif message == "debug" then
@@ -486,9 +483,15 @@
 local function destroy()
   dt.destroy_event(MODULE, "post-import-film")
   dt.destroy_event(MODULE, "post-import-image")
+  dt.destroy_event(MODULE .. "_collect", "shortcut")
+  dt.destroy_event(MODULE .. "_select", "shortcut")
   dt.destroy_event(MODULE, "darkroom-image-history-changed")
   dt.destroy_event(MODULE, "inter-script-communication")
   dt.destroy_event(MODULE, "exit")
+  dt.preferences.destroy(MODULE, "screen_width")
+  dt.preferences.destroy(MODULE, "cull_first")
+  dt.preferences.destroy(MODULE, "full_size")
+  dt.preferences.destroy(MODULE, "build_cache_on_import")
 end
 
 script_data.destroy = destroy
```
[responsive_cache.txt](https://github.com/user-attachments/files/26546069/responsive_cache.txt)

wpferguson:
Don't destroy preferences and shortcuts when exiting because you lose the settings/mappings so you would have to re-enter all of that every time you restart.

You can wrap shortcuts in 

```lua
if not dt.query_event(MODULE .. "_collect", "shortcut) then
  -- register the event
end
```

IIRC preferences aren't a problem.

jdchristensen:
Ah, good to know.  I have removed the shortcuts and preferences from the destroy routines, and protected the registration of shortcuts as you indicated.  I think similar protection must be needed for preferences too, as each time I unload and reload, an additional set of preferences is added to the list.  Maybe this is something that should be fixed in the C code?  If a lua module registers preferences that already exist, ignore the request and return an error?  (While debugging a script, it's handy to unload and reload it, and in any case, since the UI offers it, we should make it work correctly.)

Should we discuss this script in a separate PR, so we can use the proper github tools for commenting on lines, seeing what changed, etc?  If you want me to create it, I'm happy to, but it's your script, so I thought you might prefer to do it.  (But then I won't be able to push to it.)

[responsive_cache.txt](https://github.com/user-attachments/files/26547929/responsive_cache.txt)

wpferguson:
> Should we discuss this script in a separate PR

Yes

> I think similar protection must be needed for preferences too

The only time it would be an issue is when you're starting and stopping the script repeatedly.  In that instance, I would say just use destroy, because you want to start clean.  In normal usage I would leave the preferences, so that they would remain if the user decides to turn the script back on at a later date/invocation of darktable.  I don't really see users turning it on and off multiple times, unless they are developing.

jdchristensen:
Actually, even with destroy, the preferences get duplicated.  I think the lua API should not duplicate preferences that already exist.  But maybe there is a workaround a script can use to only add preferences if they aren't already there?  I didn't see a way to check for this.

Issue #627, opened 2026-01-05
function "dtutils.string.substitute()" fails when no time data is provided in Exif data


jeanrenaud:
When running a script using "dtutils.string.substitute()" function on a file without any time data in Exif, the "dtutils.string.substitute()" function fails.

```
190.9832 LUA ERROR : ...\AppData\Local\darktable/lua/lib/dtutils\string.lua:748: time result cannot be represented in this installation
stack traceback:
	[C]: in ?
	[C]: in function 'os.time'
	...AppData\Local\darktable/lua/lib/dtutils\string.lua:748: in upvalue 'exiftime2systime'
	...AppData\Local\darktable/lua/lib/dtutils\string.lua:843: in function 'lib/dtutils.string.build_substitute_list'
	...AppData\Local\darktable/lua/lib/dtutils\string.lua:1298: in function 'lib/dtutils.string.substitute'
``` 

After looking at the "string.lua" code, it seems that "build_substitute_list" function tries to use "0000:00:00 00:00:00" as default value if the exif time is not available, but the "os.time{}" conversion called later through "exiftime2systime(datetime_taken)" does not like it. The minimum time value accepted on several systems is January 1st, 1970. I updated the "string.lua" code on my installation (replacing "0000:00:00 00:00:00" with "1970:01:01 00:00:00") and this fixes the issue.

Code from string.lua :

```lua
local function exiftime2systime(exiftime)
  local yr,mo,dy,h,m,s = string.match(exiftime, "(%d-):(%d-):(%d-) (%d-):(%d-):(%d+)")
  return(os.time{year=yr, month=mo, day=dy, hour=h, min=m, sec=s})
end

[...]

function dtutils_string.build_substitute_list(image, sequence, variable_string, username, pic_folder, home, desktop)
[...]
  if image.exif_datetime_taken and image.exif_datetime_taken ~= "" then
    datetime_taken = image.exif_datetime_taken
  else
    if use_millisecs then
      datetime_taken = "0000:00:00 00:00:00.0"
    else
      datetime_taken = "0000:00:00 00:00:00"
    end
  end
[...]
``` 

I see this issue on Windows 11, DarkTable 5.4.0. Other systems could behave the same, or not.

deekayhd:
@jeanrenaud This issue has been fixed. Maybe you would like to have a look at it and close this issue, then.

PR #653, opened 2026-03-29
fix wrong month in EXIF.MONTH.LONG and EXIF_MONTH.SHORT when no exif date available in image

deekayhd:
PR #594 fixed a bug in rename_images.lua with images that have no EXIF datetime. However, the fix (setting datetime_taken to 0000.00.00) does not work on Windows, because the date is smaller than what can be represented on this platform. 
In addition, on linux the date 0000.00.00 results in EXIF.MONTH.LONG and EXIF.MONTH.SHORT as "November" and "Nov", respectively.
In this PR I propose to skip the calculation of EXIF.MONTH.LONG and .SHORT for images without a date, and set them as follows:
EXIF.MONTH.LONG = "unknown"
EXIF.MONTH.SHORT = "???"

closes issue https://github.com/darktable-org/lua-scripts/issues/627

### DIFF
--- a/lib/dtutils/string.lua
+++ b/lib/dtutils/string.lua
@@ -793,6 +793,7 @@ function dtutils_string.build_substitute_list(image, sequence, variable_string, 
         string.match(datetime_taken, "(%d+):(%d+):(%d+) (%d+):(%d+):(%d+)$")
     end    
   else
+    log.msg(log.warn, "no capture datetime found in EXIF data")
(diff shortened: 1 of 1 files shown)


deekayhd:
Changes done as you requested.
One question about the logging: The default log level in string.lua is error, so a warning is never shown unless the user changes the source code. Is this intended or am I missing something? Is there a way to override the defaul log level?

PR #663, opened 2026-04-10
fix incomplete removal of datetime_taken in PR #653

### LINKED ISSUE #661: Rename image script fails when calling `build_substitute_list`
I am trying to rename using the image renamer. It used to work perfectly well approx 2-3 months ago. May be related to Darktable update?
Running Darktable 5.4.1 on macOS 26.4.

```
    25.6720 LUA ERROR : [HOME]/.config/darktable/lua/lib/dtutils/string.lua:792: bad argument #1 to 'match' (string expected, got nil)
stack traceback:
	[C]: in ?
	[C]: in function 'string.match'
	[HOME]/.config/darktable/lua/lib/dtutils/string.lua:792: in function 'lib/dtutils.string.build_substitute_list'
	.../.config/darktable/lua/contrib/rename_images.lua:144: in upvalue 'do_rename'
	.../.config/darktable/lua/contrib/rename_images.lua:211: in function <.../.config/darktable/lua/contrib/rename_images.lua:210>
	[C]: in ?
```

No matter what pattern I define or if am trying to rename JPEG or ARW file, I am getting the error above.

deekayhd:
Fixes a bug introduced with PR #653. Variable datetime_taken in string.lua, lines 788 and 792 is not used anymore and therefore not initialized correctly.

close #661.

### DIFF
--- a/lib/dtutils/string.lua
+++ b/lib/dtutils/string.lua
@@ -890,7 +890,7 @@ function dtutils_string.build_substitute_list(image, sequence, variable_string, 
                         home_folder,                           -- FOLDER.HOME
                         desktop_folder,                        -- FOLDER.DESKTOP
                         "",                                    -- OPENCL.ACTIVATED - wont be implemented
-                        user_name,                             -- USERNAME
+                        (user_name ~= nil) and user_name or "",-- USERNAME
                         "",                                    -- NL - wont be implemented
                         ""                                     -- JOBCODE - wont be implemented
   }
--- a/lib/dtutils/string.lua
+++ b/lib/dtutils/string.lua
@@ -785,11 +785,11 @@ function dtutils_string.build_substitute_list(image, sequence, variable_string, 
     emon_short = os.date("%b", exiftime2systime(image.exif_datetime_taken))
     if use_millisecs then
       eyear, emon, eday, ehour, emin, esec, emsec = 
-        string.match(datetime_taken, "(%d+):(%d+):(%d+) (%d+):(%d+):(%d+)%.(%d+)$")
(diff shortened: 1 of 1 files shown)


wpferguson:
Fixes https://github.com/darktable-org/darktable/issues/20787

fabiangebert:
I've checked it out locally and can confirm the rename op working again. Thank you!

lib/dtutils/string.lua
wpferguson: This is wrong.  emon_long gets "" and everything else gets nil

Should be 

`emon_long, emon_short, eyear, emon, eday, ehour, emin, esec, emsec = "", "", "", "", "", "", "", "", ""`

wpferguson:
@deekayhd here's a test script, 
[test_date_time_substitution.zip](https://github.com/user-attachments/files/26642680/test_date_time_substitution.zip)


Start darktable, start the script, and  assign a shortcut to it.  In a collection hit the shortcut and it will build a substitution string on EXIF.MONTH.LONG for the first 10 images.  The 11th will get an empty date time.  What the script will do is set debug logging on it the string library and spit out a list of all the variable substitutions (which is how I found the nil instead of "" in the e... variables.

lib/dtutils/string.lua
deekayhd: Yes, you are right. I swear I tested it, and fortunately, this did not cause a runtime error, because there is a check in lines 1083 to 1091. 
But, of course, it is safer to initialize it properly.
fixed,

deekayhd:
Thanks for your thorough testing. With the script, I found that $(USERNAME) could come out as nil, as well, if not provided in the call of method substitute(). Fixed that, too.

wpferguson:
Looks good to me.  Thanks

PR #664, opened 2026-04-12
fix: autostyle deadlock on import and deprecated API removal for Windows

2421468125:
Problems:

1. Calling darktable.styles.apply() directly inside the post-import-image callback causes darktable to hang/deadlock.

2. darktable.control.read() has been removed from the Lua API in recent versions (field "read" not found for type dt_lua_singleton_control), causing the script to fail on startup.

3. Windows cmd.exe only recognizes double quotes while exiftool_attribute() use single quotes for path. 

Fix:

- Wrap the post-import-image callback logic in darktable.control.dispatch() + darktable.control.sleep(500) so that style application runs asynchronously in a background thread after the import pipeline has finished. This avoids the deadlock while still supporting both single-image and batch imports.

- Remove the darktable.control.read(fd) call from get_stdout(). This function was only used to yield the Lua thread while waiting for I/O — it did not perform any data reading itself (fd:read('*a') handles that). Since the entire exiftool call now runs inside a dispatched background task, blocking the background thread is acceptable and does not freeze the UI.

- Change path quoting from single to double quotes.

Tested on darktable 5.4.1 / Windows with Canon CR3 files.

### DIFF
--- a/contrib/autostyle.lua
+++ b/contrib/autostyle.lua
@@ -71,10 +71,10 @@ script_data.show = nil -- only required for libs since the destroy_method only h
 local function get_stdout(cmd)
   -- Open the command, for reading
   local fd = assert(io.popen(cmd, 'r'))
-  darktable.control.read(fd)
+  -- darktable.control.read(fd)
   -- slurp the whole file
   local data = assert(fd:read('*a'))
-
+  -- darktable.print(string.format(_("data is %s"), data))
   fd:close()
   -- Replace carriage returns and linefeeds with spaces
   data = string.gsub(data, '[\n\r]+', ' ')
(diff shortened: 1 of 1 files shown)


wpferguson:
The better way to handle this would have been to open an issue describing the problem and the steps to reproduce.   That way a discussion could have taken place about the actual problem and ways to solve it. 

Some of the assumptions above are wrong and while the provided solution will work on Windows, it will break MacOS and Linux. 

`darktable.control.read` is not deprecated.  It is not available on Windows due to issues with Windows and the MSYS2 environment. 

There is a string library with sanitizing functions that are OS aware.

After an import is complete darktable raises a post-import-film event.  Images can be "buffered" by collecting them into a table, then processing them in a job when the post-import-film event is caught.

2421468125:
Thank you for the detailed feedback — I appreciate the corrections.
You're right that I should have started with an issue first.  
I'll close this PR.

PR #666, opened 2026-04-18
Order functions in dependency order in group_persistance.lua

bkhl:
The call to `add_to_group` in `build_group` failed as `add_to_group` was a local function defined later. This reorders the functions to avoid that.


wpferguson:
Thanks

PR #667, opened 2026-04-18
Missing argument in group_persistence.lua

bkhl:
`add_to_group` was still failing as the leader was not provided as second argument.


wpferguson:
Thanks

PR #677, opened 2026-05-16
Fix wrong colors in ai_raw_denoise.lua

andriiryzhkov:
This one was trickier than expected. The bayer rawdenoise model needs more preprocessing than the original script was doing, and the Lua tensor API didn't expose the per-channel ops needed to do it.

### The two bugs

1. **Missing white balance.** The model is trained on daylight-WB'd input. The script was feeding it un-WB'd camRGB, so the model got wildly unbalanced R/G/B and produced wrong colors on every camera.
2. **CFA-blind packing.** `bayer_pack` splits 2×2 blocks by position only, not by CFA pattern. On non-RGGB sensors (BGGR Leica, GRBG Canon, etc.) plane 0 was G or B – the model thought it was R and chaos followed.

### What changed

- `fc()` – pure-Lua FC channel lookup.
- `daylight_wb()` – D65 daylight WB from `meta.color_matrix`, same formula as the C path.
- RGGB origin shift: find the R cell within the 2×2 starting at `(crop_y, crop_x)` and re-crop so plane 0 always lands on R.
- Apply `{wb_R, 1, 1, wb_B}` before inference, invert `{1/wb_R, 1, 1/wb_B}` after, so the saved DNG is un-WB'd camRGB.
- Bail out with a clear error on X-Trans / monochrome instead of crashing downstream.

### Dependency

Needs a new `scale_add_planes` method on the Lua tensor binding (separate PR against darktable). The existing API has no per-channel arithmetic and no element-wise multiply, so applying per-channel WB without it isn't really possible.

### DIFF
--- a/examples/ai_denoise.lua
+++ b/examples/ai_denoise.lua
@@ -312,9 +312,9 @@ dt.register_lib(
 local script_data = {}
 
 script_data.metadata = {
-  name = "ai_denoise",
+  name = "AI denoise",
   purpose = "tile-based AI denoise using the darktable.ai Lua API",
-  author = "Andrii Ryzhkov <andrii.ryzhkov@pm.me>",
+  author = "Andrii Ryzhkov",
   help = "https://docs.darktable.org/lua/stable/lua.scripts.manual/scripts/examples/ai_denoise"
 }
 
--- a/examples/ai_raw_denoise.lua
+++ b/examples/ai_raw_denoise.lua
@@ -91,6 +91,37 @@ local OVERLAP = 32
 -- F U N C T I O N S
 -- - - - - - - - - - - - - - - - - - - - - - - -
 
+-- Bayer-only CFA channel lookup. Returns 0/1/2 (R/G/B) for the cell
+-- at sensor position (r, c) given the 32-bit `filters` mask from
+-- load_raw's meta table. Used to find the R-cell origin so bayer_pack's
+-- plane 0 always holds R. X-Trans (filters == 9) uses meta.xtrans
+-- instead and is not handled here.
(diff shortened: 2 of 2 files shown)


wpferguson:
Just tested and still getting (mostly) all black images.  Tried a CR2 and CR3 and they both came back totally black.  Tried a NEF night shot and it had some colored pixels.

andriiryzhkov:
@wpferguson : Did you update models to the latest?

wpferguson:
@andriiryzhkov That was the problem.  I had to rename the models directory and then do a fresh reload.  

A thought.  Maybe the models should be put in the cache_dir.  I'm running 40+ installations of darktable (daily driver, a couple of old versions, 23 development versions and 16 test versions) and they are all sharing a common AI model directory (~/.local/darktable/models).  I wrap the various installations in a script that points to unique configuration/cache/image directories so that I can wreak havoc in a contained space and not break anything important or pollute another environment.

andriiryzhkov:
@wpferguson : well, having shared models folder was a design choice, so with separate workspaces there will be no need to install once again.

For the case you are describing, probably it would be good to have optional config key that can override default model folder location.

wpferguson:
> For the case you are describing, probably it would be good to have optional config key that can override default model folder location.

That would work.  I have scripts that build the environment and pre-condition the darktablerc file so I could just "fill in the blank" when I need to and not need a command line switch or even a preference.

andriiryzhkov:
> That would work.

Small update that implements this customization. Not sure how it fits feature freeze though.

https://github.com/darktable-org/darktable/pull/21049

PR #688, opened 2026-06-20
Check for the location of the website gallery data files

scorpi11:
First check for ~/.config/darktable/lua/, as these may be more recent. If this does not exist, use the system-wide lua-scripts installation.


wpferguson:
Thanks

PR #689, opened 2026-06-22
Prevent endless loops if luarc requires script_manager

masterpiga:
if the `luarc` in the configdir requires the script manager, this will cause an endless loop and crash the lua interpreter. I know it because it just happened to me after deleting the `lua` subdir in my config dir :smile:

### DIFF
--- a/tools/script_manager.lua
+++ b/tools/script_manager.lua
@@ -47,6 +47,16 @@
 
 ]]
 
+-- re-entrancy guard
+--
+-- when script_manager runs from the system (bundled) lua-scripts directory it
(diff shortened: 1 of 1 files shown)


masterpiga:
@wpferguson

wpferguson:
I worried about that happening but I was never able to find a scenario where it would :-/

masterpiga:
For me it was very easy to hit that scenario 😅 I just deleted the `<config_dir>/lua` without deleting `<config_dir>/luarc` and kaboom!

wpferguson:
Works for me, thanks

PR #542, opened 2024-12-02
New Immich export plugin

brainz73:
Hi, I wrote an export plugin for Immich, an open source photo management software that is becoming very interesting:

https://immich.app/

I would like to contribute to the Darktable project by sharing the script.

PR #573, opened 2025-03-16
Add Immich upload script for direct image uploads to Immich server

guigodin:
Hey !

I created a script to upload directly to an immich server that I've been using almost daily now for a few weeks. I figured that maybe you'd be interested in adding to this repo. Let me know if any change is required.

PR #692, opened 2026-06-25
Immich exporter (update of #542)

clnhlzmn:
This is based on #542 ([diff](https://github.com/clnhlzmn/lua-scripts/compare/a6b1455...immich)) which seems to be not getting the attention it needs.

Full disclosure: I don't know Lua and I have not done any previous work on DT plugins so I had Claude Code assist me with this. Not sure if there are policies against that but I don't want to present this as purely my work (though I did at least have my hand on the wheel).

I tried to address the comments in the original PR, and since I'm on MacOS I added a preference for setting the Lua path. Admittedly this was a suggestion from Claude based on the "understanding" that on MacOS (and Windows?) DT embeds the Lua interpreter so Lua dependencies installed to the system Lua aren't picked up automatically. This change was an attempt to address that. I'm happy to pivot this if the "understanding" was flawed or there is simply a better way to do it.

For what it's worth I have tested it locally (on macOS) with my immich server and it seems to work.

Also @brainz73 I left your name in the `script_data.author` field. I'm assuming that means you'd be the one to blame on new issues. Please let me know if you want me to put my name there instead. I'm happy to do so.

### DIFF
--- a/contrib/immich.lua
+++ b/contrib/immich.lua
@@ -1,0 +1,335 @@
+--[[
+  This file is part of darktable,
+  copyright (c) 2024 Giorgio Massussi
+
+  darktable is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  darktable is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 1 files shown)


clnhlzmn:
Wow I just noticed there is another Immich export PR (#573) that I didn't notice at first.. Not trying to step on anyone's toes I just want something to get merged

clnhlzmn:
@guigodin I wish I had noticed your PR first because I think using the immich CLI is the better way to do this...

It just occurred to me to check on the upcoming immich v3 changes and it looks like there will be some issues with this script that using the CLI probably (hopefully?) sidesteps.. so I think I don't actually want to merge this as is

PR #694, opened 2026-06-25
Immich uploader round 2 (or 3 or 4?) 

clnhlzmn:
I was originally inspired by #542 which led me to make #692.

Then it occurred to me that doing raw HTTP requests would be fragile when Immich makes breaking API changes (coming soon with v3). So I abandoned that PR.

This PR is based on #573, which like #542 seems to be abandoned, because I think it takes a better approach of using the Immich CLI instead of raw API requests.

Like I said in #692 I'm not a Lua or DT plugin dev so I am using Claude Code to help with this. I know some people don't love that. I get it. I'm just trying to make something that works and is helpful for people (me mostly). Don't expect this to escape critical review and I'm happy to work with the DT plugin community to make this tolerable.

[Here](https://github.com/clnhlzmn/lua-scripts/compare/7cc6b1dfb2f1e37d6dd1646619c98e635d1801bc...immich) is a diff to @guigodin's MR (#573). Also @guigodin I put my name in the `script_data.plugin` field since you left it blank.

I guess I should also say this seems to work flawlessly for me on macOS with DT 5.6 and Immich 2.7.5.

### DIFF
--- a/contrib/immich_upload.lua
+++ b/contrib/immich_upload.lua
@@ -1,0 +1,377 @@
+--[[
+    copyright (c) 2024 Guillaume Godin <godin.guillaume@gmail.com>
+
+    This program is a free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 2 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+]]
+--[[
+  immich_upload.lua - Immich storage for darktable
+
+  Immich is a self-hosted photo management service that provides features similar to Google Photos.
+  This script adds a new storage option to upload exported photos directly to an Immich server without keeping a local copy.
+  It uses the immich-cli tool to upload the images. It has 2 configurable preferences:
+  * Immich server URL (default: http://localhost:2283)
+  * Immich API key (default: empty)
+  The script will appear as "Immich Upload" in the export module storage list. It can be run in
+  dry-run mode to test the upload without actually sending files.
+
+  ADDITIONAL SOFTWARE NEEDED FOR THIS SCRIPT
+  * immich-cli (https://immich.app/docs/features/command-line-interface/)
+
+  USAGE
+    -- Configuration
+    1. Install immich-cli on your system
+    2. Configure your Immich server URL and API key in darktable preferences
+    3. The script will appear as "Immich Upload" in the export module storage list
+
+    -- Uploading images
+    1. Select images to export
+    2. Choose "Immich Upload" as the storage option in the export module
+    3. Optionally enter an album name
+    4. Enable dry run mode if you want to test without uploading
+    5. Start the export process
+
+  BUGS, COMMENTS, SUGGESTIONS
+  * Send to Guillaume Godin, godin.guillaume@gmail.com
+
+  CHANGES
+  * 2025-03-16 - Initial version
+]]
+
+local dt = require "darktable"
+local du = require "lib/dtutils"
+local df = require "lib/dtutils.file"
+local log = require "lib/dtutils.log"
+local dtsys = require "lib/dtutils.system"
+local gettext = dt.gettext.gettext
+
+-- Module namespace for preferences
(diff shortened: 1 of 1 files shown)


jscooksey:
Testing this tonight and it worked well, except that if I export the same image again it doesn't see the image(s) already uploaded and so creates another.  immich-cli option??  I must admit I haven't read through much of the documentation, but then it was that easy to get it operational.

OS: Linux Fedora 44
Darktable: 5.4.1
Immich: 2.7.5

Future feature: Be good for it to export JPG with changes, upload both the JPG and RAW/XMP file as a stack.

clnhlzmn:
> if I export the same image again it doesn't see the image(s) already uploaded and so creates another. immich-cli option??

I don't believe this can be done with the immich-cli, at least I don't see that option. It is possible using the immich api directly but I think it's better for something like this to use the immich-cli. But on top of that how would we decide whether an asset should be replaced or not? If I export again with slightly different edits should we replace or upload a copy? I think uploading a copy is desirable in that case and the user is responsible for determining which assets they want to keep in immich or remove. I think there is one clear case where there _shouldn't_ be an additional copy and that is when a new export is "identical" to a previous export. In that case we shouldn't make a new copy, but also immich itself _should_ be handling that automatically via hashing. So the question is: why don't back to back exports from DT result in jpegs that have the same hash? Is DT inserting some metadata like "export time" into the jpeg that causes the hash to be different? I wonder if that is something that can be solved on the DT config side.

> Be good for it to export JPG with changes, upload both the JPG and RAW/XMP file as a stack.

Yeah that would be interesting but if we continue to use immich-cli for this we'll be at the mercy of what it exposes. 

On this I have actually been conflicted about the need for an uploader at all. It is also possible to just mount your DT library as an external library in immich. You can tell immich to ignore raw files and then just use the default DT export setting which is to put exports as siblings to the raws. If your raw library is an immich external lib then all you need to do is do a default export in DT and when immich automatically rescans the library your assets will be found.

guigodin:
Glad to see my original work on this has been picked up. Never had the time to improve it or test it fully, always worked fine for me over the last year or so.

> On this I have actually been conflicted about the need for an uploader at all.

I would say that I'm not always working connected to my NAS, there is many occasion where it would be too slow and unpratical to work of a mounted library and working locally with the ability to export directly to immich fits perfectly in this scenario and save a bunch of time (export to file then go to immich and upload). As a matter of fact, my exported photos library is completely independent from my RAW library and fully managed by immich, so even at home, I would export directly to Immich though in this case you'd be right, I could have the photos library mounted and export to file.

clnhlzmn:
@wpferguson what do you think about this?

daniel-nicolai-hansten:
I think its a bad idea to use a wrapping around Immich-cli, when Immich have a stable API for this types of operations. Would it not be better to make it compatible with Immich V3 API?

clnhlzmn:
If we make it compatible with the v3 API then it doesn't work (maybe) for users running v2 or v4. If we give the user a way to point to the CLI then they can install the v2, v3, or v4 CLI and it still works (as long as the CLI interface is stable enough). And using the CLI means we don't need to update this when that hypothetical v4 comes out: users just update the CLI to the v4 version.

I realize I'm just shifting the burden of compatibility to the CLI and there still is a breaking point introduced where this script touches the CLI but I'm betting the interface to the CLI is more stable than the API.

daniel-nicolai-hansten:
The API is stable, and a new version should not break anything as long as its not using any deprecated API endpoints.
#542 used the replace endpoint for already exported assets, witch is deprecated and will be removed in the next major version of Immich.

Code speaks louder than words, so I rewrote the original Immich API (with some AI tokens) with the functionality I want and using stable parts of the Immich API in #704. This should work with both Immich v2, v3 and v4.

Added support for replacing assets by uploading a new image and deleting the old.

clnhlzmn:
> for already exported assets

Didn't read your PR but how do you decide what is "already exported"? If I export again from DT it's most likely because I made a change and I don't necessarily want the image replaced. If I export again and it's literally the same image (i.e. the same hash) Immich itself should be preventing the duplicate. If there is an easy way to modify that behavior I'm fine with using the API for it but I don't want to be on the hook for maintaining compatibility with the API (mostly between v2 and v3 because I know there are still a lot of people on v2) which is why I decided to use the CLI.

daniel-nicolai-hansten:
I tag the images in darktable with a Immich exported version tag. I also added a behavior menu to control what happens if you are reexporting images:
image


I will add more info in the PR to describe the behavior of the plugin.

clnhlzmn:
Cool. That looks nice.. I guess we should see what @wpferguson prefers? I'm happy to abandon this if using the API is preferred or if there is no preference even.

wpferguson:
I'll try and look at this during the weekend

PR #696, opened 2026-07-04
Fix the path to the user directory on Windows

### LINKED ISSUE #695: Lua fails to initialize due to error in dt 5.6 on Windows 11
Symptom: panels/buttons that are created by lua scripts no longer exist. The "Lua options" section is not present in the application preferences. And the log file contains:

> 3.9410 LUA ERROR : ...table\share\darktable/lua-scripts/lib/dtutils\string.lua:641: attempt to concatenate a nil value (local 'HOME')

It looks like this line tries to get the path from variable "HOMEPATH", which is not a normal windows path. It should be "USERPROFILE":
https://github.com/darktable-org/lua-scripts/blob/b24dd18ddc0c830b04b323c2202dbd505d75fb76/lib/dtutils/string.lua#L640

I tested locally by changing the variable name in `C:/Program Files/Darktable/share/darktable/lua-scripts/lib/dtutils/string.lua`, and it fixed the problem.

I am using darktable, based on commit b3c41a1ee272642eb96a9eba46e2fcbda6fa9ad9, which is a few commits past version 5.6, with some additional commits. I don't think there are lua changes in the tree.

I think this problem started occurring when darktable started using internally bundled scripts, which effectively updated my scripts. However the scripts in my user directory were only a few months old, and I can't find a recent change that looks like it would trigger this error.

lefth:
This fixes #695.

### DIFF
--- a/lib/dtutils/string.lua
+++ b/lib/dtutils/string.lua
@@ -637,7 +637,7 @@ local PLACEHOLDERS = {"ROLL.NAME",
 
 local PS        = dt.configuration.running_os == "windows" and  "\\"  or  "/"
 local USER      = os.getenv("USERNAME")
-local HOME      =  dt.configuration.running_os == "windows" and  os.getenv("HOMEPATH") or os.getenv("HOME")
+local HOME      =  dt.configuration.running_os == "windows" and os.getenv("USERPROFILE") or os.getenv("HOME")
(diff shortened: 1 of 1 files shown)


wpferguson:
I have some reservations about this since HOMEPATH is not equal to USERPROFILE, but I really don't have a feel for how anyone is using this so I'm willing to try it this way.

The other choice is to do what we do for all other values that aren't found which is return an empty string.

Issue #701, opened 2026-07-16
"disable scripts" in script manager needs update for 5.6


anoderay:
**Description**
The disable scripts button in the scripts module renames <config_dir>/luarc to luarc.disabled (the disable_scripts button callback in tools/script_manager.lua). This worked before the scripts were bundled, when darktable used <config_dir>/luarc as the Lua entry point.

Since 5.6, darktable loads the _bundled_ luarc, which calls require "tools/script_manager" regardless of config_dir/luarc existing or not. Renaming config_dir/luarc no longer disables the scripts. It only prevents the user's own config_dir/luarc from being re-run. The bundled scripts still load and start.

**Steps to reproduce**

use darktable 5.6 with bundled scripts.
scripts module / action: install/update scripts / enable "disable scripts" button / click disable scripts.
observe luarc being renamed in the config dir

Restart darktable.

Expected: the Lua scripts are disabled.
Observed: bundled scripts still load and run.

**Possible fixes**
The current way to disable the scripts is the disable Lua scripts preference. 
a) have the button toggle the disable_scripts preference (with the checkbox as the re-enable path)
b) remove the button in favour of the preference

I'd favour the latter - then enabling/disabling lua scripts is reduced to the preferences.

PR #702, opened 2026-07-17
remove disable scripts option

anoderay:
removes the section for disabling lua scripts from the "scripts" utility module. It renames the `[config dir]/luarc` `luarc.disabled` but this has no effect starting with 5.6 as this file is no longer the entrypoint for lua scripts. Also in preferences/lua we have the option to disable lua scripting now. 

Adresses https://github.com/darktable-org/lua-scripts/issues/701


anoderay:
@wpferguson I know you are busy enough, so sorry for pinging you but I think this is worth having :-).

anoderay:
@wpferguson if you have time I'd be interested in your feedback :-)

PR #704, opened 2026-08-02
Immich export plugin via Immich API

daniel-nicolai-hansten:
Immich export plugin, rewrite of #542 

USAGE
This plugin allows you to upload selected photos to an Immich server (https://immich.app/)
Each image is tracked with darktable tags so re-exporting is safe and cheap:
* an "immich|asset|<id>" tag remembers the uploaded asset id for that image
* an "immich|change|<timestamp>" tag remembers the darktable change timestamp at export time
If an image's change timestamp still matches the cached tag and "Skip unchanged" is
enabled, the export is skipped locally without contacting the Immich server at all.
Otherwise the plugin looks up the cached asset on the server; if it no longer exists
there, or has been trashed, it is treated as gone and the image is uploaded fresh.
When the cached asset is still present, the "Re-export Behavior" option controls what
happens to it: delete/trash it, stack the new upload with it, both, or leave it alone
and upload the image as a new, separate asset.

Album assignment is controlled by the "Album mode" option:
* No album: exported assets are not added to any album
* Custom / Copy existing + Custom: assets are added to the album named in the "Album Title" field
* Copy albums from existing asset: assets are added to whatever albums the previous
  version of the asset already belonged to on the server
* Use Album Title: same as Custom, using the "Album Title" field as the album name
The "Album Title" field is prefilled with the folder name of the exported image when
a title-based mode is selected, but can be edited freely.

In the lua options you must specify:
* the hostname of the server immich
* an api key generated in the Account settings - API Keys menu of the immich server
The lua options also let you set the defaults for "Album mode", "Skip unchanged", and
"Re-export Behavior" shown above, so the export panel opens with your preferred choices.
An "Immich debug logging" option is also available there to log every request/response
to the darktable debug log, which is useful when troubleshooting a failed upload.

### DIFF
--- a/contrib/immich.lua
+++ b/contrib/immich.lua
@@ -1,0 +1,995 @@
+--[[
+  This file is part of darktable,
+  copyright (c) 2024 Giorgio Massussi
+  copyright (c) 2026 Colin Holzman
+
+  darktable is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  darktable is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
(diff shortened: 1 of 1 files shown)


wpferguson:
I'll try and look at this during the weekend

PR #705, opened 2026-08-03
Document where to find script manager in UI

fblampe:
Finding the added menu item took me much longer than it should have. Please adjust my wording as you see fit, e.g. in case "module" is not the right word for the collapsable menu.


wpferguson:
Thanks

PR #707, opened 2026-08-04
Load user luarc files for users that have their own lua-scripts

tenderlove:
I have the lua-scripts repo cloned in $HOME/.config/darktable/lua

Since darktable a4003bc
(https://github.com/darktable-org/darktable/commit/a4003bcd2c9fc0d45efdd4b4ea93e2aeb3ad8ce6) (PR darktable-org/darktable#20848) removed the direct invocation of the user luarc from lua/init.c, that responsibility moved to script_manager.  But it was only called in the `if system_based` case, so my luarc was never loaded.

This commit pulls the user luarc script loading up so that it runs in both modes.

### DIFF
--- a/tools/script_manager.lua
+++ b/tools/script_manager.lua
@@ -1388,16 +1388,16 @@ end
 scan_scripts(LUA_DIR)
 log.msg(log.info, "finished processing scripts")
 
-if system_based then
-  -- exec user luarc file if it exists
-  local luarc = loadfile(USER_LUA_DIR .. "rc")
-  if luarc then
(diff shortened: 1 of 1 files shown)


wpferguson:
Did you test to make sure the change didn't introduce any regressions?

tenderlove:
I haven't run any tests in this repo, but this fixes my setup.  I'll learn how to run the tests in here :)

wpferguson:
I tested:

- with user scripts and user luarc
- with user luarc and no user scripts
- with no user luarc and no scripts

and everything seems to work fine.

Thanks

PR #711, opened 2026-08-31
rename-images - always append extension

anoderay:
Partially addresses https://github.com/darktable-org/darktable/issues/21684

This change ensures the file extension (if one is present in the original image) is always appened if the file-extension-variable isn't used. 

Also removes an erraneous gps-information-reset comment probably left over from copy-pasting the initial block.


anoderay:
git mistake..

PR #712, opened 2026-08-31
rename-images: Always append extension

anoderay:
Partially addresses https://github.com/darktable-org/darktable/issues/21684

This change ensures the file extension (if one is present in the original image) is always appened if the file-extension-variable isn't used.

Also removes an erraneous gps-information-reset comment probably left over from copy-pasting the initial block.


wpferguson:
I tested and...

- if I don't provide an extension, the script does - OK
- If I do provide an extension, the script still adds an extension - Not OK

The change forces our definition of correct on the user whether they want it or not.  darktable is about giving the user choices, not enforcing our (darktable's) definition of correct on the user.

I could live with adding a preference to let users turn this on if they want it.

EDIT: #713 will also help since the user can preview and see the mistake before they make it.

anoderay:
thanks for testing @wpferguson ! To explain my thinking for the choice of always making sure the original extension is appended:
1.) Darktable can break in weird ways when files have no extension as some internal logic relies on parsing the filename backwards to the first `.` it finds. (see https://github.com/darktable-org/darktable/issues/21684). 
2.) Is there actually any use case where you'd want to change the extension of your RAW file willingly?
3.) If `$(FILE.EXTENSION)` is appended in the pattern it will be respected and not applied twice

If you want I can change it to a checkbox below the pattern-textbox "append original extension"? That'd offer more user choice. But as long as the filename-parsing isn`t more robust and we know of at least one instance where it lead to crashes on startup (https://github.com/darktable-org/darktable/issues/21684) this also offers the freedom to unknowingly make darktable un-startable. 

Just say the word and I'll get to work :-).

